PR #15, opened 2023-10-27
Implement new MotorAxis module

falkoschindler:
The main goal of this module is to quickly respond to switches and protect the hardware from damage - even if, for example, no RoSys is running, a different serial interface is being used, or a developer sends "manual" commands via the Lizard console.

Tasks:

- [x] control a `StepperMotor` via position/speed/stop commands
- [x] use `Input` switches to stop the motor if a switch becomes active
- [x] prevent the motor from moving in the wrong direction if switch is already active

Maybe later:

- [ ] allow using inverted switches
- [ ] support one or two switches
- [ ] `reference()` with reference switch: find reference switch with predetermined speed, send message to RoSys
- [ ] approach reference again more slowly?
- [ ] set the internal position to zero when referenced?
- [ ] handle the conversion between motor and output

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -428,6 +428,25 @@ When using multiple stepper motors, they can be set to different values to avoid
 
 The optional acceleration argument defaults to 0, which starts and stops pulsing immediately.
 
+## Motor Axis
+
+The motor axis module wraps a stepper motor and two limit switches.
+It prevents the motor from moving past the limits.
+But in contrast to a simple Lizard rule, it allows to actively move out of the limits when moving in the right direction.
+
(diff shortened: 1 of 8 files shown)


falkoschindler:
Test configuration on RB28 (motor on socket 2, sensors on socket 4):

```py
rdyp = Output(15)
en3 = Output(12)
en3.on()
rdyp.on()

m = StepperMotor(5, 4)

serial = Serial(26, 27, 115200, 1)
p0 = Expander(serial, 25, 14)

in1 = p0.Input(32)
in2 = p0.Input(33)

a = MotorAxis(m, in1, in2)
```

PR #25, opened 2024-01-15
Add Euler angles to the IMU

Johannes-Thiel:
change of the Operating Mode to enabele the IMU tu provide euler angles and save these


Johannes-Thiel:
noticed a mistake in the code while testing, working on a fix

PR #27, opened 2024-01-15
Correction of imu documentation

Johannes-Thiel:
correction of propertie names in Documentation.


Johannes-Thiel:
mistake in code on earlyer commit to branch

Issue #10, opened 2023-10-26
Add esptool.py to release artifacts


falkoschindler:
As mentioned in #5, it would be nice to include esptool.py with the released zip archive.

rodja:
Maybe it should be added to the [requirements.txt](https://github.com/zauberzeug/lizard/blob/main/requirements.txt)?

PR #37, opened 2024-05-24
Add esptool to requirements and pack requirements.txt into the release zip-file

pascalzauberzeug:
As discussed in #10, I added esptool to the requirements and I also added requirements.txt to the release zip content


pascalzauberzeug:
I think we should add versions to the requirements.txt if we pack it into the releases

PR #39, opened 2024-05-24
Add a base class for all motors and adapt MotorAxis to it

pascalzauberzeug:
This PR adds a virtual motor base class and implements it in CanOpenMotor, ODriveMotor and StepperMotor for now.
MotorAxis now accepts all three types of motors, instead of just a StepperMotor like before.

Depends on [PR 40](https://github.com/zauberzeug/lizard/pull/40)


falkoschindler (changes requested):
Thanks for the pull request, @pascalzauberzeug! I like the overall implementation, but there are a few things we need to clarify:

- [ ] Like position and speed, I thing acceleration should be of type double as well. The motor-specific implementation can take care of casting it to an integer if necessary.
- [ ] Even though the Lizard commands tend to use `position(...)` for setting the target position and `position` for reading the measured position (like a property in Python), I'd rather use `position(...)` and `get_position()` in C++. Same for `speed()` and `get_speed()`. I think this is more readable, because both are functions, but with opposing purposes.
- [ ] Should we rename `check_input()` into `can_move()`, now that it returns a boolean?
- [ ] What exactly does `is_running` mean? Do we really need it? I don't see it in use anywhere.
- [ ] `StepperMotor` has two versions of `position(..., ..., ...)` (apart from the no-argument `position()`). Same for `speed(..., ...)`. I don't see that the integer versions are used anywhere. So I guess we can remove them.

PR #41, opened 2024-06-04
Odrive error handling

Johannes-Thiel:
Ad a function, where the Odrive module now supports the heartbeat message of the Motor axis. If the firmware version of the Odrive is `0.5.6` the hearbeat will return a `motor_error_flag`. The version can be given as a third variable(for example `m = ODriveMotor(can, 0x100,6)` , when caling the constructor of the OdriveMotor. When no version is given, version `0.5.4` will be used. This can be reset using the reset_motor() function. After this the motor will work again. this reset works for both firmware versions, but only `0.5.6`will give feedback.


falkoschindler (approved):

Johannes-Thiel:
I tested the code after the review. Everything works as intendet.

PR #43, opened 2024-06-18
More information on storage failure

SeaTechRC:
Hello,

As of now there is no exact explanation for when the storage module failed to write to flash.
In order to be able to differentiate between different failure modes, the error message includes the esp_err_t string in the sotrage class.

### DIFF
--- a/main/storage.cpp
+++ b/main/storage.cpp
@@ -20,15 +20,15 @@ void write(const std::string ns, const std::string key, const std::string value)
     esp_err_t err;
     nvs_handle handle;
     if ((err = nvs_open(ns.c_str(), NVS_READWRITE, &handle)) != ESP_OK) {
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
The inline assignment `if ((err = ...) != ESP_OK) {` is a bit unconventional, but fine with me.

Assuming you already compiled and tested the implementation, I happily approve the PR. Feel free to merge. 👍🏻

falkoschindler (commented):
I compiled it myself and noticed an error due to declaring `err` but using `error`. This was easy to fix.
But now I feel like testing it on a real ESP at least once before merging.

falkoschindler:
Ok, I erased the storage of an ESP using `esptool.py erase_flash` and - after flashing and restarting - got the expected error message including the error name "ESP_ERR_NVS_NOT_FOUND":

```
error while loading startup script: could not peek storage storage.num_chunks (ESP_ERR_NVS_NOT_FOUND)
```

PR #49, opened 2024-06-26
Fix Serial::write_checked_line

SeaTechRC:
Hello,

During a bit of digging I came across the fact, that the Serial::write_checked_line uses the util/uart.cpp check function as if it appends a checksum to the line.
This is a suggestion for a fix.

There are a few spots where potential buffer overflows remain unchecked.
I am unsure how this should be handled, which is why I included 2 TODOs.

Also: do note that this is a proposal and as of now is completely untested.


falkoschindler (commented):
Thanks for this pull request, @SeaTechRC!

You are right: `write_checked_line()` should _add_ a checksum before writing the buffer to UART. But it uses `check()`, which does the opposite. It remove a checksum and raises if the checksum doesn't match the string.

However, I think there's a problem with your solution. While `check()` can work on a buffer by simply stripping the final `@xx` and decrementing the length by 3, adding a checksum causes a potential buffer overflow (as you noticed). And even worse, I'm not 100% sure the `message` for `write_checked_line()` can't contain line breaks. Since we can't easily _insert_ checksums in a buffer, I tend to use the same algorithm like `echo`, looping over the buffer and printing lines with checksums on the fly.

I'll try it out and see how it goes.

falkoschindler (approved):
I think I found a pretty neat solution. By iterating over the message buffer and writing to UART line by line, we don't need to copy anything into additional buffers - apart from the checksum itself.

@JensOgorek Could you please have a brief look at the code for an additional pair of eyes? Thank you!

JensOgorek (approved):
Looks fine to me

Issue #18, opened 2023-11-08
Improve integration of expander module


falkoschindler:
- [x] When instantiating an expander module, the execution should wait until the expander has sent a Ready message.
- [x] When instantiating a proxy module, the execution should wait until the expander has sent a first broadcast message (see #66).
- [x] Later we could add constructor parameters for skipping the wait to save time when working with many expanders.
- [ ] New methods like `expander.await_ready()` or `proxy.await_broadcast()` allow awaiting many expanders at once.
- [x] A new `last_message_age` property can be used to define safety rules that stop the machine on connection loss.
- [x] Expressions for reading properties need to be evaluated at runtime, not at compile time (see PR #51).
- [x] Some kind of heart beat message should make sure a lost connection to the expander is detected.

PR #51, opened 2024-06-28
Introduce property expression to avoid evaluating properties too early

### LINKED ISSUE #36: Error Accessing change Attribute of Input on Expander Module
## Description
When attempting to access the change attribute of an input that is part of the expander module in the following code snippet:

`when bump_code == 1 and bumper_front.change > 0 then bump_code = 2; end`
an error occurs during the startup of the ESP. The issue does not arise when the input is not part of the expander.

## Error Message
The error message displayed is:

`rosys/hardware/communication/serial_communication.py:75: read: error while loading startup script: unknown property "change"@78`

falkoschindler:
This PR tries to fix #36 and part of #18 by not evaluating property getters during compilation.

Open tasks:

- [x] Fix this example:

    ```py
    when core.debug == true then core.print('DEBUG'); end
    ```
- [x] Why is this example not working?
    ```py
    let wait do; await core.debug == true; core.print('done'); end
    ```
    When calling `wait()` and setting `core.debug = true`, there should be a message "done".

- [x] Fix this example:
    ```py
    serial = Serial(26, 27, 115200, 1)
    p0 = Expander(serial, 25, 14)
    b = p0.Input(5)
    b.pullup()
    
    when b.level == 0 then core.print("DOWN"); end
    when b.change > 0 then core.print("PRESS"); end
    ```
    This startup script fails, because the compiler needs to read `b.level` and `b.change` to get their datatype. But they are undefined until the expander has started.


falkoschindler:
Yay, I finally finished this pull request! 🎉

@Johannes-Thiel @JensOgorek Let's review it together next week. This way I can introduce you a little more to how the Lizard language is parsed and compiled.

JensOgorek (commented):
Seems good

PR #55, opened 2024-07-12
Expander with 2 Pins

JensOgorek:
I've changed the expander, so I can run without a boot and enable pin set. You can't use `expander.flash()` anymore, but the rest of the functionality is the same.

---

Feature request: #54


falkoschindler (approved):
Thanks @JensOgorek!
Let's use `GPIO_NUM_NC` instead of GPIO_NUM_MAX` to indicate unconnected pins.
Apart from that I only added a tiny info to the documentation.
Let's merge! 🚀

PR #58, opened 2024-07-23
Refactor CANOpen and introduce DunkerMotor module

falkoschindler:


Johannes-Thiel:
I have tested if the Master has to send a sync counter. Sofar I haven't found a case where it has to be send.
I tested with the fieldfriend. For the D1 the PDOs dont work with or without counter.

PR #62, opened 2024-08-07
Expander last_message_age property.

SeaTechRC:
Hello,

This change enables one to write rules or higher level controls which are dependent on wether or not the expander actually transmits data.
This could be used for motor-movement based on sensor data from a sensor on an expander, where communication could fail or is highly time-critical.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -653,6 +653,10 @@ The `disconnect()` method might be useful to access the other microcontroller on
 
 Note that the expander forwards all other method calls to the remote core module, e.g. `expander.info()`.
 
+| Properties         | Description                                             | Data type |
+| ------------------ | ------------------------------------------------------- | --------- |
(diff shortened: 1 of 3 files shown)


falkoschindler (approved):
Thanks for the pull request, @SeaTechRC! The "last_message_age" property is a great idea!

I just added it to the documentation and removed some trivial comments.
And I think `this->last_message_millis = millis();` can be moved out of the if-condition checking for "!!". This condition is only for distinguishing broadcast messages that need to be evaluated as Lizard code and other messages that will simply be printed to the terminal. `last_message_millis` should always be updated, independent of the message type or content.

@Johannes-Thiel Can you briefly check this new feature with a Robot Brain? Thanks!

Johannes-Thiel (approved):

Johannes-Thiel:
I did a test on the testbrain and the result was as expected.

PR #67, opened 2024-09-10
GPIO strapping for Debugging & Update to IDF 5.3.1

JensOgorek:
Adds `core.strapping()` to read the relevant strapping pins for flashing from the registry.
The necessary functions to read all pins reliably are found in idf 5.0.0
Since we wanted to update anyway, this is the chance to do so.

- [ ] find solution
- [ ] update to IDF 5.3.1
- [ ] add new so

I will create a new clean pullrequest, since the 5.3.1 update went a bigger than I thought

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,6 @@
 [submodule "components/esp32-zeug"]
 	path = components/esp32-zeug
 	url = git@github.com:zauberzeug/esp32-zeug.git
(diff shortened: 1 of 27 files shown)


JensOgorek:
This might be enough, but someone has to check it with a robotbrain @Johannes-Thiel

JensOgorek:
This pull request was split into multiple other pull requests, since it got too big.

PR #68, opened 2024-09-16
Update Lizard to IDF 5.3.1

JensOgorek:
This contains a full upgrade to idf 5.3.1, including the components (no owl update)

- [x] Change how idf modules are loaded
- [x] update all Lizard modules
- [x] update lizard components
- [x] change esp-serial-flasher to newer esp32-serial-flasher from espressif
- [x] update the stepper motor module to work with IDF 5.3.1
- [x] test Analog module
- [x] test pwm module

Some testing might still be necessary, especially for the modules with bigger changes. OWL and the i2c driver are not updated yet.


JensOgorek:
The StepperMotor module is now working with the legacy driver. I also tested with the newer pulse_cnt, but it did not work, fully since it did not want to report any positioning. 
With the fix, all Modules should be functional.

JensOgorek:
With this approved, also the idf 5.2.1 update branch can be archived

JensOgorek:
This should now be ready @falkoschindler

falkoschindler (approved):

PR #69, opened 2024-09-17
GPIO status & setting pins

JensOgorek:
With this feature, core can now actively set pins for debugging with `core.set_pin(pin,value)` as requested and get all information about a specific pin with `core.gpio_status(pin)`, mainly to be used with rosys.

Update: Also added reading the strapping register entries for the strapping pins.

---

Related discussion: #60


JensOgorek:
With this closed, the "GPIO reader" branch can be archived. It was a testing ground for this with idf 5.3.1

falkoschindler (approved):

PR #73, opened 2024-09-23
Update PWM output frequency to 2000 Hz

Johannes-Thiel:
When testing on the Fieldfriend we noticed some flickering in the PWM on lower duty cycles. When increasing the Frequency this does not happen anymore.


JensOgorek (approved):
Looks good

falkoschindler:
@Johannes-Thiel and I just discussed that a default frequency of 2000Hz might be a bit random, and LEDs flickering at 0% duty cycle might be more of a application-specific problem and can be fixed with a line in the startup script. Therefore I'll close this PR - at least until we have a stronger argument for such a (breaking) change.

PR #74, opened 2024-09-30
Expander improvements I

JensOgorek:
This is from #18 

- [x]  Wait for full ESP Boot Dump to be written -> Delay at start.
- [x] Move "Ready." msg infront of main loop
- [x] Added reboot command to 2-pin-expander
- [x] Added "is_ready" property for expander
- [x] Added indefinite waiting from expander module for waiting msg [1. Point from #18]
- [x] Added "boot_wait_time" Parameter for indefinite waiting or waiting with timeout [3. Point from #18]
- [x] Added new Property to proxy, to see if proxied properties were created. The older "hotfix" properties could not get used for safe checks, but this one is.[2. Point from #18]
- [x] fix crashes
- [x] fix expander module not reading the "ready" signal.
- [x] Added check for Proxy if Expander is ready, without stopping the program execution


JensOgorek:
Delaying the main task is generally not a good idea. If it is delayed too long, the esp watchdog will trigger and will reset the esp.
The solution is to initiate the expander module and not wait for the expander in the initialization. But check regularly if the expander itself is ready. If it is, the expander module will be set ready, and will be able to handle proxies, for example.

falkoschindler (approved):

JensOgorek:
It seems to be working. But I muted the __PONG__ answer, since it leads to a lot of unnecessary spam.

PR #77, opened 2024-10-15
Remove size validation

JensOgorek:
The manual size validation is removed and is fully handled by ```--flash_size detect```, that is already added the esptool.py.

In some cases, the esptool.py flash_id call could not connect to the esp32, even if the flashing worked. For example, with the Zauberzeug Robotbrain.

### DIFF
--- a/flash.py
+++ b/flash.py
@@ -41,18 +41,6 @@ if 'enable' in sys.argv:
         esp.activate()
     sys.exit()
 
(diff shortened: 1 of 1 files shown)


JensOgorek:
We tested it yesterday without the check lines, which worked.

falkoschindler (approved):

Issue #79, opened 2024-10-17
monitor Odrive velocity


JensOgorek:
We want to Monitor the velocity of the Odrive motors in the Mower attachment. There are no external sensors on the Motors, so we have to use the sensorless mode of the Odrives. The [Can Protocol](https://docs.odriverobotics.com/v/0.5.6/can-protocol.html) is provided from the Docs and the [Sensorless mode](https://docs.odriverobotics.com/v/0.5.6/commands.html#setting-up-sensorless) needs to be setup and tested.

PR #84, opened 2024-10-31
Add speed tracking to Odrive

Johannes-Thiel:
A user wants to track the speed of the rotation of the Motor https://github.com/zauberzeug/lizard/issues/79. To do this the Encoder estimate register is read. When the `this->properties["speed_check"] = std::make_shared<BooleanVariable>();` propertie is set to true, the speed will be checked on every loop.


Johannes-Thiel:
I tested on the testboard. The speed check is working as expected. But sofar the check if it is tuned off is ot working.

Johannes-Thiel:
The function was already implementet. I added the documentation to the docs

JensOgorek (approved):

falkoschindler (approved):

PR #85, opened 2024-11-04
Expander improvements II

JensOgorek:
This will handle the 2nd update from the #18.  With this change, the expander module will be able to detect a crash within the expander esp32. It will then unready itself until the expander esp32 is responding again.
Also, the handling of proxies is changed. Proxy initialization is tied to the expander module's state, and the proxy itself will update its ready flag based on the expander.

- [x] added heartbeat handling
- [x] possible to change is_ready on esp crash (expander & proxy)
- [x] change initialization of proxies (state)
- [x] Update is_ready of proxy with step
- [x] Testing everything

### DIFF
--- a/main/modules/core.cpp
+++ b/main/modules/core.cpp
@@ -163,6 +163,9 @@ void Core::call(const std::string method_name, const std::vector<ConstExpression
             echo("Not a strapping pin");
             break;
         }
+    } else if (method_name == "is_alive") {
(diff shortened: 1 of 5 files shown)


falkoschindler (commented):
I'll review after merging into https://github.com/zauberzeug/lizard/tree/expander_improvements.

Issue #86, opened 2024-11-05
Error while loading startup script: buffer too small.


pascalzauberzeug:
While setting up new robot brains, I noticed that 0.5.3 is not working correctly.
I get the following error which occurred with both long and short configurations. Both work with 0.5.2
```
2024-11-04 16:06:21.594 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: error while loading startup script: buffer too small. discarded line.@7f
```

full restart log:
```
2024-11-04 16:06:20.141 [DEBUG] rosys/hardware/communication/serial_communication.py:80: send: core.restart()@43
2024-11-04 16:06:20.265 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (13) boot: ESP-IDF v5.3.1 2nd stage bootloader
2024-11-04 16:06:20.274 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (13) boot: compile time Oct 28 2024 15:56:02
2024-11-04 16:06:20.285 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (13) boot: Multicore bootloader
2024-11-04 16:06:20.295 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (13) boot: chip revision: v3.0
2024-11-04 16:06:20.306 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (16) boot.esp32: SPI Speed   : 40MHz
2024-11-04 16:06:20.316 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (20) boot.esp32: SPI Mode    : DIO
2024-11-04 16:06:20.327 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (23) boot.esp32: SPI Flash Size : 8MB
2024-11-04 16:06:20.337 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (27) boot: Enabling RNG early entropy source...
2024-11-04 16:06:20.348 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (31) boot: Partition Table:
2024-11-04 16:06:20.359 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: I (34) boot: ## Label      Usage     Type ST Offset  Length
[...]

PR #89, opened 2024-11-05
Buffer comparison cast fix

JensOgorek:
This is a fix for #86.
Size_t is unsigned and pos can be negative. After the change, I could not reproduce the problem.

Also, I increased the size of the patter queue to 100, like in the main queue with uart 0. I ran into a full queue multiple times.
`W (3232) uart: Fail to enqueue pattern position, pattern queue is full.`

### DIFF
--- a/main/modules/serial.cpp
+++ b/main/modules/serial.cpp
@@ -6,7 +6,7 @@
 
 #define RX_BUF_SIZE 2048
 #define TX_BUF_SIZE 2048
-#define UART_PATTERN_QUEUE_SIZE 20
+#define UART_PATTERN_QUEUE_SIZE 100
 
 Serial::Serial(const std::string name,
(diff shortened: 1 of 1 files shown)


falkoschindler (commented):
Ok, but there's also
```c
if (num_chars > buffer_len - 1)
```
in string_utils.cpp and multiple
```c
buffer_len - pos
```
in expressions.cpp. I wonder if we need casting in these lines as well if mixing `int` and `size_t` is an issue. 🤔

falkoschindler (approved):
I see. Then the comparison addressed in this PR should be the only problematic case (as far as I can see). 👍

JensOgorek:
Right now, yes. But maybe we should have an eye open for this and change it sometimes.

Issue #92, opened 2024-11-08
Default expander timing


JensOgorek:
When I created my test setup for strapping pin check, I noticed a weird boot behavior. 
The core esp now has a boot delay (former 2 sec) and if the expander module's check interval is every 1 sec + the timeout again 1 sec, the delay + boot sequence is always longer and the expander esp will be restarted endlessly.

I checked the boot time and reduced the delay a bit (1.5 sec). Also, the default ping timeout is now 2 seconds, that results in a 3 seconds operation combined.

Additionally, the configure.py sometimes does not correctly get its checksum with the added delays

PR #93, opened 2024-11-08
update: tuned default timing

JensOgorek:
fix for #92

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -7,21 +7,21 @@ import serial
 
 if len(sys.argv) != 3:
     print(f'Usage: {sys.argv[0]} <config_file> <device_path>')
-    exit()
+    sys.exit()
 
 txt_path, usb_path = sys.argv[1:]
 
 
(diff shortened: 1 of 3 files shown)


JensOgorek:
Also changed how configure works. With the new delays, the checksum call got lost. Now it will check if it reads a "Ready." and asks for the checksum after.

JensOgorek:
> - Why do we need to reduce the initial delay from 2000ms to 1500ms?
Within time measurement, I noticed that the ESP needs way less time to fully dump all boot up information, it could even be lowered in the future.
> - Why does increasing the ping interval help?
That was actually a typo, the change is supposed to increase the timeout time. When the whole boot up takes more than 2 seconds (2 sec delay + normal boot up time) and the expander module has a 2-second window (1 sec interval + 1 sec timeout), it will always fail and send the reboot command.

falkoschindler (approved):

Issue #88, opened 2024-11-05
IMU not working


pascalzauberzeug:
Setting up an IMU is not working. We tested 0.5.2, but I guess the bug was introduced with the IDF Update in 0.5.0. It works with 0.4.0

```
2024-11-05 00:17:43.994 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.004 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.014 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.025 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.035 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.055 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2259) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.065 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2269) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.076 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2269) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.086 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2269) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.097 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2279) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.108 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2279) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.118 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2279) BNO055: (i2c WL) Error: 263
2024-11-05 00:17:44.129 [DEBUG] rosys/hardware/communication/serial_communication.py:72: read: E (2289) BNO055: (i2c WL) Error: 263
[...]

PR #94, opened 2024-11-14
Imu fix

Johannes-Thiel:
We have the issue #88. This PR is a fix for this problem

### DIFF
--- a/main/modules/imu.cpp
+++ b/main/modules/imu.cpp
@@ -25,7 +25,6 @@ Imu::Imu(const std::string name, i2c_port_t i2c_port, gpio_num_t sda_pin, gpio_n
     }
     this->bno = std::make_shared<BNO055>((i2c_port_t)i2c_port, address, GPIO_NUM_MAX, GPIO_NUM_MAX);
     try {
-        ESP_LOGI("IMU", "Initializing BNO055");
         this->bno->begin();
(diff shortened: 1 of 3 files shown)


Johannes-Thiel:
The fix is done. The problem was due to a Timing issue in the BNO class.
I also fixed a bug in the Constructor. Without this the BNO will not handle its internal states correctly.

main/modules/imu.cpp
```diff
@@ -23,7 +23,7 @@ Imu::Imu(const std::string name, i2c_port_t i2c_port, gpio_num_t sda_pin, gpio_n
     if (i2c_set_timeout(i2c_port, 1048575) != ESP_OK) {
         throw std::runtime_error("could not set i2c timeout");
     }
-    this->bno = std::make_shared<BNO055>((i2c_port_t)i2c_port, address);
+    this->bno = std::make_shared<BNO055>((i2c_port_t)i2c_port, address, GPIO_NUM_MAX, GPIO_NUM_MAX);
```
JensOgorek: This seems to be the default.
https://github.com/zauberzeug/lizard/blob/114e1ab0709ae1f6ffcff36cb5b921118cdce99b/main/modules/BNO055ESP32.h#L547

JensOgorek (approved):

Issue #75, opened 2024-10-14
Check strapping pins before flashing port expander


SeaTechRC:
Hello,

I have trouble flashing the p0 expander on the robot brain.
See the log:
![image](https://github.com/user-attachments/assets/ab9e620d-823a-4609-8c71-4763b657ee80)

To minimize interference of other components, I have tried writing the following startup configuration manually:
```
!-
!+serial = Serial(26, 27, 115200, 1)
!+p0 = Expander(serial, 25, 14)
!.
```
and then simply `p0.flash()`
I fear is, that this might actually be a flash page that is broken.

It's unfortunate that the bootloader of p0 doesn't give more details about the error.

SeaTechRC:
[https://github.com/espressif/esp-serial-flasher/blob/a4f5033895489be0f1f75d254d3a040a9231560b/private_include/protocol.h#L79](The comment) in the header file actually explains it I think. In that case it should be a broken flash page.

falkoschindler:
As it turned out, some strapping pins have not been in the correct state.
We will add a check before flashing so that the user sees an informative error message.

PR #95, opened 2024-11-14
Check relevant pins for p0.flash()

JensOgorek:
This is for #75
----
Before flashing, the p0.flash() call will set the GPIO0 to the correct state and check GPIO2 and GPIO12 on the p0 for their states. 
GPIO0 and 2 are responsible for booting to the bootloader upon restart. 
GPIO12 on low sets the flash voltage to 3.3V. On high, which sets it to 1.8V, it can cause issues when flashing.

The strapping register itself does not need to be checked in that stage of flashing. Since the configuration after the restart would be the relevant one, but then the p0 would be in bootloader mode, which we would not have to check and could not check.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -691,7 +691,8 @@ The expander module allows communication with another microcontroller connected 
 | `expander.flash([force])` | Flash other microcontroller with own binary data | `bool`    |
 | `expander.restart()`      | Restart other microcontroller                    |           |
 
-The `flash()` method requires the `boot` and `enable` pins to be defined. The optional `force` argument skips the check for strapping pins.
+The `flash()` method requires the `boot` and `enable` pins to be defined.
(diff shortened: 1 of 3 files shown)


falkoschindler (approved):
@JensOgorek I reviewed and simplified the code a bit. This avoids code duplication for message handling. And I think we can simply use `strstr` twice instead of reading and comparing the `int level`. Maybe you can have a look and check if it does work as expected.

JensOgorek:
I tested it, it worked like before. 
Also, I added a `force` argument to skip the pin check. When the expander esp is in an invalid state, it would never answer and could never be flashed. 
To make the flashing more reliable, I added a restart and a full disconnect before flashing. The flash would sometime not work (and work again after restart). Now it seems to work flawlessly.

falkoschindler (approved):
Ok, I added another check for the correct number of flash arguments and extended the documentation just a tiny bit.

PR #97, opened 2024-11-15
Proxy defaults

JensOgorek:
Handle default properties for proxies on creation without expander esp being ready.

This is a followup for #18

The defaults are handled by a static function in the module class.

- [x] create functionality to get defaults
- [x] handle new creation of properties
- [x] handle properties in proxy
- [x] change input & output module
- [x] test current version
- [x] add all modules


main/modules/module.cpp
```diff
@@ -419,3 +419,12 @@ void Module::write_property(const std::string property_name, const ConstExpressi
...
 }
+
+std::map<std::string, Variable_ptr> Module::get_default_properties_for_type(const std::string &module_type) {
+    if (module_type == "Output")
+        return Output::default_properties();
+    if (module_type == "Input")
+        return Input::default_properties();
+    // We'll add other module types as we implement them
+    return {};
+}
```
falkoschindler, line 430: @JensOgorek Can't we avoid the need for countless if-conditions and use inheritance instead?

JensOgorek:
Replaced the static default_properties() methods with a virtual get_default_properties() function in the Module class, allowing each module type to define its own properties while eliminating the need for a separate list to maintain.

main/modules/output.cpp
```diff
@@ -3,9 +3,14 @@
...
+std::map<std::string, Variable_ptr> Output::get_default_properties() const {
+    return {
+        {"level", std::make_shared<IntegerVariable>(-1)},
+        {"change", std::make_shared<IntegerVariable>(-1)}};
+}
+
 Output::Output(const std::string name) : Module(output, name) {
-    this->properties["level"] = std::make_shared<IntegerVariable>();
-    this->properties["change"] = std::make_shared<IntegerVariable>();
+    this->properties = this->get_default_properties();
```
falkoschindler, line 13: @JensOgorek I think we can do that in the `Module` base class. Then individual modules only need to overwrite `get_default_properties` and are done.

falkoschindler (changes requested):
As we just discussed, we should keep the initial "level" and "change" values 0.

Apart from that, we can't call `Module.create` on the core ESP. We need a different solution.

falkoschindler:
@JensOgorek Whatever you did, somehow this PR now also contains changes from main which makes it hard to review. 😕

JensOgorek:
> @JensOgorek Whatever you did, somehow this PR now also contains changes from main which makes it hard to review. 😕

I got it fixed. Somehow, the main > proxy_defaults merge failed.

falkoschindler (changes requested):
There are still changes from main contained in this PR, like the whole "check_strapping_pins" topic in expander.cpp.

Maybe we need to start over with a fresh branch, cherry-pick individual commits and delete and burn this branch afterwards... 🔥

JensOgorek:
> There are still changes from main contained in this PR, like the whole "check_strapping_pins" topic in expander.cpp.
> 
> Maybe we need to start over with a fresh branch, cherry-pick individual commits and delete and burn this branch afterwards... 🔥

Good idea, I bring the flamethrower tomorrow

JensOgorek:
Flamethrower was active, and we have a new pull request with #104

PR #103, opened 2024-11-19
Introduce pre-commit hooks

falkoschindler:
To finally put an end to changing file endings back and forth, this PR introduces pre-commit hooks.

To install it on your system (once for all repositories you're working with):
```bash
pip install pre-commit
```

To install it in your local checkout (once for this repository):
```bash
pre-commit install
```

To check all files manually (every now and then):
```bash
pre-commit run --all-files
```

But once installed, `pre-commit` will automatically run before each commit or push, fixing things like training whitespace or wrong end-of-files. Changes will appear as unstaged changes.


Johannes-Thiel (approved):

JensOgorek (approved):
I like that :+1:

PR #104, opened 2024-11-20
Proxy defaults with static functions

JensOgorek:
This is a fixed version of the #97 merge


falkoschindler (approved):
Ok, I did just a few changes:

- removed some blank lines,
- removed the unused `cached_default_properties`
- as well as the private `module_type` member (which was only used for an error message) and
- added the missing `this->name` to the error message

Feel free to merge.

PR #109, opened 2024-12-03
Property assignment fix

JensOgorek:
This fixes the creation of properties. They were all linking to the same property. The examples show the same lizard startup script

Before (just changed the input of the last added pin):
```
core 3913755 0 false 0 false 0 false 0 false 0 false 0 false 
core 3913765 1 true 1 true 1 true 1 true 1 true 1 true
```
Now (again, changing the last added pin):
The first line looks different, since it is now configured correctly.
```
core 290031 1 true 1 true 1 false 0 false 1 true 1 true
core 290021 1 true 1 true 1 false 1 true 1 true 1 true
```

### DIFF
--- a/main/modules/analog.cpp
+++ b/main/modules/analog.cpp
@@ -6,12 +6,11 @@
 #include "freertos/task.h"
 #include "uart.h"
 
-const std::map<std::string, Variable_ptr> &Analog::get_defaults() {
-    static const std::map<std::string, Variable_ptr> defaults = {
+const std::map<std::string, Variable_ptr> Analog::get_defaults() {
+    return {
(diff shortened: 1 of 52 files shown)


falkoschindler (approved):
Oh, what a bug! Now the code is (hopefully) correct - and cleaner. 🙂
Fingers crossed that we don't need to touch all these 52 files once again for a while...

JensOgorek:
> Oh, what a bug! Now the code is (hopefully) correct - and cleaner. 🙂 Fingers crossed that we don't need to touch all these 52 files once again for a while...

My tests say no, but you never know :D

PR #110, opened 2024-12-04
Proxy persistent property handling

JensOgorek:
This adds the ability for the expander module to set expressions even if the expander esp is not ready yet. It will restore them on reboot as well.


falkoschindler (approved):
I got a little confused by the method names: `write_property` calls `add_property` which calls `setup_property`. But I'm out of better ideas. Feel free to merge. 👍

Issue #24, opened 2024-01-10
core.print breaks some characters


falkoschindler:
Minimal reproduction:
```py
core.print('boo\n')
```

This doesn't print "boo", but only "oo". The "b" is converted to a backspace.

- The following characters are affected: "b", "f", "n", "r" and "t".
- The problem doesn't occur without the trailing newline "\n".

Somehow these five characters are converted into their escaped counterparts backspace, form feed, new line, carriage return and tab.

JensOgorek:
Adding more characters after the `\n` gets even funnier:
```
> core.print("boo\nbooboo")
oo
ooo"
```
```
> core.print("boo\nboobooboo")
oo
oooo"
```
```
> core.print("boo\nbooboobiibaa")
oo
ooiaa"
```
```
> core.print("boo\nbooboobiibaauu")
oo
ooiaauu"
```
```
> core.print("boo\nbooboobiibaauub")
oo
ooiaau"
```
There is something funny happening in the parser, I guess.

JensOgorek:
It is definitely a problem how the parser handles ESCAPE CHARs
`#define ESCAPE_CHAR(c, info) ((c) == 'b' ? '\b' : (c) == 'f' ? '\f' : (c) == 'n' ? '\n' : (c) == 'r' ? '\r' : (c) == 't' ? '\t' : (c))`
in its code.

falkoschindler:
Maybe we can reproduce it online?
https://ianh.github.io/owl/try/

JensOgorek:
When processing escape sequences (like '\n'), the code used the original string length instead of the processed length. For example, "\n" has length 2, but should be length 1 after processing.

PR #112, opened 2024-12-05
Fix escape character handling in parser.h

JensOgorek:
This is a fix for #24 

Test outputs:
```
> core.print('boo\nbobobobobob')
"boo
bobobobobob"

> core.print('rbnr\nt')
"rbnr
t"

> core.print('boo\n')
"boo
"
```

No characters were lost in printing

### DIFF
--- a/gen_parser.sh
+++ b/gen_parser.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-echo "This generator is does not include changes from PR #107 and should not be used."
-exit 1
-
 echo "Generating parser..."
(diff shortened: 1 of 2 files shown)


falkoschindler (approved):
Crazy!
Like with PR #107, we shouldn't forget to add this fix to the code that generates parser.h, e.g. forking from https://github.com/ianh/owl.

JensOgorek:
> Crazy! Like with PR #107, we shouldn't forget to add this fix to the code that generates parser.h, e.g. forking from https://github.com/ianh/owl.

I added it to #113

PR #114, opened 2024-12-05
Function calls with properties

JensOgorek:
This will change the mute(), unmute() and broadcast() call, so they also work with properties.

- [x] Implement module properties instead of variables -> module level properties, that can be changed as call or property
- [x] add call properties -> Properties that handle like calls
- [ ] documentation

### DIFF
--- a/main/modules/analog.cpp
+++ b/main/modules/analog.cpp
@@ -15,7 +15,7 @@ const std::map<std::string, Variable_ptr> Analog::get_defaults() {
 
 Analog::Analog(const std::string name, uint8_t unit, uint8_t channel, float attenuation_level)
     : Module(analog, name), unit(unit), channel(channel) {
-    this->properties = Analog::get_defaults();
+    this->merge_properties(Analog::get_defaults());
 
     if (unit < 1 || unit > 2) {
         echo("error: invalid unit, using default 1");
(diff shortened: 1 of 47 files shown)


JensOgorek:
I tested it for Input, this solution works but might be not final, if I find a simpler solution.
Do we only want to do that with the default calls or module specific calls as well, like pullup, pulldown in Input? These need more coding, since they are not a simple bool setting.

Also, @falkoschindler yes, I need to change all modules again :D

JensOgorek:
I added the most calls that only use one (or one with a default) parameter to be handled as a property by the proxy/expander. Some modules with multiple settings on different calls, like `Input`'s pull modes are combined to `pullmode`.

It works like this
```
proxy_input = p0.Input(x)
proxy_input.pullmode="pullup" 
```
It now does the same as `proxy_input.pullup()`, but it will be saved in the property list inside the expander and will be resent to the expander, if it is rebooted.
The "call properties" are only saved on the proxy side, they are not saved on the expander itself (except the module was already saving something in the call). This makes no functional difference, since they still can be read directly, for example with `proxy_input.pullmode`.
Calls with two parameters are not supported, since we have no handling for multiple parameters. If we need it, we have to create a solution for expressions with multiple arguments like `->evaluate_array()` or something similar.
Also, I don't know if we need all those calls to be "call properties", if they are too much, we can remove some.

JensOgorek:
I will add the documentation, once we finalize what has to be added or removed and how we add the call properties to the documentation

falkoschindler (changes requested):
We decided to put this PR on hold and evaluate a different approach, pausing the execution while booting a port expander. This would avoid the need for keeping track of methods and properties.

JensOgorek:
Should we archive this now? Since #117 is merged into main?

Issue #115, opened 2024-12-11
Improve expander module handling during startup


JensOgorek:
After #111 & #114 we decided to change the way expander are handled in startups.

What is to change:

- [x] after initiating an expander module, stop the normal execution of the startup script or the console with a while loop, until the expander is ready or a timeout is surpassed.
- [x] execute the startup script "normally" (most expander/proxy functionality won't be set up correctly)
- [x] make it possible to flash an expander esp in the error state 
- [x] let the error state be broadcasted → `core.has_error` can be subscribed with broadcast

Error handling, example implementation for expander module:
- [x] error code functionality
- [x] set errors 
- [x] have an error status readable → `core.has_error` is a bool property, that can be subscribed with broadcast
- [x] have a way to receive all errors → `core.get_errors()`

Ideas for future additions:
- [ ] if a proxy is created while the expander module is not ready, there should be an error
- [ ] don't let critical functions execute when in the error state
- [ ] Handle different error states → ideas: No Error, Connection Timeout, Connection Failed
- [ ] Notify the user about the error state → other than broadcast, maybe when using Lizard
- [ ] important modules have their own error states → Core, Expander?
- [ ] when meeting the initial timeout, try to reboot the expander esp until it is finally in an error state

PR #117, opened 2024-12-12
Expander module startup handling improvements

JensOgorek:
This is the pull request for #115 

Current state:
- Holds the execution of the startup script (or the main loop) when an expander module is created. It gets held until the esp sends "ready." or the creation times out.
- The is_ready property can only be switched to true during the expander esp boot up.
- The proxy module not set up the proxy, when the expander module is not ready (maybe will also add an error in the future)
- The proxy module is not writing its commands with the expander's serial. It will write its commands with the expander now (instead of using the proxy_information list in Lizard 0.6.2).
- The expander can set two different error codes, that can be read with the core module with its property core.has_error and get all the errors collected with core.get_errors() → moved to its own pull request


JensOgorek:
I also added an example error code handling with just the expander module. With that, core can see if there are errors and get them. In this test implementation, only the expander module can set errors.

falkoschindler (approved):
Great! I just did two minor simplifications. Ready to merge! 🚀

PR #118, opened 2024-12-13
Expander boot handling with error codes

JensOgorek:
This is a possible implementation of error codes on the expander module.

### DIFF
--- a/main/modules/core.cpp
+++ b/main/modules/core.cpp
@@ -1,6 +1,7 @@
 #include "core.h"
 #include "../global.h"
 #include "../storage.h"
+#include "../utils/error_handling.h"
 #include "../utils/ota.h"
 #include "../utils/string_utils.h"
 #include "../utils/timing.h"
@@ -21,12 +22,14 @@ Core::Core(const std::string name) : Module(core, name) {
     this->properties["millis"] = std::make_shared<IntegerVariable>();
     this->properties["heap"] = std::make_shared<IntegerVariable>();
     this->properties["last_message_age"] = std::make_shared<IntegerVariable>();
+    this->properties["has_error"] = std::make_shared<BooleanVariable>(false);
 }
 
 void Core::step() {
     this->properties.at("millis")->integer_value = millis();
     this->properties.at("heap")->integer_value = xPortGetFreeHeapSize();
     this->properties.at("last_message_age")->integer_value = millis_since(this->last_message_millis);
+    this->properties.at("has_error")->boolean_value = Error_handling::has_error();
     Module::step();
 }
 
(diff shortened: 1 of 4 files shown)


JensOgorek:
@falkoschindler I moved the error handling to its own pull request

falkoschindler (commented):
I have some questions and doubts about the current concept:

1. Are error codes meant to be binary flags that can be added to get combined errors like, e.g., 5 = 1 (low voltage) + 4 (no response from motor driver)? If so, `set_error` would overwrite existing errors when setting another one.
2. Currently, `set_error` can be called with `ERROR_NONE`, which results in a wrong value for `has_error_`. It looks like the `ErrorHandling` API would need additional methods like `add_error`, `remove_error` and `clear_errors` to cover all cases. But then it might be easier to just expose the `error_codes` map directly.
3. Is it a good idea to share a common list of error codes across all modules? I can imagine conflicts when a module needs an error similar but not identical to one from another module. We would either need to introduce many error codes with similar names and meanings, or pricisely document error codes _per module_. If so, I don't see the point in a common list of error codes.

Maybe the current implementation with only two errors in a single module is too early to see the big picture. But at the moment I lean towards a new "error" property in the expander module with two additive codes (connection timeout and connection loss). This way it can be accessed like any other property, used in rules, and added to the core output line.

If we really want a global error property or method in the core module, we can iterate over modules and aggregate errors. But right now there is no benefit.

JensOgorek:
Branch is restored and diff should be correct now.

JensOgorek:
>     1. Are error codes meant to be binary flags that can be added to get combined errors like, e.g., 5 = 1 (low voltage) + 4 (no response from motor driver)? If so, `set_error` would overwrite existing errors when setting another one.

I thought about it, and using a bit mask could solve this. Each error would be a single bit, allowing multiple errors to be combined without overwriting. 

>     2. Currently, `set_error` can be called with `ERROR_NONE`, which results in a wrong value for `has_error_`. It looks like the `ErrorHandling` API would need additional methods like `add_error`, `remove_error` and `clear_errors` to cover all cases. But then it might be easier to just expose the `error_codes` map directly.

With a bit mask, ERROR_NONE is unnecessary, since a value of 0 naturally represents no errors. However, I think we would still need these API methods (`add_error`, `remove_error`, `clear_errors`) to handle the bit mask more easily and efficiently.

>     3. Is it a good idea to share a common list of error codes across all modules? I can imagine conflicts when a module needs an error similar but not identical to one from another module. We would either need to introduce many error codes with similar names and meanings, or pricisely document error codes _per module_. If so, I don't see the point in a common list of error codes.

I think some error codes could be in all modules (`TIME_OUT`), but having individual error code lists for the modules gives more flexibility and avoids conflicts between similar errors.

JensOgorek:
This looks good so far. I will verify it to work with #125 once it is merged with main.

JensOgorek:
This is somewhat what I imagine for error codes. Yet, there is a bug in this somewhere with flashing the expander.

PR #120, opened 2025-01-07
fix reset and enable confusion

Johannes-Thiel:
When debugging some Hardware Problems, we noticed, that enabel resets the ESP and does not simply enable it. To fix this I added a new enable function to the flash script and changed the enable function to reset. 
This should fix the confusion of the naming.

### DIFF
--- a/flash.py
+++ b/flash.py
@@ -6,7 +6,7 @@ from esp import Esp
 
 
 def show_help() -> None:
-    print(f'{sys.argv[0]} [nano | xavier | orin] [nand | v05] [usb | /dev/<name>] [enable] [-e | --erase]')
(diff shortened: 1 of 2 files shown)


flash.py
```diff
@@ -38,7 +38,13 @@ def show_help() -> None:
 if 'enable' in sys.argv:
     with esp.pin_config():
         print('Enabling ESP...')
-        esp.activate()
+        esp.enable()
+    sys.exit()
+
+if 'reset' in sys.argv:
```
falkoschindler, line 44: We should add this command to the help text.
By the way, we should switch to using `argparse` in the future. But this is out of scope of this PR.

PR #121, opened 2025-01-09
Update owl without hotfixes

JensOgorek:
This updates owl to the current main version without the need for any hotfixes.


JensOgorek:
Also, I tried the new error code with memory allocation, that works as well.

falkoschindler (approved):
Looks good to me! 👍🏻

PR #124, opened 2025-01-22
Expander fixes: Proxy call with wrong name and ping conditions

JensOgorek:
These are two fixes for the expander module.

The first fixes proxy calls, they were using the wrong name. 

The second fix changes the ping conditions. There are proxy modules (like Serial) that can be created, but do not create an output immediately, these would cause a timeout of the expander module, even if everything is correct. I noticed this while debugging for the first fix.


falkoschindler (approved):

JensOgorek:
Will merge when @pascalzauberzeug gives his ok.

pascalzauberzeug (approved):
Works just fine!

PR #125, opened 2025-01-24
Register module defaults using static initialization pattern

JensOgorek:
This pull request will remove the need for a separate list for get_defauts() from each module.


falkoschindler:
For reference, here is a great explanation by claude.ai how the macro works:
```c++
#define REGISTER_MODULE_DEFAULTS(module_name)                                   \
    namespace {                                                                 \
    struct RegisterDefaults {                                                   \
        RegisterDefaults() {                                                    \
            Module::register_defaults(#module_name, module_name::get_defaults); \
        }                                                                       \
    } register_defaults;                                                        \
    } // namespace
```

> This code snippet shows a macro definition `REGISTER_MODULE_DEFAULTS` which is used to automatically register default values for different types of modules in the system. Let me break down how it works:
> 
> 1. The macro takes a single parameter `module_name`
> 
> 2. It creates an anonymous namespace containing:
>    - A struct called `RegisterDefaults`
>    - A global instance of that struct named `register_defaults`
> 
> 3. The struct's constructor:
>    - Calls `Module::register_defaults()` with two arguments:
>      - `#module_name`: The stringified name of the module (the # operator converts the parameter to a string literal)
>      - `module_name::get_defaults`: A reference to the static get_defaults function of the module class
> 
> 4. The purpose of this pattern is to automatically register module defaults when the program starts up, before main() runs. When the global `register_defaults` object is constructed during program initialization, it will automatically call the registration function.
> 
> This is a common C++ technique called "static initialization" or the "registration pattern". It allows modules to self-register their defaults without requiring explicit registration calls in the main code.
> 
> For example, if used like this:
> ```cpp
> REGISTER_MODULE_DEFAULTS(BluetoothModule)
> ```
> 
> It would expand to code that registers the defaults for the BluetoothModule by calling `Module::register_defaults("BluetoothModule", BluetoothModule::get_defaults)` during program startup.

falkoschindler (approved):
Looks good to me!
I just renamed `defaults_func` to `defaults_function` and introduced the type aliases `DefaultsFunction` and `DefaultsRegistry` to make the method signatures more readable.

Issue #129, opened 2025-02-19
Twai driver not restarting


Johannes-Thiel:
When working on the f10 robot there is a issue with the Twai driver. When the ESP is started at first the driver will work. After the Estop is pressed and released the driver will not work and there will be the ```error in bluetooth message handler: could not send CAN message and could not restart twai driver``` message. This does not happen on every press and release, but once it is broken it won't fix itself.

When the EStop is pressend and everyting is normal there is the message ```error in routine "stop": could not send CAN message```. This is due to the Odrives being disabled, that beahviour is expected. 
When the communication breaks, at first the individual commands like ```wheels.speed(1,0)``` will not work. The error message is similar to the bluetooth message. 
I am woking with this script.
<details>
<summary>script</summary>

```
rdyp = Output(15)
en3 = Output(12)
bluetooth = Bluetooth("fieldfriend-f10")
can = Can(32, 33, 1000000)
l0 = ODriveMotor(can, 0)
r0 = ODriveMotor(can, 512)
l1 = ODriveMotor(can, 256)
r1 = ODriveMotor(can, 768)
l0.m_per_tick = 0.03953674121405751
r0.m_per_tick = 0.03953674121405751
l1.m_per_tick = 0.03953674121405751
r1.m_per_tick = 0.03953674121405751
l0.reversed = true
r0.reversed = false
l1.reversed = true
r1.reversed = false
wheels = ODriveWheels(l0, r0)
wheels_front = ODriveWheels(l1, r1)
wheels.width = 0.78
wheels_front.width = 0.78
wheels.shadow(wheels_front)

serial = Serial(26, 27, 115200, 1)
p0 = Expander(serial, 25, 14)




bms = p0.Serial(26, 27, 9600, 2)
bms.unmute()

estop_1 = Input(34)
estop_2 = Input(35)
battery_control_reset = p0.Output(15)
battery_control_status = p0.Input(13)
# TODO: remove when lizard issue 66 is fixed.
battery_control_status.level = 0
battery_control_status.active = false

bumper_front_top = p0.Input(35)


rdyp_status = Input(39)
vdp_status = p0.Input(39)
# TODO: remove when lizard issue 66 is fixed.
vdp_status.level = 0
vdp_status.active = false

let stop do wheels.speed(0, 0);end
when estop_1.level == 0 then stop(); end
when estop_2.level == 0 then stop(); end
when core.last_message_age > 1000 then wheels.speed(0, 0); end
[...]

PR #130, opened 2025-02-21
Add status check and recovery from BUS_OFF for twai driver

Johannes-Thiel:
As discussed in https://github.com/zauberzeug/lizard/issues/129, in some cases the twai will go into the BUS_OFF state and can't be resatret directly. It has to be recovered first. 
This PR implements a check for the status of the twai driver, if a can massege cant be send. It will check if the Bus is off and if it is, it will start a recovery. Afterwards it will restart the Bus as before.


JensOgorek (approved):
Looks fine with me

PR #131, opened 2025-02-21
External expander for IDF 5.3.1

JensOgorek:
This is the external expander module for 5.3.1. It is recreated in here, since there were major changes on Lizard since the last development of the external expander.

- [ ] Untested / not confirmed working changes from the old 4.7 external expander
- [ ] Initialize external expander with _<ID>_READY_ check
- [ ] Confirmed one-way communication with ID-tag and ID being handled correctly
- [ ] Activate and deactivate external expander mode on esp
- [ ] Confirmed two-way communication with ID-tag and ID
- [ ] Setting up proxies
- [ ] Update proxies with external expander
- [ ] Work with multiple external expanders
- [ ] Handle mode switches correctly (no spam from multiple expanders on deactivation)

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -369,6 +369,26 @@ void process_uart() {
             break;
         }
         int len = uart_read_bytes(UART_NUM_0, (uint8_t *)input, pos + 1, 0);
+
+        if (input[0] == '0x80') {
+            core_module->set_external_mode(true);
+        } else if (input[0] == '0x81') {
+            core_module->set_external_mode(false);
+        } else if (core_module->is_external()) {
+            if (input[0] == '@' && !(input[1] == core_module->get_expander_id())) {
+                echo("Debug: not for me");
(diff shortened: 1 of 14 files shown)


JensOgorek:
The External Expander works with one Expander connected. When another expander is connected (even just one pin for reading), the first connected expander will stop working. With the UART mode we use, the ESPs most likely interfere with each other. 
I will change it up to use half duplex mode with UART when using the UART connection to an external expander.

JensOgorek:
I did some tests and finally figured out the UART communication problems and why the UART behaved so unpredictable.

JensOgorek:
The current version works with one (UART0) external expander. With UART0, I can't fully deactivate the interfering signal cause of the UART USB converter. Using UART1 seems to work. I will wait for the Plexus expander to arrive and update the code for it to work as an external expander when it's here.

PR #132, opened 2025-03-11
Add more IMU properties and allow setting the operation mode

falkoschindler:
This PR allows to access more properties (calibration, acceleration, magnetometer, gyroscope, Euler/quaternion orientation, linear acceleration, gravity, temperature) and to change the operation mode.

Open tasks:

- [x] measure time for reading _all_ data -> about 15ms
- [x] change `set_mode` method to a read-/writable `mode` property -> maybe later, not necessary right now
- [x] selectively read attributes from BNO055 based on current value of ~~this property~~ a new `data_select` property

The latter is an idea to reduce the I2C traffic. If the operation mode is GYROONLY, it doesn't make too much sense to fetch acceleration data.
But on the other hand, it might still be useful to fetch orientation data, even though the datasheet says such fused data isn't available in non-fusion modes. Empirically it looks like orientation data is still provided and would be useful to avoid the need to integrate angular velocities on the host system.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -189,35 +189,50 @@ The constructor expects up to five arguments:
 - `address`: client address of the MCP (0x28 or 0x29, default: 0x28)
 - `clk_speed`: I2C clock speed (default: 100000)
 
-| Properties    | Description                                     | Data type |
-| ------------- | ----------------------------------------------- | --------- |
-| `imu.cal_sys` | calibration: system (0 to 3)                    | `int`     |
-| `imu.cal_gyr` | calibration: gyroscope (0 to 3)                 | `int`     |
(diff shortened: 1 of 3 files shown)


falkoschindler:
Reading _all_ data takes about 15ms which is quite significant. Reading only calibration and gyro takes only 3ms.

falkoschindler:
I decided to introduce a new `data_select` property that can be used to control which data is read from the IMU. E.g. 0x09 would read the calibration (0x01) and the gyroscope (0x08). This is a little bit more cryptic than 9 individual properties, but much more compact and probably also more efficient.

On one of our robot the code runs fine for several weeks now. The `data_select` property has just been added today, but seems to work as well.

@Johannes-Thiel Can you briefly check if the code looks good to you? Thanks!

Johannes-Thiel (approved):

PR #133, opened 2025-03-13
Fix mismatch between `can.status()` and `can.get_status()`

falkoschindler:
The documentation mentions `can.get_status()`, but the implementation checks for `can.status()`. This PR changes the implementation to check for `can.get_status()` because this is a better name for a method. Since this method is probably rarely used, the breaking change shouldn't be a problem.


JensOgorek (commented):
👍

Issue #134, opened 2025-03-21
UART being overloaded with too many requests fix


JensOgorek:
We had a problem with a robot, it crashed with a lot of checksum mismatch outputs.
The traceback pointed towards the UART. 

It seems the UART was overloaded with too many requests that flooded in while it was still sending.

I recreated the issue with two esps. The expander esp had 21 input modules to create a higher load on the connection.
After both were running, I could trigger the issue with spamming a command like `p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.info();p0.version()`

After spamming it for multiple seconds, it resulted in the same behavior the robot displayed.

The fix is to wait for the TX to be done sending within the echo method.

JensOgorek:
Yes, we had internal discussions about that issue, but we appreciate all feedback. I'm currently doing other important tasks, but when these are done I will go back to this.
I was experimenting a few weeks ago with better parallelization of all communication in Lizard. Since everything is just handled by the main thread at this time.

falkoschindler:
Likely the same root cause as #213 (_Main loop is throttled by blocking console writes_), just with a different symptom.

`echo()` → `printf` writes to UART0 with **no TX ring buffer** ([main/main.cpp:405](https://github.com/zauberzeug/lizard/blob/main/main/main.cpp#L405)), so it busy-waits the bytes into the 128 B hardware FIFO and the main loop is blocked for the whole transmission. While it's blocked pushing a long line it isn't draining RX (`process_uart` runs once per loop, _before_ the module-step echoes) — so under a command flood plus many broadcasting modules (the 21-input-module repro here), incoming bytes pile up faster than they're consumed and the RX side overflows → checksum mismatches. Same blocking-TX root cause as #213, but here RX integrity is the casualty instead of loop rate.

That also fits why #135 ("wait for uart TX to be done") didn't pan out and #136 is still open: blocking _longer_ in `echo` widens the window the loop spends not reading RX, so it makes this worse, not better.

The #213 mitigations pull the other way — shrink or remove the blocking window:

- **Raise the console baud** (e.g. 921600): shortens the TX block ~8×, so the loop gets back to draining RX much sooner. Cheapest thing to try. (Caveat: on the expander link the baud is effectively hardcoded at 115200, so there it's a coordinated change on both ends, not just config.)
[...]

PR #135, opened 2025-03-21
Wait for uart TX to be done

JensOgorek:
This will fix the problem from #134 
It will wait for each echo to be fully send out. We tested this already on a robot, and my stress test also does not result in checksum mismatches.

Sometimes, there are still checksum mismatches, but they are caused by a different problem. Yet, they are not breaking like this one was.

### DIFF
--- a/main/utils/uart.cpp
+++ b/main/utils/uart.cpp
@@ -4,6 +4,7 @@
 #include <stdexcept>
 #include <stdio.h>
 #include <string>
+
 void echo(const char *format, ...) {
     static char buffer[1024];
 
--- a/main/utils/uart.cpp
+++ b/main/utils/uart.cpp
@@ -1,9 +1,9 @@
 #include "uart.h"
+#include "driver/uart.h"
 #include <cstdarg>
 #include <stdexcept>
 #include <stdio.h>
 #include <string>
-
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
@pascalzauberzeug I'm wondering if there is a significant performance impact when we wait for each message transmission individually, effectively disabling the FIFO buffer. Could you, please, run a short test comparing this PR with the main branch. What I'm interested in is basically how many core messages are sent per second, which is a good indicator for how fast the main loop runs. If there isn't a big difference, you can approve and merge this PR. Thanks!

JensOgorek:
I tested it on my ESPs, and it will decrease the rate of messages being sent out.

pascalzauberzeug:
@falkoschindler I will look into it on tuesday, I wanted to test it on another robot anyway

JensOgorek:
We can wait with that PR for now. It seems there might be a different problem in Can right now. The UART handling needs some work, but this is not the root of the problem.

JensOgorek:
This PR should not be pushed. It slows down so much, that can looses even more messages, than it is already missing.
Having 2 High speed UARTs active at the same time must be handled somehow, but not like this.

JensOgorek:
This is not fixing the problem.

PR #138, opened 2025-04-04
improved syncignore for working with the build folder

JensOgorek:
This is just a small change to the syncigonore folder. Now, it will just sync lizard.bin, lizard.elf, bootloader.bin and partition_table.bin. For easier use when developing with a robot that uses livesync.


falkoschindler (approved):
Not nice, but we can't find a better solution (until livesync supports negative excludes in some way).

PR #139, opened 2025-04-15
add enable and disable to actuators

Johannes-Thiel:
When discussing the safety of our robots, we made the decision to introduce an enable property to our actuators and outputs. These will be used to define rules in the startup script to enable and disable specific actuators in specific scenarios.


main/modules/canopen_motor.cpp
```diff
@@ -533,7 +537,42 @@ double CanOpenMotor::get_speed() {
...
+    // Check if enabled state has changed
+    if (this->properties["enabled"]->boolean_value != this->enabled) {
+        if (this->properties["enabled"]->boolean_value) {
+            this->enable();
+        } else {
+            this->disable();
+        }
+    }
+
+    // ... rest of existing step() code ...
```
falkoschindler, line 562: ?

main/modules/motor_axis.cpp
```diff
@@ -5,7 +5,9 @@
 REGISTER_MODULE_DEFAULTS(MotorAxis)
 
 const std::map<std::string, Variable_ptr> MotorAxis::get_defaults() {
-    return {};
+    return {
+        {"enabled", std::make_shared<BooleanVariable>(false)},
```
falkoschindler, line 9: Why does `enabled` default to `false` for this module?

docs/module_reference.md
```diff
@@ -122,11 +131,16 @@ The PWM output module is associated with a digital output pin that is connected
 | ------------------ | ---------------------------------------- | --------- |
 | `output.duty`      | Duty cycle (8 bit: 0..256, default: 128) | `int`     |
 | `output.frequency` | Frequency (Hz, default: 1000)            | `int`     |
+| `output.enabled`   | Whether the output is enabled            | `bool`    |
 
 | Methods        | Description             | Arguments |
 | -------------- | ----------------------- | --------- |
 | `output.on()`  | Turn on the PWM signal  |           |
 | `output.off()` | Turn off the PWM signal |           |
+| `output.enable()` | Enable the PWM output  |           |
```
falkoschindler, line 140: Some table columns are not well-aligned. Usually this should happen automatically if you have [Prettier](https://prettier.io/) installed.

main/modules/motor_axis.cpp
Johannes-Thiel: This seems to be a module I forgot to change back to `true`. In the beginning I put every module in `false`, so that the module needs to be `enabled` specificly. This might be interesting in the long run. But to emulate the current behaviour I changed it to `true` for every module.

JensOgorek (approved):
Looks good. The setup thing might need a change if this is critical.

main/modules/d1_motor.cpp
```diff
@@ -154,12 +168,16 @@ void D1Motor::handle_can_msg(const uint32_t id, const int count, const uint8_t *
 }
 
 void D1Motor::setup() {
```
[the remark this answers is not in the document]
Johannes-Thiel: You are rigth. With this function the Motor will be enabeld and ready to operate. We want to ensure that this is not blocked by the enabel due to it not moving the axis. For this reason it will not have a damgerous impact, but is very important to execute.

JensOgorek (approved):

main/modules/motor_axis.cpp
```diff
@@ -14,6 +16,9 @@ MotorAxis::MotorAxis(const std::string name, const Motor_ptr motor, const Input_
 }
 
 bool MotorAxis::can_move(const float speed) const {
+    if (!this->properties.at("enabled")->boolean_value) {
```
falkoschindler, line 19: Shouldn't we use `this->enabled` instead?

main/modules/motor_axis.cpp
```diff
@@ -58,7 +72,32 @@ void MotorAxis::call(const std::string method_name, const std::vector<ConstExpre
...
         Module::call(method_name, arguments);
     }
 }
+
+void MotorAxis::enable() {
+    this->properties.at("enabled")->boolean_value = true;
+    this->enabled = true;
+    try {
+        this->motor->enable();
+    } catch (const std::exception &e) {
```
falkoschindler, line 90: Why do we need to catch exceptions here (and in `disable()`)?

falkoschindler (commented):
Ok, I finally reviewed this PR. I made quite some changes which we can discuss in person later.

Some quick thoughts:

- `output.pulse` doesn't seem to be guarded if the output is disabled, right? Shouldn't we do this?
- Including `output.invert` and `output.active` in this PR makes it even bigger and more complicated to review. As it turned out, there was a pretty significant copy&paste bug hiding. I hope I spotted all problems and we haven't missed some combination of level, inverted, enabled, active...
- It is pretty much boilerplate code to add `enabled` properties to each module. In Python I would introduce a mixin, but in C++ I'm not so sure. Let's leave it this way.
- Why did you add `.deconfigure = False` in stepper_motor.cpp? It seems unrelated to this PR, but probably won't hurt.

falkoschindler (approved):
We discussed the remaining points in person. Ready to merge! 🚀

PR #141, opened 2025-05-07
Add disable function to esp class

pascalzauberzeug:
The esp class already supports enabling and resetting the microcontroller, this adds functionality to disable it.
This will be used in a future security feature to shut the microcontroller down, when it doesn't respond anymore.

After it's disabled, it needs to be reset to get running again; not to be enabled.

### DIFF
--- a/.syncignore
+++ b/.syncignore
@@ -4,6 +4,7 @@ __pycache__/
 *.tmp
 .env
 .venv
+.mypy_cache
 
 # Ignore everything in build except li*, bo* and pa*
 build/[c-kC-K]*
--- a/esp.py
+++ b/esp.py
@@ -83,7 +83,5 @@ class Esp:
 
     def disable(self) -> None:
         print('Disabling microcontroller...')
-        self.write(f'{self.gpio_g0}/value', self.on)
-        sleep(0.5)
         self.write(f'{self.gpio_en}/value', self.on)
         print('disable complete')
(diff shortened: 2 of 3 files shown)


falkoschindler (approved):

falkoschindler:
> After it's disabled, it needs to be reset to get running again; not to be enabled.

I'm a bit confused that `g0` is used at all for disabling. It should be enough to "release" `en`. Maybe this would allow to _enable_ after using _disable_? The possible pin inversions make it hard to predict.

Anyway, I'm ok with the current implementation. If we want to make sure it is 100% correct, we might want to discuss it once again, maybe with @Johannes-Thiel.

pascalzauberzeug:
@falkoschindler you are correct. I changed it and it works, thank you

Johannes-Thiel (approved):

Johannes-Thiel:
> > After it's disabled, it needs to be reset to get running again; not to be enabled.
> 
> I'm a bit confused that `g0` is used at all for disabling. It should be enough to "release" `en`. Maybe this would allow to _enable_ after using _disable_? The possible pin inversions make it hard to predict.
> 
> Anyway, I'm ok with the current implementation. If we want to make sure it is 100% correct, we might want to discuss it once again, maybe with @Johannes-Thiel.

The inversion of the pins is managed in the beginning of the esp class and the use of decorators like `nand` in the `flash.py`script.

PR #143, opened 2025-05-09
Introduce espresso.py to replace flash.py in the future

falkoschindler:
This PR introduces a new script to flash and configure an ESP microcontroller from a Jetson computer inside a Zauberzeug Robot Brain. It uses argparse to define a clean command line API. In order to avoid breaking changes with downstream libraries like RoSys, we decided to start a new script next to flash.py, which we can deprecate (and remove in Lizard 1.0).

The script comes with a `-d` or `--dry-run` option to test it without a Robot Brain:
```bash
./espresso.py --help
./espresso.py flash -j nano -d /dev/foo
```
Screenshot 2025-05-09 at 15 34 01

We chose argparse over Typer to avoid an extra dependency (which can be cumbersome in robotics projects).

@pascalzauberzeug will test the new script on Monday. Everyone else is invited to share thoughts and opinions.

We plan to release it in version 0.7.1, even though it is technically a new feature and should wait for 0.8. PR https://github.com/zauberzeug/rosys/pull/284 needs it to prevent conflicts with older versions of flash.py. So we'd argue it is more like an infrastructure tool and can be sneaked into a patch release.

Open tasks:

- [x] update getting_started.md and tools.md
- [x] add espresso.py to zip_content.txt
- [x] update upload_ssh.py (use livesync?) or add SSH mode to espresso? --> maybe later
- [x] deprecate flash.py?

### DIFF
--- a/.github/zip_content.txt
+++ b/.github/zip_content.txt
@@ -6,6 +6,7 @@ backtrace.sh
 configure.py
 core_dumper.py
 esp.py
+espresso.py
 flash.py
 monitor.py
 README.md
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,7 @@
   "editor.formatOnSave": true,
   "prettier.printWidth": 120,
   "pylint.args": [
+    "--disable=C0103", // Invalid name
     "--disable=C0114", // Missing module docstring
     "--disable=C0115", // Missing class docstring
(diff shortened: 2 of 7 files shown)


pascalzauberzeug (changes requested):

JensOgorek (approved):

pascalzauberzeug (approved):
Tested and approved

PR #146, opened 2025-05-27
Add UART queue to prevent reset crashes

JensOgorek:
Fixed a crash that occurred when resetting the ESP while it was connected over serial. The UART now has a proper queue to handle incoming data during boot, preventing interrupt crashes.

The issue manifested as a boot loop with the following error when connected via serial:
`assert failed: xQueueGenericSendFromISR queue.c:1180 (pxQueue)`

I noticed the fail with RoSys and Basekit.


falkoschindler (approved):
👍🏻

PR #148, opened 2025-06-03
Esp32s3 Compability

JensOgorek:
This PR contains a Lizard version, that can run on ESP32 and ESP32s3. 

`build.py` was created to build for both supported types quickly by hand and for the workflow to use.

There are two versions of the workflow, both do work and create the outputs as the following:

```
- lizard_firmware_and_devtools_vx.x.x_esp32.zip
- lizard_firmware_and_devtools_vx.x.x_esp32s3.zip
- lizard_vx.x.x_esp32.bin
- lizard_vx.x.x_esp32s3.bin
- Source code (zip)
- Source code (tar.gz) 
```
One is doing the workflows parallel, but is complexer, the other is doing it one by one like the current workflow. When merging, we should just use one of them.

ESP32s3 Support is required for the Plexus (External Expander).

### DIFF
--- a/main/modules/analog.cpp
+++ b/main/modules/analog.cpp
@@ -68,6 +68,12 @@ Analog::Analog(const std::string name, uint8_t unit, uint8_t channel, float atte
     };
     ESP_ERROR_CHECK(adc_cali_create_scheme_curve_fitting(&cali_config, &adc_cali_handle));
 #else
+    adc_cali_line_fitting_efuse_val_t cali_val;
+    esp_err_t cali_check = adc_cali_scheme_line_fitting_check_efuse(&cali_val);
+    if (cali_check != ESP_OK || cali_val == ADC_CALI_LINE_FITTING_EFUSE_VAL_DEFAULT_VREF) {
+        echo("warning: eFuse calibration data not available, using default reference voltage.");
+    }
+
     adc_cali_line_fitting_config_t cali_config = {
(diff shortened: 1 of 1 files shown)


main/modules/analog.cpp
```diff
@@ -59,19 +59,23 @@ Analog::Analog(const std::string name, uint8_t unit, uint8_t channel, float atte
     };
     ESP_ERROR_CHECK(adc_oneshot_config_channel(adc_handle, static_cast<adc_channel_t>(channel), &config));
 
-    adc_cali_line_fitting_efuse_val_t cali_val;
-    esp_err_t cali_check = adc_cali_scheme_line_fitting_check_efuse(&cali_val);
-    if (cali_check != ESP_OK || cali_val == ADC_CALI_LINE_FITTING_EFUSE_VAL_DEFAULT_VREF) {
-        echo("warning: eFuse calibration data not available, using default reference voltage.");
-    }
-
+#ifdef CONFIG_IDF_TARGET_ESP32S3
```
falkoschindler, line 62: @JensOgorek Why is `adc_cali_scheme_line_fitting_check_efuse` gone? Isn't it needed anymore?

JensOgorek: It was an optional check to warn the user about possible inaccurate readings. I forgot to re-add that in case the ESP32 is used.

falkoschindler (approved):

PR #150, opened 2025-06-12
Fix error handling in espresso.py

falkoschindler:
We noticed that a context isn't properly exited when calling `sys.exit` inside of it, causing pins to remain occupied.

This PR replaces `sys.exit` calls with exceptions which are handled at the very end.


falkoschindler:
Sorry, I started with a wrong branch...

PR #151, opened 2025-06-12
avoid calling sys.exit within context managers

falkoschindler:
We noticed that a context isn't properly exited when calling `sys.exit` inside of it, causing pins to remain occupied.

This PR replaces `sys.exit` calls with exceptions which are handled at the very end.

### DIFF
--- a/espresso.py
+++ b/espresso.py
@@ -130,8 +130,7 @@ def erase() -> None:
                 'erase_flash',
             )
             if not success:
-                print('Failed to erase flash.')
-                sys.exit(1)
+                raise RuntimeError('Failed to erase flash.')
 
 
 def flash() -> None:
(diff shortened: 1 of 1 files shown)


espresso.py
```diff
@@ -156,8 +155,7 @@ def flash() -> None:
                 '0x20000', args.firmware,
             )
             if not success:
-                print_fail('Flashing failed. Maybe you need different parameters? Or you forgot "sudo"?\n')
-                sys.exit(1)
+                raise RuntimeError('Flashing failed. Maybe you need different parameters? Or you forgot "sudo"?')
```
pascalzauberzeug, line 158: Maybe a little shorter and more 'commanding'?
```suggestion
                raise RuntimeError('Flashing failed. Use "sudo" and check your parameters.')
```
or 'Make sure to use "sudo"...'

PR #160, opened 2025-09-30
ODrive safeguards for transient CAN/reset glitches

JensOgorek:
This is to handle transient/malformed ODrive 0x009 frames during CAN resets and rare tiny-dt spikes in wheel speed derivation, which previously caused large speed outliers even at standstill.

- Parse 0x009 only when DLC >= 8 and both floats are finite.
- In ODriveWheels, skip derivative update for very small dt (< 2000 µs) and ignore implausible speeds (|v| ≥ 10 m/s).

Impact: Robust against brief bus disturbances; normal behavior unchanged. Thresholds (dt, max speed) are conservative and can be tuned.

- [x] Implement
- [x] Test Odrive motor improvements
- [x] Test Odrive wheels improvements
- [x] Test if we need Odrive wheels improvements

### DIFF
--- a/main/modules/odrive_motor.cpp
+++ b/main/modules/odrive_motor.cpp
@@ -106,7 +106,7 @@ void ODriveMotor::handle_can_msg(const uint32_t id, const int count, const uint8
         break;
     }
     case 0x009: {
-        if (count < 8) {
+        if (count != 8) {
             break;
         }
         float tick;


JensOgorek:
I removed the changed to the reset for now, but left them in code as a comment block. The safeguards for wheels kept their debug output, so we can later see if they get triggered or not.

PR #161, opened 2025-09-30
Add temperature sensor for Plexus Mini & Analog updates

JensOgorek:
- Introduces AnalogUnit module that owns the ADC one shot unit handle. Multiple analog readers on the same hardware unit now share one handle (no re-init, no crashes). This should work like expander and serial work together.
- Updates Analog to take a GPIO pin plus an AnalogUnit instead of unit/channel. The pin is mapped to the ADC channel internally.
- Adds TemperatureSensor module for NTC-based temperature using two pins on the same unit (sensor pin + reference pin). Computes temperature via beta model. This is only implemented for Esp32s3 Modules we use in our Plexus Mini's.

Sensor details:
Thermistor: TDK/EPCOS NTC B57332V5103F360
R25 = 10 kΩ, Beta B25/85 = 3380 K (default used)

- [ ] check if Analog works on normal ESP32 with these changes


falkoschindler (approved):

JensOgorek:
I tested it and it works after the review changes

PR #172, opened 2025-11-12
First spike for BNO085 integration

Johannes-Thiel:
We want to integrate the BNO085 into Lizard. This spike is an implementation of the [Adafruit_BNO08x](https://github.com/adafruit/Adafruit_BNO08x)
 component. I also created an I2C master module to prevent code duplication.


falkoschindler (commented):
@Johannes-Thiel I iterated with Claude for a while, improving several minor issues. Now there are only two things remaining that need clarification:

1. **`I2cBusManager` uses `std::mutex` but Lizard is single-threaded**: The main loop is single-threaded, so the mutex adds unnecessary overhead. If multi-threading is genuinely needed (e.g., for future FreeRTOS task usage), this is fine — but it's inconsistent with the rest of the codebase which doesn't use mutexes for module operations. Consider removing it for simplicity, or add a comment explaining why it's needed.

2. **`ndof_fmc_off` mode enables fewer sensors than `ndof`**: `ndof_fmc_off` only enables acc/mag/gyr/rot (indices 0–3), while `ndof` enables all 7 reports via `desired.fill(true)`. On the BNO055, both modes enable the same sensors — the only difference is whether Fast Magnetometer Calibration is active. Since the BNO085 doesn't have an FMC concept (it uses report-based configuration), `ndof_fmc_off` should either match `ndof` or be removed entirely.

falkoschindler (approved):

PR #194, opened 2026-03-05
Integrate new Motor Module

JensOgorek:
This was converted a rolling/developing PR. The final changes will be moved to their own pull request. Once there is no diff in here, it will be closed.


JensOgorek:
The new motor module has been migrated to an internal repository. Since the motor controller is closed hardware, its driver will be developed and maintained internally rather than in the public Lizard repository.

As already noted, this was a rolling/developing PR and there is no clean, generally useful upstream contribution left to merge here, so we are closing it. Thanks!

PR #199, opened 2026-04-16
Extend MKS servo module: auto-configure bus mode, status property, set_bitrate

felixwrz:
### Motivation

Make the `MksServoMotor` module usable on a multi-drop CAN bus without manual configuration, and expose enough status/control to diagnose and reconfigure drives from Lizard.

- The 0xF5/0xF6 bus motion commands silently do nothing unless the drive is in `SR_vFOC` mode, which previously had to be set by the user before any motion would work.
- There was no way to observe when a mode-change was rejected by the drive.
- The CAN bitrate (0x8A) could not be changed from Lizard, making it awkward to match a drive's bitrate to the rest of the bus.
- Every drive ships listening on CAN id 1, so putting more than one on a bus (e.g. multiple grippers on one delta arm) required a way to reassign ids.

### Implementation

**Auto-configure bus mode and expose set_mode_status**
- `subscribe_to_can()` sends `MODE_SR_vFOC` (0x05) on startup so `0xF5`/`0xF6` work out of the box. The initial sends run *after* the CAN subscription is established, so the `0x82` acknowledgement is observed rather than dropped.
- Replaced the `0x05` magic `MAX_MODE` with named `MODE_CR_OPEN/CLOSE/vFOC` and `MODE_SR_OPEN/CLOSE/vFOC` constants; `MAX_MODE` is now defined in terms of `MODE_SR_vFOC`.
- Added a `set_mode_status` property (default `STATUS_OK`). `send_set_mode` marks it `STATUS_SET_MODE_PENDING` on each attempt; the `0x82` handler validates the 3-byte CRC and resolves it to `STATUS_OK` only on an explicit `0x01` success, otherwise `STATUS_SET_MODE_FAILED`.

**`set_bitrate` method**
- New `send_set_bitrate(hz)` wraps the `0x8A` CAN command.
- `set_bitrate` is exposed to Lizard as an integer in Hz (`125000`/`250000`/`500000`/`1000000`) — the socketcan/python-can/datasheet convention. The rate is exact-match validated against the four supported values; anything else is logged via `echo` and ignored (no clamping onto a wrong bus speed).
- Added `BITRATE_125K/250K/500K/1M` constants in the header.

**`set_can_id` method**
- New `send_set_can_id(id)` wraps the `0x8B` CAN command and exposes `set_can_id(id)` to Lizard (range 1-2047; id `0` is the broadcast address and is rejected).
- This is a **permanent, persisted** change on the drive: afterwards the motor only responds on the new id, so the Lizard script must reconstruct the module with the new `can_id`. The `echo` is worded as a *request* ("requested CAN id change to …") since it is sent before any acknowledgement.

### Breaking change

The constructor now auto-sets `SR_vFOC` (0x05) on every construction. For users who explicitly call `set_mode(N)` after construction the final state is unchanged. But setups that rely on a non-`SR_vFOC` mode **persisted on the drive** — e.g. using the MKS as a smart encoder + enable switch while a separate controller drives it via step/dir — will have that mode silently overwritten at every Lizard boot.

> **Heads up**: existing setups relying on a non-`SR_vFOC` mode persisted on the drive will have it overwritten at every construction. Call `set_mode(N)` after construction to keep using mode N.

### Progress


JensOgorek:
I added this in the dual-drive branch/PR (#194) and it works.

The `set_bitrate` documentation feels a bit overkill to me, but might be okay.

I like the correct use of `bitrate` instead of `baudrate` — by the way, we currently use `baud_rate` in `main/modules/can.cpp` and its documentation. Since CAN uses NRZ encoding (1 symbol = 1 bit) and ESP-IDF itself uses `KBITS`/`MBITS`, shouldn't that also be `bitrate`?
What do you think @falkoschindler ?

Is there more to be added @felixwrz or is this done by now?

falkoschindler:
@JensOgorek Ok, I wasn't aware that bitrate is technically more precise than baudrate in this case. As long as we only change the documentation and don't break any API, I'm fine with switching to "bitrate" for the CAN module.

Apart from that I'd also vote for dropping the bitrate paragraph. Since the available values are already listed above, it barely adds any value to the reader.

felixwrz:
> Is there more to be added @felixwrz or is this done by now?

@JensOgorek there is also the option to change the CAN ID, which might be useful here since the motor always defaults to ID 1. What do you think?

JensOgorek:
@felixwrz yeah for sure. We use something similar with the dual drives, but they default at id5. 
It might be needed somewhat soon, actually. Once we use more than one delta arm on one robot, there might be more than one gripper on one CAN bus.

JensOgorek:
I added a way to configure the can ID. Next week I will test it on a robot and once that is done, the PR is ready to review.

JensOgorek (commented):
I finally found the time to test and review this. I tested `set_can_id` on hardware (switched a drive from id 1 to 2 and back) and it works as expected.

While reviewing I introduced a `crc_ok()` helper that de-duplicates the MKS checksum validation, which was copy-pasted across all four CAN response handlers (`0x31`, `0x32`, `0x39`, `0x82`) — each new handler can now validate in one line. I also added an `echo` hint to `set_can_id`, since the drive replies on the *new* id and this module (still subscribed to the old one) goes silent afterwards, so it needs to be reconstructed with the new `can_id`.

Two minor things left:

**Redundant clamp in `send_set_bitrate`** — the `std::clamp(rate, 0x00, MAX_BITRATE)` is dead code: `call()` only ever passes the already-valid `BITRATE_*` constants, and the `0` lower bound is moot for a `uint8_t`. Can be dropped. (Unlike `send_set_mode`, where the clamp is needed because the raw user integer is forwarded.)

**`::toupper` on a `char` is technically UB** — `toupper` wants its argument as `unsigned char`, but `char` is signed on most platforms, so a non-ASCII byte goes negative. Harmless for ASCII rate strings, but the safe form is:

```cpp
std::transform(rate.begin(), rate.end(), rate.begin(),
               [](unsigned char c) { return std::toupper(c); });
```

What do you think @falkoschindler

falkoschindler:
@felixwrz @JensOgorek Claude and I still found some issues to address. Nothing catastrophic, should be straight-forward.

---

1. Constructor sends before it subscribes — failure can't be observed

   ```cpp
   auto motor = std::make_shared<MksServoMotor>(name, can, motor_id);  // ctor sends 0x82
   motor->subscribe_to_can();                                          // subscribes here
   ```

   The whole point of the new `status` property is to surface a rejected mode change. But the constructor's `send_set_mode(MODE_SR_vFOC)` runs before `subscribe_to_can()`, so the 0x82 acknowledgement is dropped and `status` will stay `STATUS_OK` even if the drive refused. The auto-configure goal is partially defeated.

   Fix: either call `subscribe_to_can()` before sending in the factory, or move the initial `send_set_mode` / `send_working_current` calls into `subscribe_to_can()` (or a new `init()` step), so the subscription is in place when the response arrives.

2. 0x82 response: "anything that isn't 0x00 is OK" is too permissive

   ```cpp
   if (data[1] == 0x00) {
       this->properties.at("status")->integer_value = STATUS_SET_MODE_FAILED;
   } else {
       this->properties.at("status")->integer_value = STATUS_OK;
   }
   ```

   MKS responses typically use `0x01 = success`, `0x00 = fail`, `0x02 = error/timeout`. The current branch treats `0x02` (and any future code) as success. Prefer an explicit positive check (`data[1] == 0x01` → OK; otherwise → failed) so the property reflects what the drive actually reported.

3. `status` is named generically but only tracks set_mode

   The property is called `status` and exposed in the Lizard docs as a general "status code (0 = OK, 1 = set_mode failed)", but `set_bitrate` (0x8A) and `set_can_id` (0x8B) — which also return an ack — don't update it. Either:
   - Rename to something narrower (e.g. `set_mode_status` or `last_command_status` with a clear contract), or
   - Extend the response handlers for 0x8A / 0x8B to set matching status codes.

   The current shape implies broader semantics than it delivers.

4. Stale status after a new `set_mode` call

   If a `set_mode` call fails (`status` → 1) and the next `set_mode` succeeds but its 0x82 response is lost on the bus, `status` stays at 1 forever. Consider resetting `status` to `STATUS_OK` (or a new `STATUS_PENDING`) in `send_set_mode` so the property reflects the most recent attempt rather than the last known failure.

5. `set_can_id` `echo` is misleading — sent before acknowledgement

   ```cpp
   this->send(data, 3);
   echo("%s set_can_id: motor now responds on CAN id %d - reconstruct the module with the new can_id", ...);
   ```

   The "motor now responds on CAN id N" message fires immediately after enqueuing the send, regardless of whether the drive ever accepts the change. If the bus rejects it, the user is told the change took when it didn't. Either tone down the wording ("requested CAN id change to N…") or gate the echo on the 0x8B response.

6. PR description undersells set_can_id

   The title and "Implementation" section cover auto-mode, `status`, and `set_bitrate`, but `set_can_id` (a permanent, persisted change to the drive) is only mentioned indirectly via the docs section. Worth adding a bullet so reviewers / users don't miss a significant new feature.

7. `set_bitrate` should take an integer, not a string

   `set_bitrate("250K")` is the only string-typed method on the module — every other setter (`set_mode`, `set_working_current`, `set_holding_current`, `set_can_id`, `position`, `speed`, `stop`) takes integers. It's a stylistic outlier in an otherwise integer-typed API.

   Other concerns with the string form:
   - **Typos fail silently to `echo`.** `motor.set_bitrate("250k ")` (trailing space) just logs and moves on; the script keeps running on the wrong bitrate. An integer would either match a known rate or trip type-checking at the call site.
   - **The vocabulary is internally inconsistent**: `"125K" / "250K" / "500K" / "1M"`. Why is 1000K spelled `"1M"`? Everyone has to remember the spelling rule, not just the rate.
   - **Parsing cost for no real win**: `std::transform` to uppercase + 4-way string compare, versus a single switch over an integer.

   Prefer `set_bitrate(250000)` (Hz). That's the universal CAN-tooling convention (`socketcan`, `python-can`, every MKS datasheet expresses rates this way), so anyone copy-pasting from a datasheet just types the number. `set_bitrate(250)` (kbps) is a reasonable alternative if Hz feels heavy (but this would be another convention the user would need to know or look up in the documentation). Either way the docs can still list the four legal values.

   Note that an integer would fix both minor issues raised by @JensOgorek.

8. Auto-setting the mode in the constructor is a (small) breaking change — call it out

   For users who explicitly call `set_mode(N)` after construction, the final state is unchanged (their call still wins). The affected population is the people who _don't_ call `set_mode` because they rely on the drive's persisted mode — e.g., using the MKS as a smart encoder + enable switch while a separate controller drives it via step/dir. For them, every Lizard boot now silently overwrites the persisted mode with `0x05`.

   Small audience, but the change is visible to the drive even when the user never touches `set_mode`, so it's not zero. A one-line note in the PR description (and ideally a CHANGELOG / release-notes entry) is proportionate, e.g.:

   > **Heads up**: existing setups relying on a non-`SR_vFOC` mode persisted on the drive will have it overwritten at every construction. Call `set_mode(N)` after construction to keep using mode N.

felixwrz:
Thanks @falkoschindler, all points addressed. Rundown:

**1. Constructor sends before it subscribes** ✅
Moved the initial `send_working_current` / `send_set_mode(MODE_SR_vFOC)` out of the constructor into `subscribe_to_can()`, after the subscription is established. The `0x82` ack is now observed, so `set_mode_status` actually reflects a rejected mode change at boot.

**2. `0x82` response too permissive** ✅
The handler now treats `0x01` as the only success; `0x00` (fail) and anything else (e.g. `0x02` error/timeout) map to failed.

**3. `status` named generically but only tracks set_mode** ✅
Renamed the property to `set_mode_status`. I went with the rename rather than extending to `0x8A`/`0x8B`: changing the bitrate or CAN id makes the drive respond on a different bus/id, so those acks can't be reliably observed on the subscribed channel anyway — the property genuinely only tracks set_mode.

**4. Stale status after a new set_mode** ✅
Added `STATUS_SET_MODE_PENDING` (2), set on every `send_set_mode`. The handler resolves it to OK/failed, so a lost ack no longer leaves a stale failure code — the property always reflects the most recent attempt.

**5. `set_can_id` echo misleading** ✅
Reworded to "requested CAN id change to N — if accepted, the motor responds on the new id …" so it no longer asserts an outcome that wasn't acknowledged.

**6. PR description undersells set_can_id** ✅
Added a motivation bullet and a dedicated `set_can_id` section to the PR description, flagging it as a permanent, persisted change.

**7. `set_bitrate` should take an integer** ✅
Switched to integer Hz: `set_bitrate(250000)`. Exact-match validated against `{125000, 250000, 500000, 1000000}` — anything else is logged and ignored rather than clamped onto the wrong bus speed. This also resolves @JensOgorek's two earlier nits (silent typos, the `"1M"`/`"…K"` vocabulary). Dropped the string parsing and the now-dead `MAX_BITRATE`.

**8. Auto-setting mode is a small breaking change** ✅
Added a "Breaking change" heads-up to the PR description: setups relying on a persisted non-`SR_vFOC` mode will have it overwritten at every construction; call `set_mode(N)` after construction to keep mode N.

**All of the above changes were also tested on hardware.**

falkoschindler (approved):
Thanks @felixwrz, this is great — every point from the last round is cleanly handled, and the hardware retest is reassuring.

I pushed a few tiny follow-ups straight to the branch rather than bounce them back to you:

- `set_can_id` now rejects out-of-range ids (logs and returns) instead of clamping — same spirit as your `set_bitrate` validation, and it keeps the documented "id `0` not accepted" honest, since clamping quietly turned `0` into `1`.
- `subscribe_to_can` reads the `working_current` default from the property instead of repeating the `1700` literal.
- Minor docs hard-wrapping and comment tidy.

Approving. 👍

PR #206, opened 2026-04-28
Fix CAN BUS_OFF boot loop caused by ESP-IDF TWAI assertion

felixwrz:
### Motivation

When the CAN bus enters BUS_OFF state (e.g. no peer connected or mismatched baud rate), the ESP32 gets stuck in an infinite crash-reboot loop. ESP-IDF v5.3.1 has a bug where `twai_handle_tx_buffer_frame()` asserts `tx_msg_count >= 0` inside an ISR during BUS_OFF transitions. Since it runs in an interrupt handler, application code cannot catch or prevent it — `abort()` is called, the device reboots, and because the bus is still in the same bad state, it immediately crashes again.

The problem is compounded by the current code: `send()` calls `twai_transmit()` regardless of bus state (queuing more frames into a broken bus), and `reset_can_bus()` uses `twai_initiate_recovery()` which itself triggers the same ISR assertion.

### Implementation

Three-part fix:

1. **Disable assertions** (`CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y` in `sdkconfig.defaults.esp32`) so the ISR assertion no longer calls `abort()`.

2. **Guard `send()`** with a bus state check — silently drop messages when the bus is not in RUNNING state instead of calling `twai_transmit()`, which would pile up errors.

3. **Move BUS_OFF recovery into `step()`** and **replace `twai_initiate_recovery()` with full driver uninstall/reinstall**. The old recovery path could trigger the same assertion, so the fix side-steps it entirely by tearing down and rebuilding the driver. TWAI config structs were promoted to class members in `can.h` so they're available for reinstallation.


falkoschindler:
> _Note: this draft is from Claude based on a quick read of the diff and commit history — a human review is still pending and may revise or override this._

Thanks for digging into this — the diagnosis matches what I see in the code, and fixes 1 (guard `send()` on `RUNNING`) and 2 (uninstall/reinstall instead of `twai_initiate_recovery()`) look solid. Before going into a deeper review, two things I'd like to sort out first:

**1. Scope.** Looking at the commit history, this branch is essentially `new_motor_module` (the merge commits even say so) with the CAN fix added as the last few commits. Of the 54 commits, only three touch the CAN-fix files:

- `02940124` — `add printf-style formatting to bus.send` (the breaking `SerialBus.send()` API change + `echo("%s", ...)` cleanups across several modules)
- `a1fb7932` — `update can recovery`
- `bd8245bd` — `fix CAN BUS_OFF boot loop caused by ESP-IDF TWAI assertion`

Everything else is the new motor / delta arm / wheels work, the Dual Drive docs, the `format` utility, the `inno_*.liz` scratch files, etc. Could you split this into three PRs? Concretely:

1. **new notor feature PR** — base it off `main` at `43db746c` (the last commit before the CAN fix work began). This is the bulk of the work and deserves its own review.
2. **CAN BUS_OFF fix PR** — fresh branch off current `main`, cherry-pick `a1fb7932` and `bd8245bd`. Should be ~125 lines and can land quickly on its own.
3. **SerialBus printf-style `send()` PR** — cherry-pick `02940124` onto its own branch. This is a breaking API change to an existing public module method, so it deserves its own discussion separate from a bug fix.

If the cherry-picks have textual conflicts (likely for `02940124` since it touches files the new motor dev branch also changed), the `echo("%s", ...)` cleanups can ride with the SerialBus PR — they're the same kind of formatting change.

While splitting, please also drop or relocate:

- The `inno_*.liz` files at the repo root (look like scratch/dev scripts; if they're worth keeping, an `examples/` dir would be a better home).
- The stale note in `dual_drive_todo.md`: it says "verify fix: `reset_can_bus()` now handles BUS_OFF via `twai_initiate_recovery()` directly" — but this PR removes `twai_initiate_recovery()`, so the note describes an earlier approach.

**2. `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y`.** This disables _all_ ESP-IDF assertions globally, not just the TWAI ISR one — heap, queues, NVS, scheduler invariants. That's a big surface to give up to work around one driver bug. Two questions before this lands:

- With fixes 1 + 2 in place (no more `twai_initiate_recovery()`, no more `twai_transmit()` while non-running), can you still reproduce the ISR assertion on hardware? If not, fix 3 may not be needed at all.
- If it is still reachable, would `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT` (logs but doesn't `abort()`) work instead? That keeps the checks active everywhere else.

Minor: the `RECOVERING`-state branch in `reset_can_bus()` is dead code now that `twai_initiate_recovery()` is gone — worth removing or commenting why it's kept.

Happy to do a full review on the CAN-fix PR once it's split out and the assertions question is settled.

PR #207, opened 2026-04-28
Fix CAN BUS_OFF boot loop caused by ESP-IDF TWAI assertion

### LINKED ISSUE #205: CAN BUS_OFF state causes infinite crash-reboot loop due to ESP-IDF TWAI assertion in ISR
When the CAN bus enters BUS_OFF state (e.g. no peer connected or mismatched baud rate), the ESP32 gets stuck in an infinite crash-reboot loop.

felixwrz:
### Motivation

When the CAN bus enters BUS_OFF state (e.g. no peer connected or mismatched baud rate), the ESP32 gets stuck in an infinite crash-reboot loop. ESP-IDF v5.3.1 has a bug where `twai_handle_tx_buffer_frame()` asserts `tx_msg_count >= 0` inside an ISR during BUS_OFF transitions. Since it runs in an interrupt handler, application code cannot catch or prevent it — `abort()` is called, the device reboots, and because the bus is still in the same bad state, it immediately crashes again.

The problem is compounded by the current code: `send()` calls `twai_transmit()` regardless of bus state (queuing more frames into a broken bus), and `reset_can_bus()` uses `twai_initiate_recovery()` which itself triggers the same ISR assertion.

### Implementation

Three-part fix:

1. **Guard `send()`** with a bus state check — drop messages with a rate-limited log when the bus is not in RUNNING state instead of calling `twai_transmit()`, which would pile up errors.

2. **Move BUS_OFF recovery into `step()`** — only triggers on the initial BUS_OFF entry transition (not every tick), tracked via a `previous_state` member.

3. **Replace `twai_initiate_recovery()` with full driver uninstall/reinstall** in `reset_can_bus()`. The old recovery path could trigger the same ISR assertion, so the fix side-steps it entirely by tearing down and rebuilding the driver. TWAI config structs (`g_config`, `t_config`, `f_config`) were promoted to class members in `can.h` so they're available for reinstallation.

Additionally, the `recover()` method was removed since `reset()` now handles all recovery, and the docs were updated accordingly.

### Progress


main/modules/can.cpp
```diff
@@ -242,46 +230,39 @@ void Can::reset_can_bus() {
...
             if (twai_get_status_info(&status_info) != ESP_OK) {
                 throw std::runtime_error("failed to get status during recovery");
             }
 
             if (status_info.state != TWAI_STATE_RECOVERING) {
-                echo("Recovery completed, state: %s",
-                     status_info.state == TWAI_STATE_STOPPED   ? "STOPPED"
-                     : status_info.state == TWAI_STATE_RUNNING ? "RUNNING"
-                     : status_info.state == TWAI_STATE_BUS_OFF ? "BUS_OFF"
-                                                               : "UNKNOWN");
```
falkoschindler, line 272: Why did you remove this output? Seems unrelated to the original problem.

main/modules/can.cpp
```diff
@@ -242,46 +230,39 @@ void Can::reset_can_bus() {
...
-
-    if (status_info.state == TWAI_STATE_BUS_OFF) {
-        if (twai_initiate_recovery() != ESP_OK) {
-            throw std::runtime_error("could not initiate recovery");
-        }
-
+    } else if (status_info.state == TWAI_STATE_RECOVERING) {
         const unsigned long start_time = millis();
-        const unsigned long timeout_ms = 500;
-
```
falkoschindler, line 261: Can we keep this constant to reduce the diff?

main/modules/can.cpp
```diff
@@ -242,46 +230,39 @@ void Can::reset_can_bus() {
...
             if (status_info.state != TWAI_STATE_RECOVERING) {
-                echo("Recovery completed, state: %s",
-                     status_info.state == TWAI_STATE_STOPPED   ? "STOPPED"
-                     : status_info.state == TWAI_STATE_RUNNING ? "RUNNING"
-                     : status_info.state == TWAI_STATE_BUS_OFF ? "BUS_OFF"
-                                                               : "UNKNOWN");
                 break;
             }
-
-            if (millis_since(start_time) > timeout_ms) {
```
falkoschindler, line 276: If we keep the constant, this diff can be reverted as well.

falkoschindler (commented):
@felixwrz This PR seems to contain some unrelated changes (see my comments below). In 364fa9ed117b20e797f09e879ebc817f43d06404 I already reverted one to keep the code simple. Do you agree in reverting the other ones as well? Or are they intentional?

main/modules/can.cpp
felixwrz: When the CAN bus has no peer (e.g. motor driver is off, e-stop triggered), the bus immediately goes back to BUS_OFF after each reset. Since step() runs every loop iteration, it calls reset_can_bus() repeatedly, and the echo() at the end floods the serial output with "CAN bus reset successful" messages. That's why the echo was removed, not as an unrelated cleanup, but to stop spamming the console in this common scenario.

falkoschindler (commented):
Ok, there are four open questions remaining. Maybe you can have a look:

1. **`CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE=y` is a very heavy hammer**   The new line in `sdkconfig.defaults.esp32` disables `assert()` globally across the **entire firmware**, not just TWAI. This defines `NDEBUG` and removes every `assert()` in ESP-IDF and your own code. Concerns:
   - You lose asserts as a defense everywhere (BLE, FreeRTOS, your own modules), to silence one ISR bug.
   - It's only set for `esp32`, not for `esp32s3`. If S3 ever uses TWAI it remains vulnerable; if it doesn't, why is the asymmetry intentional?
   - If part 3 (full reinstall) genuinely sidesteps the buggy ISR path, is the global disable still needed? Worth confirming whether you saw the assertion fire even with the new recovery in place, or whether it's belt-and-suspenders.
   - Suggestion: try dropping the sdkconfig change first. If the assertion still fires (e.g. on frames already queued when BUS_OFF hits, before `step()` can react), fall back to `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT=y` rather than `DISABLE` — same boot-loop fix, but asserts still log so you keep the signal everywhere else.

2. **`recover` method still calls `twai_initiate_recovery()` directly**

   The `recover` branch in `Can::call` still invokes `twai_initiate_recovery()` directly — the exact function the PR description identifies as buggy. With assertions disabled this won't `abort()`, but it's inconsistent with the rest of the fix. Either route `recover` through `reset_can_bus()` or remove it.

3. **`send()` now silently drops on non-RUNNING state**

   In the new `Can::send`, the previous code threw on transmit failure; the new code never throws and never even logs. Two effects:
   - Callers like the SDO write retry loop in `canopen_motor.cpp` used to retry off the send-thrown exception; now they only learn about failure indirectly via the `wait_for_sdo_writes(100)` timeout, adding ~100ms of latency to each retry. Functional but slower.
   - Silent drops with zero log are hard to diagnose in the field. Consider at minimum a rate-limited `echo` or incrementing a dropped-message counter (you already expose `tx_failed_count` etc.).

   Suggested patch (max one log per second; `last_drop_log = 0` keeps boot-time silent):

   ```cpp
   void Can::send(const uint32_t id, const uint8_t data[8], const bool rtr, uint8_t dlc) const {
       twai_status_info_t status_info;
       if (twai_get_status_info(&status_info) != ESP_OK || status_info.state != TWAI_STATE_RUNNING) {
           static unsigned long last_drop_log = 0;
           if (millis_since(last_drop_log) > 1000) {
               echo("CAN send dropped: bus not RUNNING");
               last_drop_log = millis();
           }
           return;
       }
       // ...
   }
   ```

4. **Re-entry storm when the bus stays in `BUS_OFF`**

   If the underlying cause persists (no peer, wrong baud), every `step()` will re-trigger `reset_can_bus()`. After reinstall the driver returns to `RUNNING`, errors climb back to 256, transitions to `BUS_OFF`, resets again — burning CPU cycles and flooding `echo` with "CAN bus state before reset" / "CAN bus reset successful" log lines on every cycle. Consider throttling (e.g. minimum interval between resets) or only logging on state transitions.

   Suggested patch — gate the reset on the BUS*OFF \_entry* transition so we don't reset every step. If reset succeeds the driver becomes RUNNING; if the bus immediately re-enters BUS_OFF that's a new transition and we'll retry naturally. Add `twai_state_t previous_state = TWAI_STATE_RUNNING;` as a private member of `Can` and update `step()`:

   ```cpp
   if (status_info.state == TWAI_STATE_BUS_OFF && this->previous_state != TWAI_STATE_BUS_OFF) {
       try {
           this->reset_can_bus();
       } catch (const std::exception &e) {
           echo("CAN recovery failed: %s", e.what());
       }
   }
   this->previous_state = status_info.state;
   ```

   Tradeoff to consider: if `reset_can_bus()` itself throws, the gate above won't retry until state changes again. Minor concern in practice (a systemic reset failure usually won't be fixed by retrying every 10 ms), but if you want belt-and-suspenders, only update `previous_state` on success: `this->previous_state = reset_succeeded ? TWAI_STATE_RUNNING : status_info.state;`.

JensOgorek:
What is the state with this @felixwrz ?

felixwrz:
@falkoschindler All four points from your review have been addressed:

1. **Global assertions disable removed** (`4a79e93`) — `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE` is no longer in `sdkconfig.defaults.esp32`.
2. **`recover` routed through `reset_can_bus()`** (`711b3c2`) — no longer calls `twai_initiate_recovery()` directly.
3. **Rate-limited drop log in `send()`** (`da147dd`) — logs "CAN send dropped: bus not RUNNING" at most once per second.
4. **BUS_OFF entry transition gating** (`2e2f904`) — `reset_can_bus()` only fires on the BUS_OFF *entry* transition via `previous_state`, preventing the re-entry storm.

I will perform hardware tests on it later this day.

falkoschindler:
@felixwrz Great! Two final thoughts:

- Now the two methods `recover` and `reset` do exactly the same. This isn't clear from the module reference. I'd suggest to drop one of them or - to preserve backward compatibility - at least mark one of them as deprecated.
- Please update the PR description now that `CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE` has been removed. Thanks!

falkoschindler (approved):

PR #209, opened 2026-05-05
Add DualDriveMotor and DualDriveWheels modules

JensOgorek:
### Motivation

Add Lizard support for a CAN-based dual-track drive (two motors with hall feedback) so it can be used through the standard Lizard module interface.

This PR is a minimal slice with only the driving parts (base class, drive motor, wheels) and is part of the #194.

### Implementation

Three new modules under `main/modules/`:

- **`DualDriveMotorBase`** — shared base for the dual-track drive over CAN. Handles common state (voltage, temperature, error codes, switch state, firmware version), enable/disable, and the `Configure` (0x0B) and `SwitchState` (0x0A) CAN commands.
- **`DualDriveMotor`** — drive-mode specialization. Auto-switches the controller into drive mode (`Configure 0x02 / 0xA5A5`) on construction. Implements the `Motor` interface so it can be used in shared abstractions (e.g. odometry). Exposes:
  - `motor.speed(vel)` — target angular velocity (rad/s) via SpeedCmd 0x01
  - `motor.drive_ticks(vel, ticks)` — relative move via DriveTicksCmd 0x04
  - `motor.position` — accumulated position (m), derived from the wrapping int16 hall counter and the configurable `m_per_rad`
  - properties: `m_per_rad`, `reversed`, `rad_limit`, per-motor currents
- **`DualDriveWheels`** — combines two `DualDriveMotor` instances for differential steering. Provides `wheels.speed(linear, angular)` and a new `wheels.drive(speed, distance)` for straight-line relative moves (uses each motor's `m_per_rad` and the fixed 600 ticks/rev for tick conversion).

Notes on the firmware contract:

- Currently only the G350 motor (600 hall ticks/rev) is supported in drive mode; this is hardcoded as `MOTOR_TICKS = 600`.
- Acceleration / jerk parameters of SpeedCmd 0x01 are intentionally left out for now: the firmware does not evaluate them yet. Tracked in the Motortreiber TODO inbox so we can re-add them once the firmware catches up.
- `position(target, ...)` (closed-loop absolute position) is not supported in drive mode and is stubbed accordingly.

Module registration follows the same pattern as `DunkerMotor` / `DunkerWheels`.

### Progress

- [x] Tested on hardware.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -1017,18 +1017,18 @@ Currently only the G350 motor (600 hall ticks per revolution) is supported in th
 | `motor.enabled`     | Whether the motor is enabled                      | `bool`    |
 | `motor.debug`       | Enable CAN debug output                           | `bool`    |
 
-| Methods                                       | Description                       | Arguments                     |
-| --------------------------------------------- | --------------------------------- | ----------------------------- |
-| `motor.speed(vel[, acc, jerk])`               | Set target angular velocity rad/s | `float`\[, `float`, `float`\] |
-| `motor.drive_ticks(vel, ticks)`               | Relative move in hall ticks       | `float`, `int`                |
-| `motor.switch_state(state)`                   | Set state: 1=off, 2=brake, 3=on   | `int`                         |
-| `motor.configure(setting_id, value1, value2)` | Send raw configure command        | `int`, `int`, `int`           |
-| `motor.configure_node_id(new_id)`             | Set CAN node ID                   | `int`                         |
-| `motor.on()`                                  | Turn motor on                     |                               |
-| `motor.off()`                                 | Turn motor off                    |                               |
-| `motor.stop()`                                | Brake the motor                   |                               |
-| `motor.enable()`                              | Enable the motor                  |                               |
-| `motor.disable()`                             | Disable the motor                 |                               |
+| Methods                                       | Description                       | Arguments           |
+| --------------------------------------------- | --------------------------------- | ------------------- |
+| `motor.speed(vel)`                            | Set target angular velocity rad/s | `float`             |
+| `motor.drive_ticks(vel, ticks)`               | Relative move in hall ticks       | `float`, `int`      |
+| `motor.switch_state(state)`                   | Set state: 1=off, 2=brake, 3=on   | `int`               |
+| `motor.configure(setting_id, value1, value2)` | Send raw configure command        | `int`, `int`, `int` |
+| `motor.configure_node_id(new_id)`             | Set CAN node ID                   | `int`               |
+| `motor.on()`                                  | Turn motor on                     |                     |
+| `motor.off()`                                 | Turn motor off                    |                     |
+| `motor.stop()`                                | Brake the motor                   |                     |
+| `motor.enable()`                              | Enable the motor                  |                     |
+| `motor.disable()`                             | Disable the motor                 |                     |
 
 ## Dual Drive Wheels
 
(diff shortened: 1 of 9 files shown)


falkoschindler:
@JensOgorek Here is the result of a first AI-assisted review. Let's address these findings first before I do another thorough review:

1. **`motor.speed(vel)` unit inconsistency with sibling motors _(important)_**

   `DualDriveMotor::call` for `speed` takes **rad/s** ([dual_drive_motor.cpp:118-122](main/modules/dual_drive_motor.cpp#L118-L122)), while every other motor's DSL `speed(...)` takes **m/s** (cf. [dunker_motor.cpp:118-120](main/modules/dunker_motor.cpp#L118-L120) → `m_per_turn`-scaled). At the same time `wheels.speed(linear, angular)` correctly uses the m/s `Motor::speed(double, double)` overload at [dual_drive_wheels.cpp:48-49](main/modules/dual_drive_wheels.cpp#L48-L49), so within the same PR `speed` means different things at the motor level vs. wheels level.

   This is a footgun: a user who has been writing `motor.speed(0.5)` against a `DunkerMotor` would unknowingly drive 0.5 rad/s on a `DualDriveMotor`. Suggest either:
   - Align the DSL method to take m/s (apply `m_per_rad` inside `call("speed")`, same shape as the C++ `speed(double, double)` override) (preferred). Or
   - Rename the rad/s entry point (e.g. `speed_rad`, `angular_speed`) and make `speed` m/s.

2. **`int16` overflow in `drive_meters` _(correctness)_**

   [dual_drive_motor.cpp:107-112](main/modules/dual_drive_motor.cpp#L107-L112):

   ```cpp
   const int16_t ticks = static_cast<int16_t>(distance / m_per_rad * MOTOR_TICKS / (2.0f * M_PI));
   ```

   With the default `m_per_rad = 1.0` and 600 ticks/rev, the int16 ceiling is reached at roughly `distance ≈ 343 m` (`32767 * 2π / 600`). For configurations with smaller `m_per_rad` (typical for geared tracks), the safe distance shrinks proportionally. The wrap is silent: `wheels.drive(0.5, 400)` would behave as a much shorter — possibly negative — relative move.

   Suggest at minimum clamping with a warning, or splitting into multiple commands when out of range. A simple `if (std::abs(...) > INT16_MAX) echo("warning: ...")` would already prevent silent surprises.

3. **Empty `position()` stub _(style)_**

   [dual_drive_motor.cpp:172-174](main/modules/dual_drive_motor.cpp#L172-L174) silently no-ops a closed-loop position request. A user who reaches it via shared-abstraction code (it isn't exposed in `call`) would get nothing back and no clue why. Either log a one-time warning or throw — the current "silent no-op" is hard to debug.

4. **Constructor-time CAN write _(behavior)_**

   [dual_drive_motor.cpp:25](main/modules/dual_drive_motor.cpp#L25) sends `Configure(0x02, 0xA5A5, 0)` from the constructor. TX-only is fine, but if a frame is dropped (bus contention, controller still booting) there's no ack, retry or echo, and the controller may stay in the previous mode silently. Worth confirming this is acceptable on real hardware, or adding a small retry-on-status-mismatch in `step()` once `state` is observed.

5. **PR description doesn't surface the concrete sibling _(docs)_**

   The PR body says "part of #194", but #194 is a stub for an unrelated new motor — which made me doubt the `DualDriveMotorBase` split was justified. The actual concrete sibling is the stacked draft #210 (`DualDriveDeltaMotor`), which inherits from `DualDriveMotorBase` and does pay back the abstraction. A one-line link to #210 in the description would have made the split self-evident; consider adding it.

JensOgorek:
Closing — these modules target hardware that isn't publicly available, so they're out of scope here. They now live in a private integrator project as self-registering components, no core changes needed.

PR #211, opened 2026-05-07
Make modules self-registering

falkoschindler:
### Motivation

Adding a new module currently requires editing two open-source files (`ModuleType` enum in `modules/module.h` and the if/else chain in `Module::create()`), which makes out-of-tree modules — e.g. closed-source hardware drivers shipped in a private repo — impossible without forking or patching. It also means the 350-line if/else chain in `Module::create()` keeps growing, and the `// TODO: rmd_motor, roboclaw_motor` line in `MotorAxis` is wedged behind enum-based dispatch that can't be cleanly extended.

This PR refactors module registration so each module's `.cpp` is the single source of truth for its DSL name, factory, and defaults, with no central edit required. A future closed-source motor in a separate component just needs to call `REGISTER_MODULE(...)` once; nothing in the public repo has to change.

### Implementation

- **`Module` base class** (`main/modules/module.h`, `module.cpp`):
  - `ModuleType` enum dropped; `Module::type` is now `std::string`.
  - New `REGISTER_MODULE(ClassName, &factory_fn)` macro, combining the old `REGISTER_MODULE_DEFAULTS` with factory registration. Same construct-on-first-use static map idiom — no static-init-order surprises. Throws on duplicate registration.
  - `Module::create()` collapses to a registry lookup; the previous 350-line if/else chain disappears.

- **Per-module factories**: each of ~30 module `.cpp` files gets a small static `create_X(name, args, message_handler)` factory function and one `REGISTER_MODULE` line replacing the old `REGISTER_MODULE_DEFAULTS`. Constructors now pass the type-name string to `Module(...)` instead of an enum value.

- **`module_helpers.h`** (new): `get_module_argument<T>(arg, "TypeName")` template that combines the previous `Global::get_module` + type check + `static_pointer_cast` pattern (and accepts `Proxy` as a stand-in, matching today's behaviour). Also hosts the I2C `DEFAULT_*_PIN` macros that used to live in `module.cpp`.

- **`MotorAxis` dispatch** (`motor_axis.cpp`): switched from `module->type == odrive_motor` + `static_pointer_cast` to `std::dynamic_pointer_cast<Motor>(module)`. This required enabling `CONFIG_COMPILER_CXX_RTTI=y` in `sdkconfig.defaults.{esp32,esp32s3}`. Binary-size cost: **+10,016 B (+1.5%)** for ESP32, **+11,520 B (+1.6%)** for ESP32-S3 — both targets still 65–66 % free in the app partition. The win is that any future motor type plugs in by simply inheriting `Motor`; no opt-in method or central edit needed.

- **Type comparisons in `main.cpp` / `bus_backup.cpp`**: enum-based `module->type != serial_bus` style checks converted to `module->type != "SerialBus"` etc.

#### Side findings (worth flagging in review)

- `MotorAxis` was declaring a local `name` that shadowed the `Module::create()` parameter, so the axis ended up tagged with the inner motor's name in `echo()` messages and broadcasts. The local is now `motor_name`. Small behaviour change, almost certainly an improvement.
- `LinearMotor::LinearMotor` was passing `Module(output, name)` (wrong type tag — copy/paste from `Output`). Now correctly `"LinearMotor"`.

#### Out of scope

- **`RmdMotor` / `RoboClawMotor` × `MotorAxis`**: the existing `// TODO: rmd_motor, roboclaw_motor` is preserved (RmdMotor has API signature drift; RoboClawMotor is missing `stop`/`position`/`get_speed` entirely and may not even support absolute positioning at the protocol level). The new dispatch fails gracefully for both with the same error users see today. These are independent follow-ups.

#### Verification

Builds clean for ESP32 and ESP32-S3 with no warnings. An algorithmic check extracted each pre-refactor `Module::create()` branch and each new `create_X` body, normalised them (substituting `enum_name` → `"ClassName"` and the old multi-line dependency-lookup pattern → `get_module_argument<T>(...)`), and diffed: 1/30 byte-identical, 29/30 differ only in expected categories (rename, dropped redundant arg, added `const`, `auto`, `if/else` → early-return). The MotorAxis name-shadowing fix and the `LinearMotor` type-tag fix were surfaced this way.

### Progress

- [ ] Tested on hardware.
- [x] Test building with a private hardware module.


JensOgorek:
Finally found some time to actually test this — really like the direction. While going through it a few ideas came up that I'd like to float before we land it, because they'd mean another sweep through the PR rather than something I'd tack on after the fact.

## 1. `static TYPE` constant instead of an instance `type` string

Right now every module instance carries a `const std::string type` set from the constructor. What if each class instead declared

```cpp
class Analog : public Module {
public:
    static inline constexpr const char *TYPE = "Analog";
    ...
};
```

and the base constructor became `Module(name)` instead of `Module("Analog", name)`?

Three reasons I think it's worth it:

- One `std::string` allocation per module instance less. Small, but free.
- The DSL name becomes an explicit, reviewable declaration. With `#class_name` in the macro, the C++ identifier and the DSL identifier are the same symbol — rename `Analog` to `AnalogPin` and every Lizard script in the wild silently breaks. An explicit `TYPE` constant makes that a deliberate, visible change.
- `REGISTER_MODULE(Analog, &create_analog)` then keys off `Analog::TYPE` instead of stringifying, which lines the macro up with the rest of the registration.

## 2. `get_module_argument<T>(arg)` without the type-string parameter

Today:
```cpp
get_module_argument<AnalogUnit>(arguments[0], "AnalogUnit")
```

What if the helper internally did `std::dynamic_pointer_cast<M>(module)` and used `M::TYPE` for the error message, so the call site collapses to:
```cpp
get_module_argument<AnalogUnit>(arguments[0])
```

The Proxy stand-in behaviour can stay as-is. The win: the type is named only once, no risk of the template parameter and the string drifting apart at some call site.

## 3. `dynamic_pointer_cast` everywhere we're checking type

`bus_backup.cpp` and `main.cpp` (the Expander check when constructing a Proxy) still do `module->type == "..."` comparisons. Since RTTI is on for `MotorAxis` anyway, those could go to `dynamic_pointer_cast` too — saves the `static_pointer_cast` that usually follows the type check, and is consistent with how `MotorAxis` already dispatches.

The shadow path in `Module::call("shadow", …)` would have to come along as well: `typeid(*this) != typeid(*target_module)` once the instance `type` field is gone, which is arguably the more honest check anyway.

## 4. `WHOLE_ARCHIVE` is doing important work and probably deserves a doc note

If I'm reading the CMake right, the public `main` component needs `WHOLE_ARCHIVE` for this to work at all: the `REGISTER_MODULE` static initializer lives in an anonymous namespace and therefore has no externally visible symbol. Without `WHOLE_ARCHIVE` the linker is free to drop the entire `.o` as "unused" — build stays green, the static init never runs, and at runtime you get `unknown module type "ClassName"`.

It's the classic foot-gun for this pattern, and it equally bites any out-of-tree component that uses the macro. Probably worth a paragraph in `AGENTS.md` so a private-component author doesn't lose an afternoon to it.

## 5. Small consequences if we go this way

- Each module header gets `static inline constexpr const char *TYPE = "...";` under `public:`.
- The AGENTS.md example would need to reflect the new pattern (TYPE constant, `Module(name)` ctor, `get_module_argument` without the string arg).

---

The two I'd most like your take on are **TYPE constant instead of `#class_name` stringify** and **`get_module_argument` losing the explicit type string** — both of those would mean another pass over the per-module changes already in this PR, which is why I'd rather ask before doing it. The `WHOLE_ARCHIVE` note feels worth doing regardless.

falkoschindler:
Thanks Jens — all five landed in d7ec081→HEAD on this branch.

1. **TYPE constant**: each module header now declares `static inline constexpr const char *TYPE = "ClassName"`. `REGISTER_MODULE` keys off `class_name::TYPE` instead of stringifying. C++ rename and DSL identifier are now separate decisions.
2. **`get_module_argument<T>(arg)` without the type string**: helper does `std::dynamic_pointer_cast<T>(module)`, falls back to `dynamic_pointer_cast<Proxy>` for the remote-stand-in case (unchanged behaviour), and uses `T::TYPE` for the error message. Drift is impossible now.
3. **`dynamic_pointer_cast` everywhere**: `bus_backup.cpp` and the Expander check in `main.cpp` are converted. Shadow path now uses `typeid(*this) != typeid(*target_module)`. `Module::type` is deleted — there's no instance type-string field anymore.
4. **`WHOLE_ARCHIVE` doc**: AGENTS.md `Module Registration` pitfall has a new subsection covering it. Confirmed by working on the integrator-pattern demo repo (`lizard-zz`) — without `WHOLE_ARCHIVE` on the private component the closed module's static initializer is silently dropped and Lizard reports `unknown module type "Demo"`. Easy afternoon-killer for a private-component author, as predicted.
5. **Consequences**: TYPE constants present on every public module class; AGENTS.md "Adding a New Module" example refreshed to match.

Tested clean for ESP32 and ESP32-S3. Binary size delta vs the pre-followup commit on this branch: +480 B (ESP32) / +512 B (ESP32-S3) — essentially noise from the extra dynamic_pointer_cast call sites.

Ready for another look.

JensOgorek:
Thanks @falkoschindler  — TYPE / RTTI / `get_module_argument` all look great. But: just flashed `0acea57` on a test ESP32, startup script with `Output(15)` dies at boot with:

```
error while loading startup script: unknown module type "Output"
```

Same DCE bug as for out-of-tree components — `REGISTER_MODULE`'s registrar in `output.cpp` is in an anonymous namespace, nothing outside the TU references it, linker drops `output.cpp.o` from `libmain.a`. Your commit message assumes ESP-IDF whole-archives `main` by default; in v5.3.1 it doesn't.

Your lizard-zz test didn't catch it because `Demo` lives in `lizard_private` (which already has `WHOLE_ARCHIVE`); the main lizard modules were never actually instantiated. Repro it there by swapping `Demo()` for `Output(15)` in the startup script.

Fix: add `WHOLE_ARCHIVE` to `main/CMakeLists.txt`. Verified locally — script runs, GPIOs toggle. The AGENTS.md out-of-tree note is right, but should drop the "in-tree is safe" framing since the same applies to main. Happy to push the CMakeLists.txt + AGENTS.md tweak if you'd rather review than write.

(Aside: `void Module::Module::expect(...)` in `module.cpp:29` — pre-existing typo, easy follow-up.)

falkoschindler:
Good catch, thank you — confirmed and fixed in 937d489.

You diagnosed it exactly. `main/CMakeLists.txt` was missing `WHOLE_ARCHIVE`; the linker was dropping every `REGISTER_MODULE` static initializer from `libmain.a`. `nm` on the pre-fix ELF found zero `RegisterModule_*` symbols. With the flag added: 60 of them (30 modules × C1/C2 each).

Embarrassingly, the original commit was verified only by `nm`-ing `Demo` symbols in the lizard-zz build, where the private component had `WHOLE_ARCHIVE` already. Never flashed the standalone Lizard build. The "in-tree is safe by default" framing in AGENTS.md was wrong and is now removed — the doc just says `WHOLE_ARCHIVE` is mandatory on both `main` and any out-of-tree component using `REGISTER_MODULE`.

Correction on the size delta I quoted earlier — my "+10 KB / +1.5%" claim was numerically close but methodologically bogus: I compared two broken builds against each other (both were missing the registrar code). Re-measured against pre-refactor `main` (d6c46e9):

- ESP32:    1,273,984 B → 1,287,040 B (**+13,056 B / +1.0%**)
- ESP32-S3: 1,266,288 B → 1,279,168 B (**+12,880 B / +1.0%**)

About 13 KB per target, essentially the RTTI machinery cost. Pre-refactor had every module linked via the `Module::create()` if/else chain referencing each one; post-refactor has them linked via `WHOLE_ARCHIVE`. Same net effect on what's in the binary; the ~1% delta is just `dynamic_pointer_cast` + `typeid`. Both targets still 37% free in the 2 MB partition.

Also fixed the `Module::Module::expect` stutter typo while in the area.

JensOgorek (approved):
Now it works

PR #214, opened 2026-06-11
Add scheduled commands: `at <time> do <actions> end`

falkoschindler:
### Motivation

Hosts streaming time-critical sequences (e.g. trajectory setpoints) need actions to fire at precise times, decoupled from the ~100 Hz main loop. This PR adds a new statement that schedules a one-shot block of actions at a given time in milliseconds since boot (matching `core.millis`):

```
at 125015 do en3.on() end
at core.millis + 500 do en3.off() end
```

### Implementation

- **Grammar**: new `schedule_definition` statement in `language.owl`; `parser.h` regenerated. `at` and `do` become reserved keywords.
- **Scheduler** (`utils/scheduler.{h,cpp}`): min-heap of (deadline µs, sequence, routine); a single one-shot `esp_timer` always armed at the earliest deadline notifies a dedicated task (8 KB stack, priority 10, core 0) that drains all due entries and re-arms. The task blocks on a notification between deadlines.
- **Interpreter mutex** (`utils/interpreter_lock.{h,cpp}`): recursive FreeRTOS mutex with priority inheritance behind an RAII guard, acquired per module step, per rule/routine step and around command processing. This bounds scheduler fire latency to one module step — and fixes a pre-existing unprotected race where BLE commands called `process_lizard` from the NimBLE host task.

Semantics and design decisions:

- The time expression is evaluated once, at definition time.
- One-shot: each block fires once and is discarded; past deadlines execute immediately.
- Equal deadlines fire in definition order (FIFO tiebreaker).
- `await` is not allowed inside scheduled blocks (compile error); calling a routine that itself awaits is fine — the routine starts at the scheduled time and continues at loop rate.
- Capacity is capped at 100 pending entries; further `at` statements fail with "schedule is full" as a flow-control signal to the host.
- New `core.clear_schedule()` discards all pending entries, e.g. in safety rules: `when core.last_message_age > 500 then motor.stop(); core.clear_schedule() end`
- With `core.debug = true`, each firing echoes its lateness (e.g. `at 125015: fired 0.3 ms late`).

### Progress

- [x] Tested on hardware.

Hardware verification on ESP32: `at core.millis + 1000 do ... end` fired 0.3 ms late (target ≤1 ms); past deadlines fire immediately; equal deadlines fire in definition order; `await` in a scheduled block is rejected at compile time while routine calls with awaits work; 105 streamed entries produce 5× "schedule is full"; `core.clear_schedule()` cancels pending blocks; 2×300 blocks streamed over 60 s with no errors and stable heap. Both esp32 and esp32s3 build clean (idf v5.3.1).


JensOgorek:
Reviewed the code and tested the feature on an ESP32 (`v0.11.0-19-g00d6e80`). Looks solid — all the documented behavior holds.

### Hardware results

| Check | Result |
|-------|--------|
| Past deadline fires immediately | ✅ |
| `at core.millis + 4000` | ✅ fired 0.3 ms late |
| Equal deadlines fire in definition order (FIFO) | ✅ A/B/C at 0.2 / 1.6 / 2.8 ms late |
| Scheduled block calling an awaiting routine | ✅ runs at scheduled time, finishes at loop rate |
| `core.clear_schedule()` drops pending blocks, scheduler keeps working | ✅ |
| `await` inside a block | ✅ rejected at compile time |
| Non-number time | ✅ rejected at compile time |
| 105 entries streamed (cap 100) | ✅ exactly 5× `schedule is full` |

On the concurrency side the per-step locking looks sound, and concurrent inserts into the `Global` containers are safe because they're node-based (`std::map` / `std::list`). The big `parser.h` diff vs. a local `owl` regen is only non-deterministic state renumbering, not a functional change.

### Two small notes (not blockers)

- **`core.millis` in a startup script reads `0`** (the core module hasn't stepped yet), so `at core.millis + X` in startup schedules `X` ms on the *boot clock*, not "X ms from now". Pre-existing `core.millis` behavior, but easy to trip over with `at` — might be worth a doc line.
- **The `schedule is full` signal only shows once per submission.** Sending 105 `at` lines as a *single* batch yields just one error (the first overflow aborts the rest of that submission); the expected 5× only appears when the lines are *streamed* one per command — which is the realistic host case.

### Test scripts

<details>
<summary>Normal — should run cleanly</summary>

```
# 01 — timing, past deadline, FIFO, and a block calling an awaiting routine.
# In a startup script core.millis reads 0, so times are on the boot clock;
# offsets >= 4000 stay in the future. Lateness then < 1 ms.
core.debug = true

at core.millis - 500 do core.print("past: immediate") end       # past -> fires now
at core.millis + 4000 do core.print("timing: fired") end        # ~4 s later

int target = core.millis + 5000                                 # equal deadlines -> FIFO
at target do core.print("fifo: A") end
at target do core.print("fifo: B") end
at target do core.print("fifo: C") end

let r do core.print("routine: start"); await core.millis > 0; core.print("routine: done") end
at core.millis + 6000 do r() end                                # block may call an awaiting routine
```

```
# 02 — core.clear_schedule() drops ALL pending blocks; scheduler still works after.
core.debug = true
at core.millis + 4000 do core.print("cleared: should NOT fire") end
at core.millis + 4000 do core.print("cleared: should NOT fire either") end
core.clear_schedule()
at core.millis + 4000 do core.print("after clear: SHOULD fire") end
```
</details>

<details>
<summary>Should fail — on purpose</summary>

```
# Expect compile error: "await is not allowed in scheduled blocks"
at core.millis + 100 do await core.millis > 0 end
```

```
# Expect compile error: "schedule time must be a number"
at "soon" do core.print("should not happen") end
```

```
# Capacity: stream 105 of these as separate commands (not one batch).
# Expect 5x "schedule is full" (cap is 100); clear with core.clear_schedule() after.
at core.millis + 60000 do core.print("entry") end
# ... repeated 105 times
```
</details>

falkoschindler:
@JensOgorek Thanks for the positive feedback! So are you approving? Or is anything missing?

JensOgorek (approved):

JensOgorek:
@falkoschindler yes, I forgot to hit approved :)

PR #215, opened 2026-06-16
Add automatic code-quality enforcement (clang-format + pre-commit CI)

evnchn:
*Drafted by Evan (@evnchn) with Claude Code (Opus 4.8). Opened as a draft — see "Known limitation" below.*

### Motivation

Lizard documents detailed C++ standards in CONTRIBUTING.md (ESP-IDF style, 4-space indent, naming, brace style), but **nothing enforces them automatically**: `.pre-commit-config.yaml` only has generic hooks (whitespace / EOF / mdformat) with no C/C++ tooling, and the sole CI workflow (`release.yml`) runs no checks. So on any PR where the contributor hasn't installed the hooks locally, code quality is unverified. This addresses the "automatic enforcement of code quality" wishlist item.

### Implementation

- **`.clang-format`** encoding the CONTRIBUTING.md C++ style, tuned so it formats the **entire existing codebase with zero changes** — adopting clang-format introduces **no reformatting diff**.
  - Verified: `clang-format -i` over `main/**` yields an empty diff for clang-format **20.1.8, 21.x and 22.1.5**. The pinned version matters — 18.x / 19.x are *not* zero-diff (9 files change). The hook is pinned to the latest verified version, **`v22.1.5`** (the current `mirrors-clang-format` tag).
- **`.clang-format-ignore`** + a hook `exclude` for generated (`main/parser.*`) and vendored (`main/modules/BNO055ESP32.*`) files.
- **clang-format hook** added to `.pre-commit-config.yaml` (`mirrors-clang-format` `v22.1.5`).
- **`.github/workflows/check.yml`** runs `pre-commit` on every PR/push — finally enforcing both the new clang-format hook **and** the pre-existing hooks, which until now ran only locally.

Verified locally: `pre-commit run --all-files` passes every hook with **zero** changes to the tree.

### Known limitation (why this is a draft)

To get the zero-diff property, `.clang-format` uses **`ColumnLimit: 0`** (no line re-flowing). The trade-off: clang-format **does not enforce the 120-column limit** from CONTRIBUTING.md. Any `ColumnLimit > 0` makes clang-format re-wrap lines to its own width-optimal shape, churning ~550 lines across 73 files — mostly *joining* lines that were deliberately split (constructor init-lists, empty bodies), not fixing real violations. (For reference, the tree already has ~155 lines over 120, so strict 120 enforcement isn't free under any approach.)

If you want 120 enforced, the clean follow-up is a **width-only check** (e.g. `editorconfig-checker` or a small pre-commit `pygrep` rule) that flags long lines without reformatting — happy to add it, but it needs a decision on the ~155 existing over-120 lines (fix vs. grandfather), so it's intentionally out of scope here.

Kept as a draft so you can try it and tune the config before it lands.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,14 @@
+# Encodes the C++ style documented in CONTRIBUTING.md.
+# Chosen to format the existing codebase with ZERO changes (verified zero-diff for clang-format
+# 20.1.8 through 22.1.5; the pre-commit hook is pinned to the latest, 22.1.5).
+# NOTE: ColumnLimit is 0 (no line re-flowing) so adopting clang-format does not churn the tree;
+# the 120-column rule from CONTRIBUTING.md is therefore NOT enforced here (see that file).
+BasedOnStyle: LLVM
+IndentWidth: 4
+ColumnLimit: 0
+AccessModifierOffset: -4
+BreakBeforeBraces: Attach
+PointerAlignment: Right
+SortIncludes: false
+AllowShortFunctionsOnASingleLine: Inline
+ReflowComments: false
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,5 @@
+# Files clang-format must not touch.
+main/parser.c              # generated from language.owl via gen_parser.sh
+main/parser.h              # generated from language.owl via gen_parser.sh
+main/modules/BNO055ESP32.cpp   # vendored third-party driver (upstream style)
+main/modules/BNO055ESP32.h     # vendored third-party driver (upstream style)
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,5 +1,9 @@
-# Files clang-format must not touch.
-main/parser.c              # generated from language.owl via gen_parser.sh
-main/parser.h              # generated from language.owl via gen_parser.sh
-main/modules/BNO055ESP32.cpp   # vendored third-party driver (upstream style)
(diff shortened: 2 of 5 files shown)


falkoschindler (approved):
Nice addition — this fills a real gap (the C++ standards in CONTRIBUTING.md had no automated enforcement). The zero-diff `.clang-format`, the version pin rationale, and wiring `pre-commit` into CI all look solid. I went ahead and pushed a few small follow-ups on top:

- **Fixed `.clang-format-ignore`:** the trailing inline comments (`main/parser.c   # generated…`) were silently breaking every pattern — clang-format treats the whole line as the glob (only lines starting with `#` are comments), so none of the four files were actually being ignored on direct `clang-format -i` runs. Moved each comment onto its own line.
- **Removed duplication:** the excluded-file list lived in both `.clang-format-ignore` and the hook's `exclude` regex, and the C++ style was re-encoded inline in `.vscode/settings.json` (`C_Cpp.clang_format_fallbackStyle`). Now `.clang-format` / `.clang-format-ignore` are the single source of truth — clang-format honors the ignore file in the hook and on direct runs, and cpptools picks up the root `.clang-format`.
- **Bumped CI actions:** `actions/checkout` v4→v7, `actions/setup-python` v5→v6.

The `ColumnLimit: 0` trade-off is fine by me — I'd rather not commit to strictly enforcing the 120-column limit, so leaving line width unenforced for now is the right call. LGTM. 👍

PR #216, opened 2026-06-17
Deadline-based main loop (sleep the remainder of the 10 ms budget)

### LINKED ISSUE #213: Main loop is throttled by blocking console writes
## Summary

The firmware's main-loop period is implicitly coupled to console-output volume and baud rate. Because `echo()`/`printf` writes to the host UART (UART0) with **no TX ring buffer**, large telemetry lines block the loop while they clock out over the wire. At 115200 baud a ~250 B `core.output` line stretches the loop from 10 ms to ~21 ms, and realistic telemetry pulls the nominal 100 Hz loop down to ~35–50 Hz. Detailed measurement and reasoning in #212.

## Root cause

- UART0 (host console) is installed with **TX buffer size 0** ([main/main.cpp:405](main/main.cpp#L405)), and Lizard never routes stdout through the UART driver (`esp_vfs_dev_uart_use_driver`), so `printf` busy-waits bytes straight into the 128 B hardware FIFO.
- The main loop emits telemetry via `echo()` → `printf` ([main/utils/uart.cpp:31](main/utils/uart.cpp#L31)), then sleeps a fixed `vTaskDelay(10)` ([main/main.cpp:461](main/main.cpp#L461), [main/utils/timing.cpp:6-7](main/utils/timing.cpp#L6)) — the sleep does not subtract elapsed work.
- Net loop period ≈ `10 ms + max(0, message_bytes − 128) × (10 / baud)`. The blocking write time scales with output size and inversely with baud.

## Impact

- Loop rate silently degrades as broadcast/output fields are added — a hidden coupling between logging and control timing.
- Realized loop ~35–50 Hz vs. the assumed 100 Hz under normal telemetry, with jitter tied to message length.
- CPU busy-waits on the FIFO instead of doing useful work.
- Back-pressure only (no data loss, nothing queues unboundedly) — a performance/timing issue, not a correctness one.

## Where this applies

Same root cause (blocking `echo()`/`printf` on a `tx_buffer_size = 0` UART0), so it surfaces on every chip running this firmware:

- **Host console (UART0 → PC):** the primary case, worked through above and in #212.
- **Expander → core (expander UART0 → core's `Serial` RX):** the expander runs the same firmware and reports back over its own UART0 via the identical blocking path, so its main loop is throttled the same way — and it's hit _by design_, since the expander broadcasts its module properties every loop. Two wrinkles make the cheap fix harder here: the link baud is effectively hardcoded at 115200 ([main/main.cpp:394](main/main.cpp#L394)) and must stay in sync on both sides, so "raise the baud" is a coordinated firmware change, not config. (Not yet measured on hardware — same mechanism, expected to apply; worth confirming via `core.millis` deltas on the expander.)
- **Core → expander (commands via the `Serial` module):** **not** affected — these writes ride `Serial::write_checked_line` → `uart_write_bytes` on the UART that _does_ have a 2048 B TX ring buffer ([main/modules/serial.cpp:55](main/modules/serial.cpp#L55)), so they return immediately. Noted only to preempt the "isn't this the 2048 B buffer?" question: that buffer is real on this path, just not on the console path.

## Possible mitigations (cheapest first)

Throttle-free output budget per 10 ms loop (8N1 → 10 bits/byte). Two regimes — a line sent **every loop** is bounded by raw wire throughput (`baud × 0.001`); a single **occasional** line additionally gets the 128 B FIFO as one-shot head-room (`128 + baud × 0.001`):

| baud         | sustained (line every loop) | single occasional line |
| ------------ | --------------------------- | ---------------------- |
| 115200 (now) | ~115 B                      | ~243 B                 |
| 921600       | ~922 B                      | ~1050 B                |

A typical core line is **~224 B every loop** — already ~2× over the ~115 B sustained budget at 115200, which is the throttle. A buffer or loop change cannot raise the sustained budget (it smooths bursts, not average throughput); **only a faster wire moves it.**

1. **Raise the console baud** (e.g. 921600) — lifts the sustained budget from ~115 B to ~922 B, so the ~224 B line stops throttling outright; also cuts the worst-case write from ~10.6 ms to ~1.3 ms. The only lever for sustained telemetry, and one line of config. Needs @JensOgorek to verify stability on one of our robots.
2. **Make the loop deadline-based** — sleep to a fixed 10 ms deadline instead of a trailing `vTaskDelay(10)`. Raises the _single-message_ no-throttle size to ~243 B (115200) / ~1050 B (921600), so occasional long lines (errors, verbose replies) don't stretch the loop. But it does **not** raise the sustained budget — a ~224 B line every loop still throttles at 115200 — so it's complementary to (1), not a substitute.

See #212 for the full analysis (experiment, FIFO/baud model, and per-baud message-length budget).

evnchn:
*Authored with Claude Code, on @evnchn's behalf.*

**TL;DR:** The main loop ends each iteration with an unconditional `vTaskDelay(10)`, so per-loop work is stacked on top of a full 10 ms sleep — loop period becomes `work + 10 ms`. This sleeps to a fixed 10 ms deadline instead, so the period is `max(10 ms, work)`: work up to the budget is absorbed for free, and beyond it the redundant sleep is dropped. Uses `xTaskDelayUntil` (absolute phase reference) so the cadence is drift-free. Baud-agnostic, no protocol change. Implements mitigation #2 from #213. One file, +12/−1.

```cpp
TickType_t last_wake = xTaskGetTickCount();          // before the loop
while (true) {
    // ... loop work ...
    // sleep to the next 10 ms boundary; on overrun xTaskDelayUntil returns pdFALSE
    // without blocking, so yield 1 tick to let the idle task feed the watchdog
    if (xTaskDelayUntil(&last_wake, pdMS_TO_TICKS(10)) == pdFALSE) {
        last_wake = xTaskGetTickCount();
        delay(1);
    }
}
```

> **Update:** switched from a `millis()`-subtraction tail to `xTaskDelayUntil` after @JensOgorek's review — the relative variant drifted ~1 ms fast per loop (tick quantization), the absolute-phase version holds a true 10 ms. Numbers below re-measured on hardware accordingly.

<details>
<summary><b>Measured on hardware (ESP32-classic, ESP-IDF v5.3.1)</b></summary>

Injected per-loop CPU work, light output (so the wire isn't the bottleneck), 115200:

| per-loop work | 0 | 4 | 8 | 12 | 18 ms |
|---|---|---|---|---|---|
| `delay(10)` (before) | 100 | 71 | 56 | 45 | 36 Hz |
| `millis()` deadline (prev revision) | 111 | 111 | 111 | 77 | 53 Hz |
| `xTaskDelayUntil` deadline (this revision) | 100 | 100 | **100** | 77 | 53 Hz |

The loop holds a clean **10.0 ms** period through 8 ms of control work (idle period measured dead-on 10.00 ms), versus 56 Hz before. Beyond the budget the period tracks the work (12 ms → 13 ms, 18 ms → 19 ms), shedding the redundant trailing sleep. The previous `millis()` revision read 111 Hz ≈ 9 ms at low work — running ~1 ms/loop *fast* from per-loop tick quantization; `xTaskDelayUntil` removes that drift and hits the true 10 ms target (correct for a control loop). This recovers control-loop rate without touching baud; it does nothing for *output* throttling (that's wire-bound, per #213) — its job is to stop per-loop work stacking on the trailing sleep.
</details>

<details>
<summary><b>Why the 1-tick floor (not 0)</b></summary>

When the work overruns the 10 ms budget, `xTaskDelayUntil` returns `pdFALSE` without blocking. Rather than spin straight into the next iteration, the overrun branch re-anchors `last_wake` and yields 1 tick. The reason for 1 and not 0: `vTaskDelay(0)` only yields to equal-or-higher-priority ready tasks, so it never lets the priority-0 idle task run — which on ESP-IDF feeds the task watchdog and frees deleted-task memory. A real 1-tick block does. Verified on hardware: with the floor at 0 and 15 ms/loop of work, the TWDT fires on `IDLE0 (CPU 0)` in ~5 s; at 1 it runs stably (15 ms sustained work for 8 s → 0 resets, no TWDT).
</details>

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -426,6 +426,7 @@ void app_main() {
     printf("\nReady.\n");
 
     while (true) {
+        const unsigned long loop_start = millis();
         try {
             process_uart();
         } catch (const std::runtime_error &e) {
@@ -458,6 +459,9 @@ void app_main() {
             }
         }
 
-        delay(10);
+        // sleep the remainder of the 10 ms budget, not a full vTaskDelay(10) after work (#213).
+        // 1-tick floor: vTaskDelay(0) only yields to equal prio, so idle (WDT) needs a real block.
+        const unsigned long elapsed = millis_since(loop_start);
+        delay(elapsed < 10 ? 10 - elapsed : 1);
     }
 }
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -425,8 +425,11 @@ void app_main() {
 
     printf("\nReady.\n");
 
+    // Anchor for the deadline loop below: xTaskDelayUntil advances this by one period
+    // each iteration, giving a drift-free 10 ms cadence (no millis() re-read per loop).
+    TickType_t last_wake = xTaskGetTickCount();
+
     while (true) {
-        const unsigned long loop_start = millis();
         try {
             process_uart();
         } catch (const std::runtime_error &e) {
@@ -459,9 +462,13 @@ void app_main() {
             }
         }
 
-        // sleep the remainder of the 10 ms budget, not a full vTaskDelay(10) after work (#213).
-        // 1-tick floor: vTaskDelay(0) only yields to equal prio, so idle (WDT) needs a real block.
-        const unsigned long elapsed = millis_since(loop_start);
-        delay(elapsed < 10 ? 10 - elapsed : 1);
+        // Sleep until the next 10 ms period boundary instead of a full vTaskDelay(10) after
+        // work, so the period is max(10 ms, work) and drift-free (#213). On overrun
+        // xTaskDelayUntil returns pdFALSE without blocking; floor at 1 tick so the idle task
+        // still runs to feed the watchdog (vTaskDelay(0) only yields to equal-prio tasks).
+        if (xTaskDelayUntil(&last_wake, pdMS_TO_TICKS(10)) == pdFALSE) {
+            last_wake = xTaskGetTickCount();
+            delay(1);
+        }
     }
 }


JensOgorek:
I independently arrived at the same deadline-loop fix on my uart_optimization branch while working through #213 — same core idea (sleep the remainder of the budget, not a trailing full delay(10)), and I landed on the same 1-tick floor for the same reason (idle task / TWDT). So I'm happy to drop my loop change and keep yours; no point maintaining two.

One refinement from my measurements: I used xTaskDelayUntil instead of millis() subtraction. The relative-period variant measured ~9 ms in my setup (1 ms quantization per loop from measuring + sleeping each iteration), while xTaskDelayUntil holds a clean, drift-free 10 ms because it advances an absolute phase reference. You already noted vTaskDelayUntil would be the cleaner idiom — given the drift, I'd suggest actually switching to it. The 1-tick floor maps directly onto its pdFALSE (overrun) return, so it keeps your watchdog fix:


```
TickType_t last_wake = xTaskGetTickCount();
while (true) {
    // ... loop work ...

    // Keep a fixed loop period without drift: sleep until the next period
    // boundary. If the work already overran the period, xTaskDelayUntil does
    // not block and returns pdFALSE; then yield at least 1 tick so the idle
    // task can run and feed the task watchdog.
    if (xTaskDelayUntil(&last_wake, pdMS_TO_TICKS(LOOP_PERIOD_MS)) == pdFALSE) {
        last_wake = xTaskGetTickCount();
        delay(1);
    }
}
```
Might be a nice little improvement to fold into your PR - worth a thought.

My PR will then focus on the parts #216 doesn't cover: raising the console baud (the only lever for the sustained output budget per #213) and moving echo() to async TX via uart_write_bytes.

evnchn:
*Posted with Claude Code, on @evnchn's behalf.*

Thanks @JensOgorek — you were right about the drift, and I've now confirmed it on hardware (ESP32-classic, ESP-IDF v5.3.1, 115200). Switched #216 to `xTaskDelayUntil`.

Measured, injected per-loop work, light output:

| per-loop work | 0 | 4 | 8 | 12 | 18 ms |
|---|---|---|---|---|---|
| `millis()` deadline (prev revision) | 111 | 111 | 111 | 77 | 53 Hz |
| `xTaskDelayUntil` (this revision) | 100 | 100 | 100 | 77 | 53 Hz |

The `millis()` variant was reading 111 Hz ≈ 9 ms — running ~1 ms/loop fast from tick quantization, exactly as you measured. `xTaskDelayUntil` holds a clean 10.0 ms (idle period dead-on 10.00). Beyond budget the two match (period = work).

The 1-tick watchdog floor survives on the `pdFALSE` overrun branch — verified stable under 15 ms sustained work for 8 s: zero resets, no TWDT. PR body table updated to match.

With this in, the loop hunk in #218 can drop as you suggested.

falkoschindler (approved):
Nice change — the reasoning checks out end to end.

Verified the load-bearing details: with `CONFIG_FREERTOS_HZ=1000` the 1-tick floor is a real `vTaskDelay(1)` (no degenerate `vTaskDelay(0)`), so the idle task genuinely runs to feed the watchdog, and re-anchoring `last_wake` on overrun correctly avoids a catch-up burst. `max(10 ms, work)` with drift-free phase is the right shape for a control loop, and the hardware numbers back it up.

LGTM 👍

PR #217, opened 2026-06-17
Optional 4 MB ESP32 support (opt-in --flash-4mb overlay)

evnchn:
*Opened by Claude Code, on @evnchn's behalf.*

**TL;DR:**

**Adds opt-in 4 MB ESP32 support** via `python build.py <target> --flash-4mb`. The default 8 MB build is left completely untouched.

**Merging is not required.** @evnchn's agents cherry-pick this branch to run Lizard on their 4 MB modules regardless — it's opened upstream only for visibility, in case it's useful, and to attract a better approach than the rough overlay here (拋磚引玉).

Why it exists: the default partition table targets 8 MB (`coredump` at `0x400000`), so a stock build `SW_RESET`-loops on a 4 MB chip — `partition 5 invalid … exceeds flash chip size`. 4 MB modules are far more common/accessible, so first-class support is handy.

<details>
<summary><b>The change (3 files, +16)</b></summary>

- `partitions_4mb.csv` — a 4 MB-fitting table: dual OTA `0x1c0000` each + 128 K coredump, all under `0x400000`.
- `sdkconfig.defaults.4mb` — overlay: `FLASHSIZE → 4MB` + `PARTITION_TABLE_CUSTOM_FILENAME=partitions_4mb.csv`.
- `build.py --flash-4mb` — appends the overlay to `SDKCONFIG_DEFAULTS` *after* the base + secret defaults (later wins), so it's strictly additive — absent the flag, nothing changes.

Partition walk (no overlaps, ends `0x3c0000` ≤ `0x400000`; app ~1.29 MB fits each 1.84 MB slot):

| entry | start | size | end |
|---|---|---|---|
| ota_0 | 0x20000 | 0x1c0000 | 0x1e0000 |
| ota_1 | 0x1e0000 | 0x1c0000 | 0x3a0000 |
| coredump | 0x3a0000 | 128K | 0x3c0000 |
</details>

<details>
<summary><b>Verified + reviewed</b></summary>

- **Confirmed on hardware:** `python build.py esp32 --flash-4mb` builds; flashed to a real 4 MB ESP32 → boots to `Ready.`, 1 reset, no partition error, `SPI Flash Size : 4MB`.
- A second-opinion (Codex) review was applied: coredump kept at 128 K (not halved), overlay ordered after the secret defaults, explicit argparse `dest`.
</details>

<details>
<summary><b>Known-rough — choice points (this is a sketch, not a finished design)</b></summary>

- Dual-OTA vs. a single `factory` app on 4 MB.
- A `--flash-4mb` flag vs. a real build target (`esp32_4mb`) vs. auto-detecting flash size in the flasher.
- esp32s3 4 MB variants: the flag is generic but untested there.
- No CI app-size guard for the 4 MB slots.
</details>

**Ask:** none required — no obligation to merge. If a polished version (target vs. flag, CI guard) would be welcome upstream, happy to iterate; otherwise this stands as a reference branch to cherry-pick.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,11 @@
 build/
+build-4mb/
 managed_components/
 owl/
 site/
 __pycache__/
 sdkconfig
+sdkconfig.4mb
 sdkconfig.defaults.secret
 *.swp
 *.bin
--- a/build.py
+++ b/build.py
@@ -12,6 +12,8 @@ def main() -> int:
                         help='Target chip type (default: esp32)')
     parser.add_argument('--clean', action='store_true',
                         help='Clean build directory before building')
+    parser.add_argument('--flash-4mb', dest='flash_4mb', action='store_true',
+                        help='BRICK: build for 4 MB flash modules (default partition table assumes 8 MB)')
     args = parser.parse_args()
 
     base_defaults = Path(f'sdkconfig.defaults.{args.target}')
@@ -23,6 +25,8 @@ def main() -> int:
     all_defaults = [str(base_defaults.resolve())]
     if secret_defaults.exists():
         all_defaults.append(str(secret_defaults.resolve()))
+    if args.flash_4mb:
+        all_defaults.append(str(Path('sdkconfig.defaults.4mb').resolve()))  # 4 MB overlay last, so it overrides
 
     os.environ['IDF_TARGET'] = args.target
     os.environ['SDKCONFIG_DEFAULTS'] = ';'.join(all_defaults)
--- a/build.py
+++ b/build.py
@@ -12,9 +12,10 @@ def main() -> int:
                         help='Target chip type (default: esp32)')
     parser.add_argument('--clean', action='store_true',
                         help='Clean build directory before building')
-    parser.add_argument('--flash-4mb', dest='flash_4mb', action='store_true',
-                        help='build for 4 MB flash modules; the default partition table assumes 8 MB, '
-                             'and on a size mismatch the device boot-loops until reflashed over serial')
+    parser.add_argument('--flash-size', dest='flash_size', choices=['4mb', '8mb'], default='8mb',
+                        help='flash size variant (default: 8mb). 8mb uses the default partition table; '
+                             '4mb builds into an isolated build dir with a 4 MB-fitting table. '
+                             'On a size mismatch the device boot-loops until reflashed over serial.')
     args = parser.parse_args()
 
     base_defaults = Path(f'sdkconfig.defaults.{args.target}')
@@ -26,33 +27,32 @@ def main() -> int:
     all_defaults = [str(base_defaults.resolve())]
     if secret_defaults.exists():
         all_defaults.append(str(secret_defaults.resolve()))
-    if args.flash_4mb:
+
+    # Each flash-size variant gets its OWN generated sdkconfig and build dir, so the two never
+    # share mutable state: switching between them can't leak a stale partition table in either
+    # direction, and the 8 MB / 4 MB artifacts coexist instead of clobbering each other.
+    idf_args = ['idf.py']
+    if args.flash_size == '4mb':
         all_defaults.append(str(Path('sdkconfig.defaults.4mb').resolve()))  # 4 MB overlay last, so it overrides
+        idf_args += ['-B', 'build-4mb', '-D', 'SDKCONFIG=sdkconfig.4mb']
 
     os.environ['IDF_TARGET'] = args.target
     os.environ['SDKCONFIG_DEFAULTS'] = ';'.join(all_defaults)
     print(f'Using target: {args.target}')
+    print(f'Using flash size: {args.flash_size}')
     print(f'Using defaults: {all_defaults}')
 
     if args.clean:
         print('Running full clean...')
-        subprocess.run(['idf.py', 'fullclean'], check=False)
-        for config in ['sdkconfig', 'sdkconfig.old']:
+        subprocess.run([*idf_args, 'fullclean'], check=False)
+        generated = 'sdkconfig.4mb' if args.flash_size == '4mb' else 'sdkconfig'
+        for config in [generated, f'{generated}.old']:
             if Path(config).exists():
                 print(f'Removing {config}')
                 Path(config).unlink()
-    elif args.flash_4mb:
-        # ESP-IDF reads SDKCONFIG_DEFAULTS only when generating sdkconfig from scratch.
-        # An existing sdkconfig (e.g. from a prior 8 MB build) is reused verbatim, so the
-        # 4 MB overlay would be silently ignored and the build would still SW_RESET-loop on
-        # a 4 MB chip. Drop the stale sdkconfig so the overlay is guaranteed to take effect.
-        for config in ['sdkconfig', 'sdkconfig.old']:
-            if Path(config).exists():
-                print(f'--flash-4mb: removing stale {config} so the 4 MB overlay is applied')
-                Path(config).unlink()
 
     print('Building...')
-    result = subprocess.run(['idf.py', 'build'], check=False)
+    result = subprocess.run([*idf_args, 'build'], check=False)
     if result.returncode != 0:
         print('Build failed!')
         return result.returncode
--- a/build.py
+++ b/build.py
@@ -13,7 +13,8 @@ def main() -> int:
     parser.add_argument('--clean', action='store_true',
                         help='Clean build directory before building')
     parser.add_argument('--flash-4mb', dest='flash_4mb', action='store_true',
-                        help='BRICK: build for 4 MB flash modules (default partition table assumes 8 MB)')
+                        help='build for 4 MB flash modules; the default partition table assumes 8 MB, '
+                             'and on a size mismatch the device boot-loops until reflashed over serial')
     args = parser.parse_args()
 
     base_defaults = Path(f'sdkconfig.defaults.{args.target}')
@@ -39,6 +40,15 @@ def main() -> int:
         for config in ['sdkconfig', 'sdkconfig.old']:
             if Path(config).exists():
                 print(f'Removing {config}')
+                Path(config).unlink()
+    elif args.flash_4mb:
+        # ESP-IDF reads SDKCONFIG_DEFAULTS only when generating sdkconfig from scratch.
+        # An existing sdkconfig (e.g. from a prior 8 MB build) is reused verbatim, so the
+        # 4 MB overlay would be silently ignored and the build would still SW_RESET-loop on
+        # a 4 MB chip. Drop the stale sdkconfig so the overlay is guaranteed to take effect.
+        for config in ['sdkconfig', 'sdkconfig.old']:
+            if Path(config).exists():
+                print(f'--flash-4mb: removing stale {config} so the 4 MB overlay is applied')
                 Path(config).unlink()
 
     print('Building...')
--- a/partitions_4mb.csv
+++ b/partitions_4mb.csv
@@ -1,0 +1,7 @@
+# Name,   Type, SubType,  Offset,    Size,      Flags
+nvs,      data, nvs,      0x9000,    24K,
+otadata,  data, ota,      0xf000,    8K,
+phy_init, data, phy,      0x11000,   4K,
+ota_0,    app,  ota_0,    0x20000,   0x1c0000,
+ota_1,    app,  ota_1,    0x1e0000,  0x1c0000,
+coredump, data, coredump, 0x3a0000,  128K,
--- a/sdkconfig.defaults.4mb
+++ b/sdkconfig.defaults.4mb
@@ -1,0 +1,5 @@
+# 4 MB flash overlay. Append AFTER sdkconfig.defaults.<target> (and the secret) so it wins,
+# flipping the default 8 MB layout to a 4 MB-fitting one. Used via: python build.py <target> --flash-4mb
+CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
+CONFIG_ESPTOOLPY_FLASHSIZE="4MB"
+CONFIG_PARTITION_TABLE_CUSTOM_FILENAME="partitions_4mb.csv"
--- a/sdkconfig.defaults.4mb
+++ b/sdkconfig.defaults.4mb
@@ -1,5 +1,5 @@
 # 4 MB flash overlay. Append AFTER sdkconfig.defaults.<target> (and the secret) so it wins,
-# flipping the default 8 MB layout to a 4 MB-fitting one. Used via: python build.py <target> --flash-4mb
+# flipping the default 8 MB layout to a 4 MB-fitting one. Used via: python build.py <target> --flash-size 4mb
 CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
 CONFIG_ESPTOOLPY_FLASHSIZE="4MB"
 CONFIG_PARTITION_TABLE_CUSTOM_FILENAME="partitions_4mb.csv"


falkoschindler (commented):
Nice, clean opt-in change — the default 8 MB build is genuinely untouched, the partition arithmetic is sound (no overlaps, app partitions 64 K-aligned, ends at `0x3c0000` ≤ `0x400000`), and the flashsize-choice override is hardware-confirmed. Two things worth a look before this becomes more than a reference branch:

**1. `--flash-4mb` is a silent no-op on a non-clean tree.** `SDKCONFIG_DEFAULTS` is only read by ESP-IDF when `sdkconfig` is generated from scratch. If a `sdkconfig` already exists (e.g. from a prior default 8 MB build), `idf.py build` reuses it and ignores the appended overlay entirely — so `python build.py esp32 --flash-4mb` _without_ `--clean` produces a binary that still carries the 8 MB partition table, and the 4 MB chip `SW_RESET`-loops on `partition … exceeds flash chip size` exactly as before. The reverse holds too: after a 4 MB build, a plain `build.py esp32` keeps the 4 MB layout until cleaned. Suggest having `--flash-4mb` imply removing `sdkconfig`/`sdkconfig.old` (or at least warn when one exists) so the flag can't quietly do nothing.

**2. The `--help` wording overstates the risk.** `BRICK: build for 4 MB flash modules` reads as if the flag bricks the device, when it's actually the fix for 4 MB modules — and a partition/flash-size mismatch doesn't truly brick an ESP32 anyway: it boot-loops and is fully recoverable by reflashing over serial. Consider something like "use only on 4 MB modules; on a mismatch the device boot-loops until reflashed." Cosmetic.

The open design questions you already listed (flag vs. a real `esp32_4mb` target vs. auto-detect, CI app-size guard, esp32s3 4 MB) aren't re-raised here.

evnchn:

**TL;DR:** Both points fixed in `52f3793` — thanks, both were good catches.

1. **Silent no-op** → `--flash-4mb` now removes a stale `sdkconfig`/`sdkconfig.old` before building (unless `--clean` already does), so the overlay can't be quietly ignored on a non-clean tree.
2. **`BRICK:` wording** → reworded to plain terms. Your read ("bricks the device") is exactly right — and is itself the evidence it was a mistake. There's a slightly embarrassing origin story below, plus an upstream fix so it can't recur.

<details>
<summary><b>1. The no-op fix — what changed & what's verified</b></summary>

```python
elif args.flash_4mb:
    # ESP-IDF reads SDKCONFIG_DEFAULTS only when generating sdkconfig from scratch.
    # An existing sdkconfig (e.g. from a prior 8 MB build) is reused verbatim, so the
    # 4 MB overlay would be silently ignored and the build would still SW_RESET-loop on
    # a 4 MB chip. Drop the stale sdkconfig so the overlay is guaranteed to take effect.
    for config in ['sdkconfig', 'sdkconfig.old']:
        if Path(config).exists():
            print(f'--flash-4mb: removing stale {config} so the 4 MB overlay is applied')
            Path(config).unlink()
```

I went with "imply the removal" over "warn", since a warning still leaves the foot-gun armed. It's gated behind `elif args.clean` so it doesn't double-remove. The reverse you noted (a plain `build.py esp32` after a 4 MB build keeps the 4 MB layout until cleaned) is still true and not addressed here — happy to extend it to regenerate whenever the partition-affecting default set changes, if you'd prefer the flag-vs-target question settled first.

**Verified:** `python3 -m py_compile build.py` passes; `--help` output confirmed (below).
**Not re-verified this round:** this is a logic-only change to the config-selection path — I did **not** re-flash hardware for it. The original "boots to `Ready.` on a real 4 MB ESP32" confirmation stands from the prior commit; the sdkconfig-removal path itself is reasoned, not re-flashed.
</details>

<details>
<summary><b>2. Where "BRICK" came from — 拋磚引玉 (and the upstream fix)</b></summary>

@evnchn's agents run off a catalog of Chinese-idiom "skills" — each chengyu is a compressed handle for a working pattern. This PR was opened under **拋磚引玉** (pāo zhuān yǐn yù), *"throw a brick to attract jade"*: ship a deliberately rough draft so a better version gets drawn out — which is precisely the "no obligation to merge, opened to attract a better approach" spirit of the original post.

What went wrong: the authoring agent internalised *"this whole thing is a brick"* and let the word **leak into the actual `--help` string**. Your reading it as "this flag bricks the device" is the proof it leaked — the metaphor is a private framing between operator and agent, and it has no business in a string an end user runs.

Fixed at the source so it can't happen again: the skill now carries a *"keep the metaphor off the artifact"* boundary + a contamination test, with this exact `BRICK:` case as the worked example → **evnchn-agentic/chengyu-skills#13**. The help string now reads:

```
--flash-4mb   build for 4 MB flash modules; the default partition table
              assumes 8 MB, and on a size mismatch the device boot-loops
              until reflashed over serial
```
</details>

No action needed beyond a re-look when you have a moment. The larger design questions you parked (flag vs. a real `esp32_4mb` target vs. auto-detect, CI app-size guard, esp32s3 4 MB) are still open — glad to take whichever direction you'd want for `0.12.0` now that it's on the milestone.

falkoschindler (changes requested):
Posted by mistake...

falkoschindler (commented):
Thanks for `52f3793` — both my earlier points are resolved: the forward no-op is gone (a stale `sdkconfig` is now dropped before a `--flash-4mb` build), and the reworded help reads correctly. 拋磚引玉 landed; the `BRICK:`-off-the-artifact fix upstream is a nice touch.

You parked the reverse leak behind the flag-vs-target question — "happy to extend it… if you'd prefer the flag-vs-target question settled first." Let's settle it, because the answer removes the leak instead of patching it.

## Direction: isolate each variant's config, don't mutate a shared one

The reverse leak (a plain `build.py esp32` after a 4 MB build keeps the 4 MB layout until cleaned) and the `sdkconfig`-deletion workaround are the same root cause: the flag mutates the single shared, generated `sdkconfig`, and ESP-IDF only regenerates it from `SDKCONFIG_DEFAULTS` when that file is absent. Extending the deletion to the reverse direction would be one more special case on shared state.

Instead, give each variant its own generated config and build dir, so there's nothing shared to leak:

```
default →  idf.py                                      build   # sdkconfig,     build/
4 MB    →  idf.py -B build-4mb -D SDKCONFIG=sdkconfig.4mb build
```

That lets us drop the deletion block entirely: no `--clean` needed to switch, both leak directions gone by construction, and the 8 MB / 4 MB artifacts coexist instead of clobbering each other. The new generated files (`sdkconfig.4mb`, `build-4mb/`) would need `.gitignore` entries — the current patterns match `sdkconfig` and `build/` exactly.

On the surface itself: I'd lean `--flash-size {4mb,8mb}` (default `8mb`) over the boolean `--flash-4mb`, so it reads explicitly and leaves room for other sizes. This is squarely the "flag vs. real `esp32_4mb` target" question you raised — my take is that a full IDF target is overkill (4 MB is a flash-size variant of the same chip, not a different chip), and per-variant config isolation gets the robustness of a target without the ceremony.

## Scope question — your call

Two follow-ons, both optional for this PR:

1. **CI app-size guard** — fail loudly if the app overflows the 1.75 MB (`0x1c0000`) 4 MB OTA slot, instead of only finding out at flash time.
2. **Release artifacts** — add `esp32`/`esp32s3` × 4 MB to the release matrix and upload distinctly-named `..._4mb.bin` files, so 4 MB users get a prebuilt binary.

Happy to help with either, or to land the config-isolation fix first and follow up. Any preference for `0.12.0`?

evnchn:
Thanks — settling it was the right call, and the config-isolation direction is cleaner than patching the reverse leak. Implemented exactly as you sketched, in `b35fd0d`.

**What changed**

- Each variant now owns its generated config + build dir: default → `idf.py build` (unchanged), 4 MB → `idf.py -B build-4mb -D SDKCONFIG=sdkconfig.4mb build`. Both leak directions are gone **by construction** — the `sdkconfig`-deletion block is deleted, no `--clean` needed to switch, and the 8 MB / 4 MB artifacts coexist instead of clobbering each other.
- Surface is now `--flash-size {4mb,8mb}` (default `8mb`), per your lean.
- `.gitignore` gains `build-4mb/` + `sdkconfig.4mb`.

<details>
<summary><b>Verification</b></summary>

Command construction confirmed per-variant with a stub `idf.py` capturing argv + env:

| invocation | `idf.py` args | `SDKCONFIG_DEFAULTS` |
|---|---|---|
| `build.py esp32` (default) | `['build']` | `sdkconfig.defaults.esp32` |
| `build.py esp32 --flash-size 4mb` | `['-B','build-4mb','-D','SDKCONFIG=sdkconfig.4mb','build']` | `…esp32;…4mb` |
| `build.py esp32s3 --flash-size 4mb` | same `-B/-D` | `…esp32s3;…4mb` |
| `build.py esp32 --flash-size 4mb --clean` | `…fullclean` then `…build`, deletes `sdkconfig.4mb[.old]` only | `…esp32;…4mb` |

- The default (8 MB) path is byte-identical to before — no `-B`/`-D`, default `sdkconfig`/`build/`.
- `git check-ignore` confirms `build-4mb/` and `sdkconfig.4mb` are ignored; pre-commit (whitespace, double-quote-fixer, mdformat) passes.
- **Hardware:** verified at the Python/command-construction level; a re-flash of the reworked invocation on a real 4 MB chip is still to follow (my earlier hardware-confirm was on the old `--flash-4mb` code) — I'll report back once it's boot-confirmed.
</details>

The two follow-ons — CI app-size guard and 4 MB release artifacts — are worth doing; I'll track them as a separate follow-up rather than expand this PR. No strong preference on `0.12.0`, happy either way — your call on the milestone.

falkoschindler (changes requested):
Thanks a lot for the well-documented PR and for framing it so clearly as a reference sketch — the write-up, the hardware verification, and the honest list of choice points are much appreciated. I took a close look, and while the core mechanism is sound (overlay ordering is correct, the partition arithmetic checks out, and the isolated build dir cleanly protects the default 8 MB path), I don't think we can merge it in its current form. The main reason: the feature currently ends at the build step, and everything downstream would set a trap for exactly the users it targets.

**Why I'm requesting changes / not merging as-is:**

1. **All flash/debug tooling still consumes `build/`, so the 4 MB build is never the one flashed.** `flash.py` hardcodes `build/...` with no override, `espresso.py` defaults to `build/...`, `upload_ssh.py` rsyncs only `build`, and `core_dumper.py`/`backtrace.sh` symbolize against `build/lizard.elf`. So after `python build.py esp32 --flash-size 4mb`, the documented `sudo ./espresso.py flash <device>` silently flashes the stale 8 MB artifacts — producing exactly the boot loop this PR sets out to fix. Merged as-is, this would hurt more people than it helps.

2. **`--clean` is silently narrowed to the selected variant.** `python build.py --clean` leaves `build-4mb/` and `sdkconfig.4mb` untouched (and no flag cleans both). Since ESP-IDF ignores `SDKCONFIG_DEFAULTS` changes when a generated sdkconfig already exists, editing `sdkconfig.defaults.*` and cleaning would still produce a 4 MB firmware built from stale settings, with no warning.

3. **No CI or release integration.** The release workflow builds only the default variant, so no 4 MB image is ever published (4 MB devices could never be updated from releases) and nothing guards the app size against the smaller 0x1c0000 slots — the first firmware that outgrows them would be discovered in the field as a boot loop. You listed this yourself as known-rough; for an upstream merge it's the main gap.

**Smaller points, in case you iterate:**

4. The partition table strands 256 K: ota_0 at 0x20000 and ota_1 at 0x200000 with 0x1e0000 each, coredump at 0x3e0000, ends exactly at 0x400000 — 128 K more headroom per slot for free.
5. Placing the generated config inside the build dir (`-D SDKCONFIG=build-4mb/sdkconfig`, ESP-IDF's multi_config idiom) would let `idf.py fullclean` handle cleanup and drop the extra unlink branch and `.gitignore` entries.
6. The new flag isn't documented (AGENTS.md "Development Commands" would be the place).

No pressure given your own framing — the branch works fine as a cherry-pick reference as it stands. But if you'd like to pursue first-class upstream support, points 1–3 plus the flag-vs-target question are what I'd want covered before merging (happy to discuss the approach here first). Thanks again for the careful groundwork and for sharing it upstream! 拋磚引玉 indeed.

evnchn:
This one I think will be mature naturally once I have a mix of ESP32s with both 4MB and 8MB flash. It's difficult to cater for a build target I do not have (same goes for you most probably)

Specs for Zauberzeug-used ESP32s?

evnchn:
Thanks Falko — this is exactly the review I hoped 拋磚引玉 would pull, and you're right that ending at the build step sets a trap. Let me commit to a concrete plan for 1–3 + flag-vs-target *before* touching code, so we're aligned on the shape.

**flag-vs-target:** keeping **`--flash-size {4mb,8mb}`** (default `8mb`). Agreed a full `esp32_4mb` IDF target is overkill for what's a flash-size variant of the same chip.

**1 — downstream tooling flashes `build/`, not the 4 MB build.** Add `--build-dir` (default `build`) to the flash/debug tools so the `build-4mb/` artifacts `build.py --flash-size 4mb` already emits are the ones consumed: `espresso.py` derives its bootloader/partition/firmware defaults from it (explicit overrides still win); `flash.py`, `upload_ssh.py` (rsync source + forwarded arg), `core_dumper.py` (elf), and `backtrace.sh` all honor it. One knob, 8 MB path byte-identical when it's unset.

**2 — `--clean`.** With the generated config living in the build dir (point 5), `build.py --clean --flash-size 4mb` → `idf.py -B build-4mb fullclean`, so editing `sdkconfig.defaults.*` then cleaning regenerates from defaults — the stale-settings trap is gone.

**5 — config in build dir.** Adopting your `-D SDKCONFIG=build-4mb/sdkconfig` idiom, which lets `fullclean` own the cleanup and drops the extra unlink branch + the `sdkconfig.4mb` `.gitignore` entry.

**4 — partition headroom.** Taking your exact table (below); reclaims the 256 K the current CSV strands.

**3 — CI/release.** Extend the release matrix to also build `--flash-size 4mb` and publish `lizard_<ver>_<target>_4mb.bin`, plus a step that fails the build if the app outgrows the `0x1e0000` slot — caught in CI, not in the field. One call for you here: **do you want 4 MB images shipped from releases**, or keep them build-from-source and I just add the size guard?

**6 —** documenting `--flash-size` in AGENTS.md's Development Commands.

<details><summary>Partition table (your numbers)</summary>

| entry | start | size | end |
|---|---|---|---|
| ota_0 | 0x20000 | 0x1e0000 | 0x200000 |
| ota_1 | 0x200000 | 0x1e0000 | 0x3e0000 |
| coredump | 0x3e0000 | 128K | 0x400000 |

+128 K/slot vs current, ends exactly at `0x400000`.
</details>

If that matches your merge bar I'll implement it as one clean commit (hardware-verified on a real 4 MB module) and push. Anything you'd shape differently — the tooling surface or the release-artifact question especially — I'll fold in first.

falkoschindler:
@evnchn Thanks for the detailed plan — it maps cleanly onto every point I raised, and doing it as a plan-before-code was the right call. Before you invest in building it out, though, I'd like to step back with you, because I think there are two ways to make this change *smaller* rather than bigger, and both are worth deciding here first. They're independent, but they share one root question: **do we complicate the build/flash chain for a 4 MB variant, or do we keep a single chain and concentrate on `espresso.py`?**

## 1. Do we even need a variant? Just build 4 MB for everyone.

Looking at the actual numbers, the 8 MB layout barely uses the extra flash:

| slot | default `partitions.csv` (8 MB) | your 4 MB table | delta |
|---|---|---|---|
| ota_0 / ota_1 (each) | `0x1f0000` = 1.9375 MB | `0x1e0000` = 1.875 MB | **−64 KB** |
| coredump | 128 K @ `0x400000` | 128 K @ `0x3e0000` | same |

The 8 MB layout's *only* use of flash above the 4 MB line is that 128 K coredump sitting at `0x400000` — which is precisely what boot-loops a 4 MB chip. Everything else already lives below 4 MB, and there's no filesystem or data partition using the upper half at all.

So collapsing to a single 4 MB-fitting partition table (with flash-size `detect`, which `espresso.py` already passes) costs **exactly 64 KB per OTA slot** — the app is ~1.29 MB, so ~585 KB headroom remains — and produces **one image that boots on both 4 MB and 8 MB hardware**. No `--flash-size` flag, no isolated build dir, no `--build-dir` threading, no release-matrix fork. The whole variant machinery disappears.

The only thing this forecloses is *someday* using the upper 4 MB on 8 MB boards (bigger OTA slots, a littlefs data partition, …). Nothing does today; if that need ever arrives, we re-introduce a variant *then*, with a concrete reason instead of speculative machinery.

One thing I'd want to confirm before committing: that a 4 MB-configured image actually boots on the 8 MB boards we ship — the flash-size header is the one thing that behaves differently across chips. Conceptually it's fine (`detect` patches the header at flash time, and a 4 MB table fits within 8 MB), but it deserves one hardware check on a real 8 MB module.

**Open question for us:** does anyone foresee needing that upper 4 MB on 8 MB boards in the foreseeable future? If no, unifying on 4 MB is by far the simplest robust answer, and this PR shrinks to: swap the default table to the 4 MB-fitting one, set the flash size accordingly, and add the CI app-size guard against the `0x1e0000` slot. If yes, we keep the variant and go with your plan.

## 2. The tooling surface — lean on `espresso.py` instead of threading a flag through everything.

Notice *why* point 1 of the plan (`--build-dir` in `espresso.py`, `flash.py`, `upload_ssh.py`, `core_dumper.py`, `backtrace.sh`) exists at all: it's only needed because two build dirs coexist. If we take the unify route above, it evaporates entirely. But even independently, I don't think we should add a new parameter to those four tools — most of them are effectively dead or already superseded, and `espresso.py` is the one we actually use:

- **`flash.py`** — already prints a deprecation banner pointing at `espresso.py` (added in #143, *"Introduce espresso.py to replace flash.py in the future"*). Everything it does — flash / erase / enable / reset — `espresso.py` already covers, and more automatically (it auto-detects the serial device from the L4T version instead of taking `nano`/`xavier`/`orin` by hand). This should just be removed, together with its only helper `esp.py`, not extended.
- **`backtrace.sh`** — hardcodes a toolchain path (`~/esp/esp-tools_4.2/…`, ESP-IDF 4.2-era) that no longer exists on a v5.3 setup; it can't have worked recently. Dead, not worth threading a flag through.
- **`upload_ssh.py`** — genuinely useful (remote rsync + flash over SSH), but it's the sole live caller of `flash.py`, so it just needs repointing to `espresso.py flash`.
- **`core_dumper.py`** — real debug value, no overlap; a natural `espresso coredump` subcommand if we want it.

So rather than making the build chain wider, the direction I'd prefer is: **concentrate on `espresso.py` as the single flash/debug entry point** and let the stale scripts go. That's a cleanup we'd want regardless of the 4 MB decision — but it's also exactly what removes the pressure this PR was putting on the tooling in the first place.

---

I'd rather settle both of these here before you write more code, since together they decide whether the 4 MB support is a variant we maintain forever or just how we build by default. My lean is strongly toward "unify on 4 MB + lean on espresso" unless the upper-4 MB question above has an answer I'm not seeing. What's your read — especially on whether anything needs that top 4 MB?

JensOgorek:
Some Zauberzeug context (I'm the one who dropped 4 MB modules from our fleet ~2 years ago): all our devices are 16 MB, so we have no 4 MB stake here — this is purely about supporting your 4 MB use case upstream. And OTA (the OTB over-the-bus update path) is rare for us: essentially only on the ESP32-S3 Plexus; our classic-ESP32 boards are flashed over serial.

A few thoughts, from "just build it" down to the smallest change:

**1. The first direction that came to my mind — on 4 MB, just drop OTA (a single `factory` app instead of dual-OTA).** This is the choice point you listed but didn't pursue, and I think it's the cleanest:

- Those devices don't need OTA anyway (OTA is an ESP32-S3-Plexus thing for us), so a factory layout matches what a 4 MB board actually does — flash over serial.
- It's essentially free firmware-side: OTB already degrades gracefully without an OTA partition. `otb.cpp` does `esp_ota_get_next_update_partition(nullptr)`, which returns `nullptr` on a factory layout, so `__OTB_BEGIN__` just replies `__OTB_ERROR__:failed to begin update` and returns — no `#ifdef`, no code change. The device simply can't self-update over the bus, which is exactly the intended trade-off. (`serial-replicator` uses `esp_ota_get_running_partition()`, which works on any layout.)
- Bonus: a single factory app gives ~3.7 MB on 4 MB (vs. ~1.87 MB per dual-OTA slot), so there's no partition squeeze at all.

**2. Alternatively, shrinking the OTA slots to fit under 4 MB (your current dual-OTA table) is also fine by me** — as long as the app stays comfortably within the smaller slot with enough headroom (today ~1.29 MB in a 1.875 MB slot → ~585 KB spare). A CI app-size guard against the slot would cover the "app grew past it" risk.

**3. And the most basic question, honestly: is 4 MB worth targeting at all, versus just using an 8 MB+ module?** Retiring 4 MB is exactly why we stopped hitting this class of problem ~2 years ago — 8 MB+ ESP32s are cheap and widely available now. You mentioned 4 MB modules are more common/accessible for you; if there's a real constraint (existing stock, a specific module, cost at volume), that's good to know — but if 8 MB+ is an option on your side, it sidesteps the whole partition question and is by far the least maintenance for everyone.

Since we're all 16 MB, I'd keep the 8 MB default untouched and make any 4 MB layout the opt-in either way. @falkoschindler, curious what you think about factory-vs-dual-OTA now that "no OTA is basically free" is on the table.

falkoschindler:
With @JensOgorek's context now in the thread, I've decided to close this — and I want to be open about the reasoning, because it's also the best recommendation for you.

**Why we won't merge a 4 MB variant:** Our fleet is entirely 16 MB, and @JensOgorek deliberately dropped 4 MB ~2 years ago, so we have zero internal need — and, more importantly, no testable target: we could never verify a merged variant on real 4 MB hardware. The exact point you raised yourself ("difficult to cater for a build target I do not have") applies to us identically. A permanently untested variant (partition table, build variant, CI matrix, release artifacts) would be pure maintenance load for hardware we don't own. That makes the state you called sufficient from the start — the branch stays around, your agents cherry-pick it — the clean outcome here.

**What survives from the thread, and the best starting point for cherry-picking:** @JensOgorek's observation that "no OTA" is essentially free on 4 MB — a single `factory` app instead of dual-OTA. OTB degrades on its own: `esp_ota_get_next_update_partition(nullptr)` returns `nullptr` on a factory layout, `__OTB_BEGIN__` simply replies `__OTB_ERROR__`, no `#ifdef` needed. That gives you ~3.7 MB in one slot instead of ~1.87 MB per dual-OTA slot and makes the partition squeeze disappear entirely. If you keep maintaining your branch, that's a more elegant base than the current dual-OTA table.

**What we're taking away:** Your tooling point (espresso.py as the single flash/debug entry point, the stale scripts gone) is valuable independent of 4 MB — we'll spin that out as its own issue and do it there. Thanks for surfacing it along the way.

Closing this changes nothing on your side: the branch lives in your fork, fully under your control, so you and your agents can keep cherry-picking it independently of the PR. Thanks again for the consistently high-quality discussion.

evnchn:
@falkoschindler i think the condensed read is:

Unable to fight for 4MB support _now_, considering team bandwidth and tooling shenanigans. 

Once those get straightened out, we can re-evaluate if we can do it.

PR #218, opened 2026-06-18
Make the UART0 baud rate configurable

### LINKED ISSUE #213: Main loop is throttled by blocking console writes

JensOgorek:
### Motivation

Raising the UART0 baud rate is the only lever that increases the *sustained* console-output budget (see #213 / #212): a typical core line is ~224 B every loop, already ~2× over the ~115 B sustained budget at 115200. But hardcoding a higher rate is a breaking change — feldfreund / robot_brain and the expander↔core link would all have to switch in lockstep.

This PR therefore makes the baud rate **configurable and persistent** instead of hardcoded, so it can be raised per device (e.g. during robot testing) without breaking anything that has not been updated yet.

The scope is deliberately limited to the baud rate:

- The **deadline-loop** part is now in `main` via #216, so it dropped out when `main` was merged in.
- The **async-TX** experiment is left out for now (log-corruption risk and questionable benefit, as discussed in the review).

### Implementation

**Firmware**

- `core.set_baudrate(baud)` persists the rate to NVS (namespace `uart`, key `baudrate`), validated against the supported rates 115200 / 230400 / 460800 / 921600. It takes effect on the **next restart**, so updating a host is never forced.
- `app_main()` reads the stored rate before configuring UART0; the default is 115200, so out-of-the-box behaviour is unchanged (no breaking change).

**Host tools take an explicit baud rate**

- `monitor.py` and `configure.py` gain a `--baud` flag (default 115200); `otb_update.py` already had one. After raising the rate on the device, reconnect the tools with the matching `--baud`.
- `monitor.py`: fixed a crash where invalid bytes (noise / baud mismatch) raised `UnicodeDecodeError` into a handler that referenced an unassigned variable (`NameError`), killing the reader thread and forcing a monitor restart. It now decodes tolerantly and guards the checksum parse.

Usage:

```
core.set_baudrate(921600)
core.restart()
./monitor.py /dev/ttyUSB0 --baud 921600
```

Note: the ROM bootloader and the early boot log always stay at 115200 regardless of this setting, so pre-application boot output will look garbled at higher rates.

### Progress

- [x] Tested on hardware (ESP32 on USB: `set_baudrate(921600)` → restart → reconnect with `--baud 921600`; `core.info()` works; revert to 115200 works; unsupported rates are rejected).
- [x] Documentation has been updated (`module_reference.md`).


falkoschindler (changes requested):
Thanks for the PR, @JensOgorek! Since Claude is down at the moment, I'll just drop my overall impression of the pull request.

There are basically three parts, each with a different status:

1. **Higher baud rate**
  This is good. But a way to configure it on demand would be really useful and would avoid a breaking change. Otherwise we might run into trouble upgrading our robots and toolchains.

2. **Deadline-based main loop**
  Now that PR #216 is merged, this is redundant. Or put differently: The merge conflict needs to get resolved. Then this change will vanish.

3. **Async TX**
  I'm not sure if this is worth it, given problems like the log line corruption you mentioned. But we can think about it.

We should consider splitting this PR into two so that the two remaining parts don't block each other from getting merged.

falkoschindler:
Claude:

**1. `main/utils/uart.cpp:36` — async `echo()` and synchronous `printf`/`ESP_LOG` share UART0 with no ordering, so a concurrent log corrupts a protocol frame.** (CONFIRMED)
`echo()` now queues into the driver TX ring buffer (drained by the UART ISR), while `printf` and every `ESP_LOGx` go through IDF's default console straight to the UART0 TX FIFO — there is no `esp_log_set_vprintf` / `uart_vfs_dev_use_driver` routing anywhere in the tree. If a module fires `ESP_LOGx` (serial-replicator, can, BNO055, …) while the ISR is mid-frame on an echoed `…@<checksum>\n`, the two byte streams interleave and the host's `check()` XOR fails → the line is silently dropped. This is exactly the "ESP_LOG hardening (open)" item in the PR body; flagging it so it isn't lost — the clean fix (`esp_log_set_vprintf` → `uart_write_bytes`) routes both paths through the same ring.

**2. `monitor.py:95` / `configure.py:40` — host tools fixed at 921600 can't read the ESP32 ROM boot phase.** (PLAUSIBLE, low–moderate)
The first-stage ROM bootloader emits its reset/boot banner at the ROM-fixed 115200 (unaffected by `CONFIG_ESP_CONSOLE_UART_BAUDRATE`), so a host opened at 921600 sees garbage until `app_main` reconfigures UART0. All application output is fine; the cost is diagnosing pre-app brownout / boot-loop cases where the chip never reaches the app. This is the standard trade-off `idf.py monitor` auto-handles but plain pyserial tools don't — worth a note in `docs/tools.md` at minimum.

**3. `main/utils/uart.cpp:18,35` — the new `static char line[1032]` + full re-`snprintf` duplicates work and code that `serial.cpp` already avoids.** (cleanup)
`Serial::write_checked_line` (main/modules/serial.cpp:98-100) frames the identical `@%02x\n` checksum protocol with two writes — `uart_write_bytes(&message[start], len)` then a ~4-byte suffix from a 16-byte `checksum_buffer` — with no payload copy. `echo()` instead reserves a second ~1 KB static buffer and `%s`-copies the whole payload on every line before sending. Reusing the two-write pattern would drop the 1 KB of permanent static RAM and the per-line memcpy on the hot path, and collapse two divergent copies of the wire framing into one.

**4. `main/main.cpp:397` — `921600` hardcoded in the driver re-config while `CONFIG_ESP_CONSOLE_UART_BAUDRATE=921600` already exists in both sdkconfigs.** (cleanup, minor)
The firmware now states the baud three times (the C++ literal + two sdkconfig entries); a future change must keep them in sync or output silently garbles. `main.cpp` could reference `CONFIG_ESP_CONSOLE_UART_BAUDRATE` so the sdkconfig is the single firmware-side source of truth. (The 7+ host-side Python copies are a softer version of the same.)

falkoschindler (changes requested):
Thanks @JensOgorek — the re-scoping resolved all four of my earlier points, and the firmware half is in good shape: making the rate configurable, persistent (NVS), default 115200 and applied-on-next-restart is exactly the non-breaking design we wanted, and the `monitor.py` decode/checksum crash fix is a nice catch. My remaining concerns are all on the host-side auto-detection, plus one question about whether we need it at all right now.

### Problems with the current auto-probe

1. **It probes by closing and reopening the port for every rate** (`serial_utils.py:29`). On boards with the standard auto-reset wiring, each `serial.Serial(...)` open toggles DTR/RTS and reboots the ESP, so a single connect now reboots the chip up to 4 times before the real connection (which reboots it once more). It also means each probe races the device's ~1.5 s boot before `app_main` reconfigures UART0 and starts speaking at the stored rate.

2. **Detection failure is silent** (`serial_utils.py:37`). When no rate validates, the function returns `default=115200`, which is indistinguishable from a genuine 115200 device — `monitor.py` even prints a confident `Connecting ... at 115200 baud`. If the chip is actually at a higher rate but the probe missed it, the user just gets garbled output with no hint that detection failed.

3. **An unresponsive port costs ~36 s** (`serial_utils.py:32`). A running Lizard device answers within the first rate, so the normal case is fast — but an unflashed ESP, a non-Lizard chip, or one stuck bootlooping makes every `read_until` block its full `0.3 s`, i.e. `30 × 0.3 × 4 ≈ 36 s` before falling back. This one is largely inherent: probing _less_ makes detection _less_ reliable (false negatives on slow/booting devices), so it's a genuine trade-off, not something we can just tune away.

### Possible enhancements (if we keep auto-probe)

- **Reuse a single open port** and switch `port.baudrate` between attempts instead of close/reopen — pyserial reconfigures a live port, so this brings us back to a single reset and removes problem 1.
- **Surface failure** — warn or raise when no rate validates, instead of silently returning 115200.
- **Cycle the rates round-robin** — one attempt per rate per round, instead of exhausting all 30 attempts on the first rate before moving to the second. This catches the device as soon as it emits a valid line at whatever rate it's on (robust to the reset/boot race), independent of rate order, without lowering the total probe budget. It doesn't change the dead-port worst case, but makes a responsive device detect faster and more reliably.

### The actual question: do we need auto-probe now?

I'm not convinced auto-detection earns its complexity at this stage. Whoever runs `core.set_baudrate(...)` knows the rate they set, and the fleet tooling scripts these calls, so a plain `--baud` flag (default 115200) is deterministic and has none of the reset / hang / silent-fallback failure modes above. Auto-detect is genuinely nicer for ad-hoc interactive use, but it's strictly more moving parts.

My preference: **ship the firmware half plus a simple `--baud` (default 115200) now**, and revisit auto-detection as a follow-up if interactive use turns out to justify it.

falkoschindler (approved):
The PR is focused cleanly on the baud-rate rework — no deadline-loop or async-TX code in the diff, so the scope matches what we agreed. `core.set_baudrate` validates the rate at the single write path and the default of 115200 keeps out-of-the-box behaviour unchanged. The NVS access was factored into shared `write_u32`/`read_u32` helpers, so `set/get_baudrate` and `set/get_user_pin` are now thin wrappers instead of duplicated `nvs_open/set/commit/close` boilerplate. The `monitor.py` rework correctly fixes the old decode crash (`NameError` on `UnicodeDecodeError`) while staying robust to empty/garbled lines. Builds green for esp32.

No actionable findings. LGTM. 👍

PR #220, opened 2026-06-19
Escape '%' in configure.py payloads sent over the serial bus

JensOgorek:
## Problem

When configuring a node over the serial bus (`configure.py <file> <port> --serial-bus <id>`), every config line is wrapped into a `bus.send(<id>, "<payload>")` call that is executed on the **coordinator**.

The coordinator's `bus.send` performs printf-style formatting (see `main/modules/serial_bus.cpp` → `format_args` in `main/utils/format.cpp`). The payload is used as the **format string**, but `configure.py` passes no additional arguments. As a result, any `%` in a config line is interpreted as a format specifier and aborts that line with:

```
format: '%.2f' has no matching argument
```

Concretely, this silently dropped **every** config line containing a format string, e.g.:

```
bus.send(1, "!!scissor_position=%.2f", lift.position)
bus.send(1, "!!scissor_enabled=%s", lift.enabled)
bus.send(1, "!!scissor_fault=%s", lift.fault)
```

Those lines never reached the target node, so the node booted with an incomplete startup script and stopped emitting the affected messages — while still appearing healthy and reachable, which made the failure hard to spot.

## Fix

When `--serial-bus` is used, escape `%` as `%%` in the payload before wrapping it in the `bus.send(...)` call. The coordinator's printf formatting collapses `%%` back into a literal `%`, so the original line is reconstructed verbatim and forwarded to the target node intact.

Direct (non-bus) configuration is unaffected — the escaping only applies on the `--serial-bus` path.

## Testing

Verified on hardware: configuring an ESP32-S3 scissor-lift node over the bus via the coordinator. Before the fix the `%`-containing `bus.send` lines were dropped and no position messages arrived; after the fix the full startup script is applied and the node resumes sending its messages.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -19,7 +19,11 @@ args = parser.parse_args()
 
 def send(payload: str) -> None:
     """Send a payload string to the ESP32, optionally over a serial bus."""
-    line_ = f'{args.bus_name}.send({args.serial_bus}, {json.dumps(payload)})' if args.serial_bus else payload
+    if args.serial_bus:
+        escaped_ = payload.replace('%', '%%')
+        line_ = f'{args.bus_name}.send({args.serial_bus}, {json.dumps(escaped_)})'
+    else:
+        line_ = payload
     print(f'Sending: {line_}')
     checksum_ = 0
     for c in line_:


evnchn:

**TL;DR — LGTM from my read, but I'm not deep in this codebase, so the final call is yours / @JensOgorek's.**

The diff is tiny (+5/−1) and I verified the mechanism end-to-end empirically (read the firmware + ran both halves), so I'm comfortable it's correct:

- **The fix round-trips exactly.** `%`→`%%` is collapsed back to a literal `%` by the coordinator's `format_args`, and the target node ends up storing the verbatim original line — confirmed against the actual firmware path, not just the PR description.
- **It also repairs the end-to-end checksum**, not only the dropped lines: `core.startup_checksum()` can only match once the `%`-lines actually reach the node.
- **One latent sibling, not blocking:** `otb_update.py:52` feeds payloads into the same coordinator-side `bus.send` without either guard this PR adds — but its current callers only pass base64 + constants, so it isn't triggerable today.

Deferring to you on whether the sibling + the pre-existing non-ASCII note (below) are worth folding in here or just tracking separately.

<details>
<summary><b>Empirical verification (ran both halves)</b></summary>

**Mechanism, traced through the real firmware:**

1. `configure.py` (bus path) sends the coordinator `bus.send(node, "<%%-escaped line>")`.
2. Coordinator evaluates it → `serial_bus.cpp:99-100` runs `format_args`, which collapses `%%`→`%` (`format.cpp:27-29`) and forwards the **collapsed** payload.
3. Node receives it → `serial_bus.cpp:269/276` calls `process_line` → `main.cpp:339` `append_to_startup` stores the **verbatim original** line.
4. `core.startup_checksum()` (`core.cpp:77-82`, `checksum += c`) therefore matches `configure.py`'s `sum(ord(c)) % 0x10000` over the original file.

**Round-trip identity — escape → `format_args` → payload-to-node** (faithful Python port of `format.cpp`, validated to reproduce the exact firmware error string):

| input line | recovers to original? |
|---|---|
| `...scissor_position=%.2f...` | ✅ |
| `...scissor_enabled=%s...` | ✅ |
| already-doubled `%%` + single `%d` | ✅ |
| embedded `"`, `\`, and `100%` | ✅ |
| trailing `%` at EOL | ✅ |

Escaped content provably never reaches the "needs argument" branch.

**Control (pre-fix):** the unescaped `%.2f` line throws on the coordinator with the exact reported error —
```
format: '%.2f' has no matching argument
```
so the line is dropped before reaching the node. No regression on the direct (non-bus) path — it's correctly left unescaped (no `format_args` there).

</details>

<details>
<summary><b>Finding A — <code>otb_update.py:52</code> has the same unescaped <code>bus.send</code> shape (latent, not blocking)</b></summary>

```python
# otb_update.py:52
dev.write(f'{args.bus}.send({args.target},"{msg}")\n'.encode())
```

Same coordinator-side `bus.send` sink, but **neither** safeguard configure.py now has: no `%`→`%%`, and no `json.dumps` (raw `"{msg}"`, so a `"` or `\` in `msg` also breaks the literal).

**Currently safe** — every `msg` passed to `transact()` is a constant (`__OTB_BEGIN__`, `__OTB_COMMIT__`, `__OTB_ABORT__`, `core.restart()`) or a base64 chunk, and standard base64 is `[A-Za-z0-9+/=]` (no `%`/`"`/`\`). So this is latent, not live.

**MRE** (what would happen if a caller ever passed such a `msg`):
```
msg = 'progress 50% done'
  otb_update.py emits : bus.send(1,"progress 50% done")
  configure.py emits  : bus.send(1, "progress 50%% done")
  coordinator: LINE DROPPED -> format: '% d' has no matching argument

msg = 'name="lift"'
  otb_update.py emits : bus.send(1,"name="lift"")   # literal closes early -> parse error
  configure.py emits  : bus.send(1, "name=\"lift\"")
```
Cheap to harden symmetrically (`json.dumps` + `%`-escape), or just a comment noting the constraint.

</details>

<details>
<summary><b>Finding B — pre-existing: <code>json.dumps</code> emits <code>\uXXXX</code>, but Lizard's parser can't decode it (bus path only)</b></summary>

Not introduced by this PR — the bus path already used `json.dumps`. Flagging only because it's the same "escaping vocab mismatch" family.

Lizard's tokenizer (`parser.h:1735` `ESCAPE_CHAR`) understands only `\b \f \n \r \t` (plus `\"`/`\\` passthrough). `json.dumps` (ASCII-default) emits `\uXXXX` for any non-ASCII char, which Lizard decodes as a literal `u` + digits:

```
[CORRUPT] original='café'    json.dumps -> "café"   Lizard parse -> 'cafu00e9'
[CORRUPT] original='arrow→'  json.dumps -> "arrow→" Lizard parse -> 'arrowu2192'
[OK ]     original='tab\there'  -> survives (overlap set)
[OK ]     original='quote"x'    -> survives
[OK ]     original='back\slash' -> survives
```

So a non-ASCII char inside a string literal in a bus-configured line is corrupted (and would then surface loudly as a checksum mismatch). Plain ASCII configs — the norm — are unaffected. Backlog note, not a blocker.

</details>

<details>
<summary><b>Refuted on the way — "the escaped <code>%%</code> ends up in the stored checksum"</b></summary>

One review pass flagged a checksum mismatch on the theory that the device stores the escaped `bus.send(...)` text with `%%`. That's not what happens: the coordinator **evaluates** `bus.send(...)` and forwards the **collapsed** inner payload; the `%%` exists only transiently on the wire to the coordinator. The node stores the collapsed (= original) line, so the checksum matches. Confirmed by reading `process_line` + the forwarding path + `append_to_startup` (chain in the verification block above).

</details>

falkoschindler (approved):
LGTM — and Evan's verification already covers the mechanism, the checksum, and the sibling, so I'll just add the one thing worth recording: this escaping is at the _right_ layer, not a bandaid.

`bus.send` is explicitly a printf-style API, so passing arbitrary data as its format string and escaping `%`→`%%` first is the idiomatic discipline (same as never writing `printf(user_string)` in C). And it has to live on the transport side: the `%` is meaningful to the _target_ node, which does the real formatting — so a single `%` must stay in the `.liz` file (the direct path needs it, and the file is transport-agnostic), and the extra escape belongs only on the coordinator-hop path, exactly where this PR puts it.

Happy to merge. No action needed on the `otb_update.py:52` sibling Evan flagged — its payload is base64, so it can't contain `%`, `"`, or `\`.

PR #221, opened 2026-06-25
espresso: flash from non-Jetson hosts even with gpiod installed

JensOgorek:
### Motivation

`espresso.py` is meant to flash an ESP32 from a Jetson, where it drives the EN/G0 pins via `gpiod`. When `gpiod` is **not** installed it already degrades to a no-op pin controller and flashes fine via esptool's DTR/RTS auto-reset (e.g. on a developer machine with a USB-serial board).

But when `gpiod` **is** installed on a non-Jetson host, the EN/G0 lines (`PR.04` / `PAC.06`) do not exist, and espresso crashed at startup while constructing the `GpioController` — so you couldn't flash from such a machine without uninstalling `gpiod`.

### Implementation

- The GPIO controller setup now degrades gracefully: if instantiating the real controller fails (chip or lines unavailable), espresso prints a short notice and falls back to the no-op `GpioController`, flashing via esptool's DTR/RTS reset — exactly the path used when `gpiod` is absent.
- `GpioControllerV1` now raises when `find_line` returns `None`, so a missing line triggers the same fallback as `GpioControllerV2` (which already raises).

On a real Jetson, where the lines resolve, nothing changes — the real controller is used as before.

### Progress

- [x] Tested on hardware (flashed an ESP32 over USB; with a stubbed `gpiod` for both the v1 and v2 APIs: missing lines → fallback + successful flash; resolvable lines → real controller path unchanged; `gpiod` absent → unchanged).

### DIFF
--- a/espresso.py
+++ b/espresso.py
@@ -38,6 +38,8 @@ class GpioControllerV1(GpioController):
             'en': chip.find_line(en),
             'g0': chip.find_line(g0),
         }
+        if None in self._lines.values():
+            raise RuntimeError(f'GPIO lines {en}/{g0} not found on gpiochip0')
 
     def request_outputs(self, consumer: str) -> None:
         for line in self._lines.values():
@@ -122,11 +124,17 @@ EN = 'PR.04'
 G0 = 'PAC.06'
 if SWAP:
     EN, G0 = G0, EN
-gpio = {
-    None: GpioController,
-    1: GpioControllerV1,
-    2: GpioControllerV2,
-}[GPIOD_VERSION](EN, G0)
+try:
+    gpio = {
+        None: GpioController,
+        1: GpioControllerV1,
+        2: GpioControllerV2,
+    }[GPIOD_VERSION](EN, G0)
+except Exception as error:  # noqa: BLE001 - any gpiod/hardware error means no controllable pins
+    # gpiod is installed but the EN/G0 pins are not available (e.g. not a Jetson):
+    # fall back to the no-op controller and rely on esptool's DTR/RTS reset for flashing.
+    print(f'No controllable EN/G0 pins ({error}); falling back to esptool reset only.')
+    gpio = GpioController(EN, G0)
 
 
 @contextmanager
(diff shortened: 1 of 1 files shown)


falkoschindler (approved):
Looks clean and nicely scoped — the fallback does exactly the right thing so you can flash from non-Jetson hosts without having to uninstall `gpiod`. 👍

I pushed two small follow-ups on top (commit 48a8647):

1. **Don't swallow real GPIO errors on a Jetson.** The broad `except Exception` would also have caught a legitimate hardware/permission error while constructing the controller on an actual Jetson and silently fallen back to the esptool-only reset path — the flash would then fail later with the generic `Flashing failed` message instead of the real cause. The fallback now only applies on non-Jetson hosts (`IS_JETSON`, derived from the existing `/etc/nv_tegra_release` check); on a Jetson the original error is re-raised.
2. **Close the V2 chip deterministically.** `gpiod.Chip` now runs in a `with` block so its fd is closed even when `line_offset_from_id` raises and the fallback kicks in.

Two asks @JensOgorek:

- Could you re-test the updated state **on a real Jetson** (flash + `enable`/`disable`/`reset`)? I want to make sure the `raise` change doesn't affect the happy path there.
- About `with gpiod.Chip(...)`: do you know why it wasn't already written that way? It makes me wonder whether some systems run an older `gpiod` v2 version that doesn't support the context manager (`__enter__`/`__exit__`) on `Chip` yet. If so, the `with` block would break there with an `AttributeError`, and we'd better go back to `try/finally` + `chip.close()`. Your test on the target hardware would show this directly.

Otherwise: approve. ✅

JensOgorek:
Safety-check on real hardware ✅

Verified the normal Jetson flash path is unaffected by this change. Tested on **rb58** (Jetson Orin, L4T R36, gpiod v2, ESP32 build via lizard-zz):

- Copied this branch's `espresso.py` onto the robot
- `sudo ./espresso.py erase --nand` → monitor confirmed the firmware was gone (no Lizard banner)
- `sudo ./espresso.py flash --nand` → robot came back up with the expected lizard-zz banner

So the refactor (`GpioControllerV2` using `with gpiod.Chip(...)` + the V1 `None` check) does not regress the production Jetson flash path. Note that being a Jetson, rb58 exercises the normal `GpioController` path, not the new non-Jetson fallback branch itself.

PR #225, opened 2026-07-03
Extract a shared Wheels base class from the three wheels modules

JensOgorek:
### Motivation

`DunkerWheels`, `ODriveWheels` and `RoboClawWheels` each inherited `Module` directly and duplicated the same scaffolding: the `width`/`linear_speed`/`angular_speed`/`enabled` properties, the enabled-sync in `step()` and the `speed`/`enable`/`disable` command flow.

This is the first half of splitting #223 (per @falkoschindler's review): a pure, behavior-preserving refactor that can be reviewed line-for-line against the old modules. The `drivable` feature lands separately in #226, stacked on this branch. #223 stays open until the robots currently depending on it are migrated to this split.

### Implementation

A new abstract `Wheels : public Module` (`wheels.h`/`wheels.cpp`) owns the common properties, the `step()` enabled-sync and the `speed`/`enable`/`disable` flow. The concrete drivetrains only implement the motor-specific hooks — `do_wheel_speeds`, `do_enable`, `do_disable`, `update_odometry`. `ODriveWheels` and `RoboClawWheels` keep their extra `power`/`off` commands; `RoboClawWheels` keeps its `m_per_tick` scaling and wrapped-u32 odometry.

**One deliberate reconciliation:** the shared `speed` gate checks the immediate `enabled` *property* — as `ODriveWheels` and `RoboClawWheels` already did. `DunkerWheels` previously gated on the once-per-`step()` `enabled` *member*, so its gate now reacts in the same cycle too. This is the only intentional behavior change, and for a nonzero `speed` it is the safer direction: a `disable` in the same cycle as a `speed` is now honored immediately for all three, instead of one extra command slipping through to still-enabled motors.

The one edge where it is marginally *less* safe: after a direct `enabled = false` property write, a same-cycle `speed(0, 0)` is now dropped rather than forwarded, so instead of an active zero-velocity command the motors coast until the next `step()` runs `do_disable()` and cuts the power stage. For statements over UART that gap is microseconds (statements are processed before module steps in the same loop iteration); for rule-driven writes it is up to one main-loop cycle (~10ms). This matches the long-standing `ODriveWheels`/`RoboClawWheels` behavior and is fine to keep; reacting to `enabled` writes eagerly would be a separate change, which the consolidation into `Wheels` actually makes easier to do later.

No new user-facing commands or properties; `module.h` and the docs are untouched.

### Progress

- [x] Tested on hardware (built clean for `esp32` and flashed to an ESP32; the wheels modules construct and take drive commands without regression).
- [x] Documentation has been updated (or is not necessary — no user-facing surface changed).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -330,7 +330,7 @@ The main loop runs every **10ms** (`delay(10)` in `app_main`). Any operation tha
 
 Modules self-register via `REGISTER_MODULE(ClassName, &create_X)` in their own `.cpp`. The macro keys off the class's `static constexpr const char *TYPE` constant — both the factory and the static `ClassName::get_defaults()` method are registered under that string. Forgetting either the macro or the `TYPE` constant causes the DSL to report `unknown module type "ClassName"`. Two modules trying to register under the same `TYPE` throw a hard error at static-init time, so collisions are caught at first boot rather than silently shadowing.
 
-The class must define `static const std::map<std::string, Variable_ptr> get_defaults()` (return an empty map if there are no defaults) and pass only the instance name to the `Module(name)` base constructor — the runtime type identity comes from RTTI on the polymorphic `Module`.
+The class must define `static const std::map<std::string, Variable_ptr> get_defaults()` (return an empty map if there are no defaults) and pass only the instance name to the `Module(name)` base constructor — the runtime type identity comes from RTTI on the polymorphic `Module`. Inheriting `get_defaults()` from an intermediate base class (e.g. `Wheels`) is fine; a subclass that adds properties must shadow `get_defaults()` and pass its result to the base constructor, so that the registered defaults always equal the constructed properties — expander proxies seed their properties exclusively from the registered `get_defaults()`, so a property that exists only in a constructor breaks proxied instances at runtime.
 
 #### `WHOLE_ARCHIVE` is mandatory
 
--- a/main/modules/dunker_wheels.cpp
+++ b/main/modules/dunker_wheels.cpp
@@ -12,68 +12,32 @@ static Module_ptr create_dunker_wheels(const std::string &name, const std::vecto
 REGISTER_MODULE(DunkerWheels, &create_dunker_wheels)
 
 const std::map<std::string, Variable_ptr> DunkerWheels::get_defaults() {
-    return {
-        {"width", std::make_shared<NumberVariable>(1.0)},
-        {"linear_speed", std::make_shared<NumberVariable>()},
-        {"angular_speed", std::make_shared<NumberVariable>()},
-        {"enabled", std::make_shared<BooleanVariable>(true)},
-    };
+    return Wheels::get_wheels_defaults();
 }
 
 DunkerWheels::DunkerWheels(const std::string name, const DunkerMotor_ptr left_motor, const DunkerMotor_ptr right_motor)
-    : Module(name), left_motor(left_motor), right_motor(right_motor) {
+    : Wheels(name), left_motor(left_motor), right_motor(right_motor) {
     this->properties = DunkerWheels::get_defaults();
 }
 
-void DunkerWheels::step() {
+void DunkerWheels::update_odometry() {
     const double left_speed = this->left_motor->get_speed();
     const double right_speed = this->right_motor->get_speed();
-
     this->properties.at("linear_speed")->number_value = (left_speed + right_speed) / 2;
     this->properties.at("angular_speed")->number_value = (right_speed - left_speed) / this->properties.at("width")->number_value;
-
-    if (this->properties.at("enabled")->boolean_value != this->enabled) {
-        if (this->properties.at("enabled")->boolean_value) {
-            this->enable();
-        } else {
-            this->disable();
-        }
-    }
-
-    Module::step();
 }
 
-void DunkerWheels::call(const std::string method_name, const std::vector<ConstExpression_ptr> arguments) {
-    if (method_name == "speed") {
-        Module::expect(arguments, 2, numbery, numbery);
-        if (!this->enabled)
-            return;
-        double linear = arguments[0]->evaluate_number();
-        double angular = arguments[1]->evaluate_number();
-        double width = this->properties.at("width")->number_value;
-        this->left_motor->speed(linear - angular * width / 2.0);
-        this->right_motor->speed(linear + angular * width / 2.0);
-    } else if (method_name == "enable") {
-        Module::expect(arguments, 0);
-        this->enable();
-    } else if (method_name == "disable") {
-        Module::expect(arguments, 0);
-        this->disable();
-    } else {
-        Module::call(method_name, arguments);
-    }
+void DunkerWheels::do_wheel_speeds(double left, double right) {
+    this->left_motor->speed(left);
+    this->right_motor->speed(right);
 }
 
-void DunkerWheels::enable() {
-    this->enabled = true;
-    this->properties.at("enabled")->boolean_value = true;
+void DunkerWheels::do_enable() {
     this->left_motor->enable();
     this->right_motor->enable();
 }
 
-void DunkerWheels::disable() {
+void DunkerWheels::do_disable() {
     this->left_motor->disable();
     this->right_motor->disable();
-    this->enabled = false;
-    this->properties.at("enabled")->boolean_value = false;
 }
--- a/main/modules/dunker_wheels.cpp
+++ b/main/modules/dunker_wheels.cpp
@@ -11,13 +11,8 @@ static Module_ptr create_dunker_wheels(const std::string &name, const std::vecto
 }
 REGISTER_MODULE(DunkerWheels, &create_dunker_wheels)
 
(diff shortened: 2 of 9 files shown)


JensOgorek:
**Reviewed with Claude Fable 5** (`/review 225`, medium effort). Three findings, all confirmed by verification, none blocking — **all applied**.

1. **Base class installs the properties it depends on.** The `Wheels` constructor now seeds `this->properties`, so the base no longer relies on every subclass remembering to assign its defaults (a missing key would throw `std::out_of_range` on the first `step()` — a `logic_error` the main loop's `runtime_error` catch would miss, boot-looping the firmware). Went the whole way Fable suggested: renamed the base defaults to a public static `Wheels::get_defaults()` and dropped the pure-forwarder `get_defaults()` in `DunkerWheels`/`ODriveWheels` (their `REGISTER_MODULE` now resolves to the inherited base static); `RoboClawWheels` keeps a shadowing one only because it adds `m_per_tick`. The private `lizard-zz` `InnotronicWheels` is updated in tandem to call `Wheels::get_defaults()`.
2. **`enabled` mirror member is now private** — only the base uses it, and a subclass writing it directly would silently suppress the enable/disable edge detection.
3. **Forward-kinematics deduplicated** into a protected `update_speeds()` helper used by all three `update_odometry()` overrides.

Built clean for `esp32` and exercised on hardware.

---

**What we'd like from you, @falkoschindler:** this is the pure, behavior-preserving half of the #223 split you asked for — reviewable line-for-line against the old `DunkerWheels`/`ODriveWheels`/`RoboClawWheels`. It isolates exactly the open question you raised: with the feature stripped out, does a shared base earn its place, or does it only work with per-subclass special-casing? Please sanity-check that the extraction is faithful and that the one deliberate reconciliation — Dunker's speed gate now reading the immediate `enabled` property, matching what ODrive/RoboClaw already did — is acceptable. The `drivable` feature lands separately in #226, stacked on this; it's not up for review yet.

falkoschindler:
Hi @JensOgorek, heads-up: this is an automated review (Claude Fable) that I triggered — I've skimmed the findings but haven't read the code myself, so please treat it accordingly.

The extraction itself verified as faithful: step ordering (odometry → enabled-sync → `Module::step()`), enable/disable semantics, the speed math including RoboClaw's `m_per_tick` factoring, and the `REGISTER_MODULE`/proxy defaults wiring all match the old modules line for line, and the declared Dunker `speed`-gate change is implemented exactly as described. No blocking bugs — the six findings below are all minor polish of the new seam.

1. **main/modules/roboclaw_wheels.cpp (line 23) — install-then-overwrite property pattern is wasteful and fragile.** The `Wheels` base ctor installs `Wheels::get_defaults()` (main/modules/wheels.cpp, line 16), then this line throws that map away and installs `RoboClawWheels::get_defaults()`, which calls `Wheels::get_defaults()` a second time — the four base Variables are allocated twice per construction. More importantly, `Wheels::step()`/`call()` unconditionally do `properties.at("enabled")`/`at("width")`, so a future subclass whose `get_defaults()` uses the plain brace-init pattern of every non-Wheels module (without the `auto defaults = Wheels::get_defaults()` merge) would throw `std::out_of_range` on every main-loop tick. A `Wheels(const std::string name, std::map<std::string, Variable_ptr> defaults = Wheels::get_defaults())` ctor would give exactly one install site, let RoboClaw pass its own defaults in the init list, and make the explanatory comment in main/modules/wheels.cpp (lines 14–16) unnecessary; the static `get_defaults()` stays for `REGISTER_MODULE`.

2. **main/modules/dunker_wheels.h (line 6) — DunkerWheels/ODriveWheels no longer define `get_defaults()`, diverging from the documented module contract and setting a proxy trap.** AGENTS.md (line 333) states "The class must define `static const std::map<std::string, Variable_ptr> get_defaults()`"; both classes now inherit it from `Wheels` instead. That works today (identical maps), but expander proxies seed their properties exclusively from the registered function (main/modules/proxy.cpp, line 13), so if a property is later added to one of these classes ctor-only — the natural move with no class-local `get_defaults()` left as a reminder — core-side proxies will lack it and `write_property`/`get_property` throw "unknown property" at runtime, only in expander setups. The ctor-parameter fix from finding 1 would make "registered defaults == constructed properties" an invariant; alternatively update AGENTS.md to bless inheriting a base-class `get_defaults()`.

3. **main/modules/wheels.h (line 28) — the private `enabled` member's role is undocumented and invites two plausible wrong "cleanups".** It holds the last value synced to the motors so `step()` can edge-detect direct writes to the `enabled` property (main/modules/wheels.cpp, line 27), but nothing says so. A maintainer could delete it as a redundant cache (reintroducing enable/disable commands every step) or gate `speed` on it "for consistency" with the motor modules, which do gate on the member (reintroducing the one-cycle lag this PR deliberately removed). A rename like `last_applied_enabled` or a one-line comment would pin it down.

4. **main/modules/wheels.h (line 16) — class section order matches neither the documented nor the de-facto convention.** The new class orders protected → private → public; CONTRIBUTING.md (line 73) says public first, and every sibling header (module.h, input.h, linear_motor.h, odrive_motor.h) uses private → protected → public. Also, within the public section the static `get_defaults()` precedes the constructor, against CONTRIBUTING.md (line 74, "constructors/destructors, then methods") and the AGENTS.md module template. As the base class new modules will imitate, it's worth matching the de-facto private → protected → public order.

5. **main/modules/wheels.cpp (line 27) — non-blocking follow-up: the enabled-sync pattern now has a shared home, but five motor modules still carry their own copies.** The same edge-detect plus dual-write bookkeeping exists in d1_motor.cpp (line 149), dunker_motor.cpp (line 188), linear_motor.cpp (line 40), mks_servo_motor.cpp (line 186) and canopen_motor.cpp (line 562). The copies differ in details (CanOpenMotor's `initialized` guard and extra property write; the motors gate `speed` on the member rather than the property), so hoisting into `Module` needs design work and clearly doesn't belong in this behavior-preserving split — but it may deserve a follow-up issue so the next enabled-capable module doesn't add a seventh copy.

6. **main/modules/wheels.cpp (line 21) — optional hardening: `update_speeds()` divides by the `width` property with no zero guard.** A script can set `wheels.width = 0` (property writes are unvalidated, main/modules/module.cpp, lines 117–119), after which `angular_speed` becomes inf/NaN in the broadcast telemetry every step. This is pre-existing in all three modules and requires a nonsensical width, so it's not a change request for this PR — but the consolidation makes a single guard fix all three drivetrains now.

JensOgorek:
Thanks @falkoschindler — findings addressed in cbbdf50 (applied with Claude Fable 5, built locally):

1. **Single install site for the defaults** (finding 1): the `Wheels` constructor now takes the defaults map as a parameter, defaulting to `Wheels::get_defaults()`. Dunker/ODrive pass nothing and get the base map; `RoboClawWheels` passes its shadowed `get_defaults()` in the init list instead of overwriting the base's map in the ctor body. The double allocation and the explanatory comment are gone, and there's no overwrite pattern left for the next subclass to copy.
2. **Module contract** (finding 2): rather than reintroducing pure-forwarder `get_defaults()` in Dunker/ODrive, AGENTS.md now explicitly blesses inheriting `get_defaults()` from an intermediate base — with the rule that a subclass adding properties must shadow it *and* pass the result to the base constructor, plus the proxy rationale (expander proxies seed exclusively from the registered function, so ctor-only properties break proxied instances). Together with finding 1's fix, the natural path now keeps registered defaults == constructed properties.
3. **`enabled` member** (finding 3): renamed to `last_applied_enabled` with a one-liner documenting its role — it's what turns direct writes to the `enabled` property into exactly one `do_enable()`/`do_disable()` instead of per-tick commands on the bus.
4. **Section order** (finding 4): `wheels.h` now follows the de-facto private → protected → public order with the constructor before the static `get_defaults()`; `wheels.cpp` matches.
5. Filed as #232 — agreed it needs design work (the member-vs-property gating difference and CanOpen's `initialized` guard) and doesn't belong in this split.
6. Skipping for now: property writes are unvalidated everywhere in Lizard (the neighboring `m_per_tick` and `d_micros` divisions have no guards either), so a lone `width` guard would be a foreign body — and a behavior change in a deliberately behavior-preserving PR. Happy to revisit as a follow-up if you think it's worth it.

The original ask still stands, by the way: since your review round was automated too, we'd still value your human judgment on the core question — whether the shared base earns its place now that the feature is stripped out.

falkoschindler (approved):
Nice, clean extraction — Fable reviewed it line-for-line against the three old modules and the `Wheels` base is a faithful consolidation: enable/disable ordering, the odometry bookkeeping, `m_per_tick` scaling and the `power`/`off` extras all carry over unchanged, and the registry/proxy defaults contract holds (inherited `get_defaults()` for Dunker/ODrive, shadowed-and-passed-up for RoboClaw, matching the new AGENTS.md rule).

The one behavioral delta is the `DunkerWheels` speed gate, and with the updated description it's now accurately framed — thanks for capturing the same-cycle `enabled = false; speed(0, 0)` coast edge.

LGTM 👍

PR #226, opened 2026-07-03
Add a locked safety interlock to the Wheels base class

JensOgorek:
### Motivation

We need a way to block driving without switching the motors off — e.g. our delta-arm drivetrain must only drive once the arm is parked, but the motors should stay enabled and holding the whole time.

This is the second half of splitting #223 (per @falkoschindler's review), stacked on the pure refactor in #225. It adds only the `locked` feature, so it can be reviewed on its own. #223 stays open until the robots currently depending on it are migrated to this split.

### The contract: `locked` vs. `enabled`/`disable`

- **`enabled` / `disable` — the ignition.** Switches the motors off entirely.
- **`locked` — a safety interlock.** While it is `true`, drive commands are ignored and the wheels are actively held at standstill with a zero-speed setpoint — the motors stay enabled. Driving resumes as soon as `locked` is `false` again.

The condition that asserts the lock lives entirely in the robot's rules (e.g. "lock the wheels while the arm is not in its end stops"); Lizard only provides the generic gate. The whole point is the difference from `disable`: `locked = true` holds the robot in place rather than letting it coast or switching the motors off.

`locked` only blocks commands: `disable()` still switches the motors off, and a locked but switched-off robot can be pushed by hand. While `locked` is `true`, `off()` does not stick — the hold re-engages within about a second; call `disable()` to switch the motors off durably.

*(The property was called `drivable` in earlier revisions of this PR; it was renamed per review because `locked` — the input-lock reading, and literally the adjective of "interlock" — reads correctly in every reachable state.)*

### Implementation

- **New `locked` property** (default `false`) on all three wheels modules.
- **One shared gate for every drive path.** `may_drive()` (true only while enabled *and not* locked) guards `speed` and — for `ODriveWheels`/`RoboClawWheels` — `power`, so both honor the same stop semantics.
- **Active standstill hold in `step()`.** While enabled but locked, the wheels are held with a zero-speed setpoint. The hold is sent once on the rising edge of `locked` and refreshed every 100 cycles (~1 s) — it engages even when no drive command arrives (e.g. the rule fired because the host went silent) and re-asserts after a motor controller reboot, without flooding the bus.
- **The hold is unconditional while enabled and locked.** `off()` does not stick while locked — the hold re-engages with the next low-rate refresh (~1 s); only `disable()` switches the motors off durably. (An earlier revision suspended the hold after `off()`; dropped per review in favor of the simpler contract, which is also the safer one for an interlock.)
- **Shadow forwarding for both gate properties.** Shadowed wheels mirror method calls but not property writes, so `Wheels::write_property` forwards `locked` *and* `enabled` to the shadow modules — otherwise a shadow would keep driving after its master was locked or disabled. Forwarding is one level deep (via `Module::write_property`, not the shadow's own override), matching the single-level `call_with_shadows` and avoiding recursion on shadow chains or mutual shadows. On `shadow()`, the gate properties are synced once onto the freshly attached shadow (master-wins), so a late-attached shadow doesn't start with stale gate values. A single `GATE_PROPERTIES` list feeds both mechanisms.
- `Module::shadow_modules` is now `protected` so `Wheels` can forward to its shadows.

### Usage

Our delta-arm drivetrain relies on this: a brain-side rule sets `wheels.locked` while the arm is not in its end stops, so the robot can only drive once the arm is parked. Because both track modules are shadowed (`wheels.shadow(wheels_front)`), the shadow forwarding locks both at once.

### Notes for the release notes (0.13.0)

Two behavior changes for existing shadow setups, both intended and reviewed on this PR — please mention them when this ships:

- Direct writes to `wheels.enabled` now reach shadow modules. Previously only the `disable()` method call was forwarded (via `call_with_shadows`); a plain `wheels.enabled = false` left shadows running.
- On `wheels.shadow(other)`, the master's gate properties (`locked`, `enabled`) are synced onto the freshly attached shadow, master-wins — overwriting any values pre-set on the shadow before attaching.

### Progress

- [x] Tested on hardware (built clean for `esp32` with IDF 5.3, flashed to a bench ESP32; verified over the serial console, 13/13 checks: the `false` default, the hold on the rising edge of `locked` plus the ~1 s refresh, `speed()` ignored while locked, `off()` not sticking (hold re-engages within ~1 s), driving resuming after unlock, gate-property forwarding to shadows, attach-time sync, and `disable()` stopping the hold durably). Physical holding force needs a real motor controller on the bus and was not measured. The `disable()`+`enable()`-in-one-cycle edge (d0daf47) was additionally A/B-verified on the bench: pre-fix firmware reproduces the missed rising edge, the fix re-engages the hold on the next cycle.
- [x] Documentation has been updated (the `locked` interlock is documented for each wheels module with a short property cell, an explanatory paragraph below the table and an explicit precedence sentence).

### DIFF
--- a/.claude/skills/release-notes/SKILL.md
+++ b/.claude/skills/release-notes/SKILL.md
@@ -1,0 +1,106 @@
+---
+name: release-notes
+description: Write Lizard release notes for a milestone to tmp.md. Use when asked to draft release notes or a release announcement for a given milestone/version (e.g. "write release notes for 0.12.0").
+---
+
+Write release notes for the milestone passed as an argument (e.g. `0.12.0`) to `tmp.md`.
+
+The milestone title is the version (e.g. `0.12.0`). If given only a milestone URL or number, resolve its title first: `gh api repos/zauberzeug/lizard/milestones/<n> --jq .title`.
+
+## How to gather data
+
+**Start with the helper script** [`release-notes.sh`](release-notes.sh) (in this skill's directory), which collects the deterministic data so nothing is forgotten:
+
+- `release-notes.sh dossier "<milestone>"` → one JSON dossier: every milestone issue/PR with labels, author, committers, reviewers, commenters, and extracted cross-reference `#`s; each cross-reference resolved to issue/PR/**discussion** with its author (discussions are the easiest contributor to miss).
+- `release-notes.sh verify "<milestone>" tmp.md` → milestone tickets missing from the draft (and `#`s in the draft that aren't milestone tickets — expected for cross-refs).
+- `release-notes.sh check-docs <slug> [slug...]` → HTTP status per docs page, to verify links before using them.
+
+The dossier gives you the raw facts; you still apply judgment for grouping tickets into stories, writing descriptions, and deciding which commenters are substantive. The manual `gh` recipes below remain valid as a fallback or for spot-checks.
+
+1. Use `gh` to list all issues and PRs in the milestone with their labels, linked PRs, authors, and participants.
+2. For each issue/PR, inspect the timeline and comments to identify contributors (see contributor rules below).
+3. For each merged PR, also fetch its reviewers and committers explicitly — `gh pr view <n> --json reviews,commits` — and fold them into the contributor list. Comment scraping alone misses reviewers who only left a formal review. (Do not credit the merger; see Contributors.)
+4. Resolve cross-referenced numbers that are **discussions**, not issues/PRs. `gh issue view <n>` / `gh pr view <n>` fail on a discussion number; query it via GraphQL instead, e.g. `gh api graphql -f query='{repository(owner:"zauberzeug",name:"lizard"){discussion(number:N){title author{login}}}}'`, and credit the discussion author.
+5. Check if any PRs reference a GHSA (GitHub Security Advisory) — these go into a separate Security section.
+
+## Structure
+
+Follow the exact structure used in prior releases at https://github.com/zauberzeug/lizard/releases (curated sections with `###` headings and one item per story — not GitHub's auto-generated "What's Changed" list).
+
+### Section mapping (by label)
+
+| Label / nature     | Section heading               |
+| ------------------ | ----------------------------- |
+| (GHSA-based PRs)   | Security                      |
+| `enhancement`      | New features and enhancements |
+| `bug`              | Bugfix                        |
+| `documentation`    | Documentation                 |
+| (CI/tooling/build) | Infrastructure                |
+
+Lizard has no `feature`, `infrastructure`, `testing`, or `dependencies` labels, so use judgment: CI, build, pre-commit, and packaging work goes under **Infrastructure** even when it carries the `enhancement` label. Items double-labelled `enhancement`+`documentation` usually belong under whichever is the dominant change. If something doesn't fit, create an appropriate section following the established pattern.
+
+Security section (if any) always comes first. Then the remaining sections in the order listed above.
+Security items start with a "⚠️" prefix (e.g. `⚠️ Prevent memory exhaustion via media streaming routes`).
+
+### Documentation links
+
+When an item mentions a DSL element or concept that has a documentation page, you may link it, e.g. [`bus.send`](https://lizard.dev/module_reference/) — prior Lizard releases mostly omit doc links, so keep this light and only add a link when it clearly helps.
+
+**Important:** Do not guess URLs. Verify each link exists by fetching it (the `check-docs` helper does this in bulk). Lizard docs pages live at `https://lizard.dev/<slug>/` (trailing slash; `/<slug>` 301-redirects to it).
+If a page doesn't exist yet (e.g. for something added in this release), ask the user whether to include the link anyway (it will go live after deployment).
+
+### Item format
+
+Each line follows this pattern:
+
+```
+- Short description of the change (#issue, #pr by @author1, @author2, @contributor3)
+```
+
+- One item per story (a story is usually an issue with a PR that fixes it, but may include follow-up fixes)
+
+- Check PR descriptions for references to feature requests, discussions, or other issues that led to the change — include those ticket numbers too
+
+- Group related tickets into a single item — all ticket numbers for the milestone should appear somewhere
+
+- Ticket numbers go in parentheses: `(#123, #456 by @user1, @user2)`
+
+- If a PR has a breaking change, add a `**Breaking change:**` block below the item, like this:
+
+  ```
+  - Short description of the change (#issue, #pr by @author1, @author2)
+
+      **Breaking change:** Explanation of what changed and how to migrate.
+      For example, if an API was removed, show the old usage and the new replacement.
+  ```
+
+### Contributors
+
+For each item, mention all involved contributors in the `by @...` list:
+
+- Issue author
+- **The author of any feature request / discussion that led to the change** (often referenced in the PR description), plus anyone who contributed substantively to that discussion — not only the issue/PR participants
+- Users with relevant contributions to discussions (substantive comments, reproductions, debugging - not just "me too" comments)
+- **Reviewers** who provided relevant feedback — include both review submissions (`gh pr view <n> --json reviews`) **and** inline review comments, not just top-level issue/PR comments. A formal review with no top-level comment is easy to miss with comment scraping alone.
+- Committers (anyone who pushed commits to the PR), ignoring AI/bot co-author logins such as `claude`
+
+Do **not** credit the merger as such — merging a PR is not, on its own, a contribution worth listing.
+
+**Make sure no contributor is forgotten** — when a borderline commenter is genuinely engaged (offered to help, attempted a fix, added environment detail), lean toward including them rather than filtering them out. Only ignore:
+
+- Simple "me too" comments with no added information
+- Review approvals without valuable comments or own commits
+
+## Verification
+
+After writing `tmp.md`, verify completeness — `release-notes.sh verify "<milestone>" tmp.md` does steps 1-3:
+
+1. Use `gh` to list all issue and PR numbers in the milestone.
+2. Extract all `#nnn` ticket numbers from the generated `tmp.md`.
+3. Compare the two sets. Report any milestone tickets missing from the release notes.
+4. If there are missing tickets, investigate them and either add them to the appropriate section or explain why they were excluded (e.g. duplicates, reverted, not actually resolved).
+
+## Output & posting
+
+- Write the draft to `tmp.md` in the project root (a leading `# Release notes for Lizard x.y.z` H1 is fine as a working title — strip it before posting to GitHub, where the tag/version is already the heading).
+- Do not create, edit, or publish the GitHub release yourself. Point at `tmp.md` and let the user post it (e.g. by editing the release for the tag with `gh release edit v<x.y.z> --notes-file tmp.md` after confirmation).
--- a/.claude/skills/release-notes/release-notes.sh
+++ b/.claude/skills/release-notes/release-notes.sh
@@ -1,0 +1,112 @@
+#!/usr/bin/env bash
+# Deterministic data helper for the release-notes skill (see SKILL.md).
+# It gathers the parts a script can get right every time, so the model only has to
+# do the judgment work (grouping tickets into stories, prose, substantive-commenter
+# calls, section placement).
+#
+# Usage:
+#   release-notes.sh dossier "<milestone>"            # full JSON dossier (default)
+#   release-notes.sh verify  "<milestone>" [file]     # milestone set vs #s in file (default tmp.md)
+#   release-notes.sh check-docs <slug> [slug...]      # HTTP status for each docs page
+#
+# Override the repo via env if ever reused elsewhere:
+#   RELEASE_NOTES_REPO=owner/repo
+set -euo pipefail
+
+REPO="${RELEASE_NOTES_REPO:-zauberzeug/lizard}"
+
+die() { echo "error: $*" >&2; exit 1; }
+command -v gh >/dev/null || die "gh not found"
+command -v jq >/dev/null || die "jq not found"
+
+milestone_numbers() {
+  local milestone="$1"
+  {
+    gh issue list -R "$REPO" --milestone "$milestone" --state all --limit 300 --json number --jq '.[].number'
+    gh pr list   -R "$REPO" --search "milestone:\"$milestone\"" --state all --limit 300 --json number --jq '.[].number'
+  } | sort -un
+}
+
+dossier() {
+  local milestone="${1:?milestone required}"
+
+  local issue_nums pr_nums
+  issue_nums=$(gh issue list -R "$REPO" --milestone "$milestone" --state all --limit 300 --json number --jq '.[].number')
+  pr_nums=$(gh pr list -R "$REPO" --search "milestone:\"$milestone\"" --state all --limit 300 --json number --jq '.[].number')
+  [ -n "${issue_nums}${pr_nums}" ] || die "no tickets found for milestone '$milestone' (check the exact name)"
+
+  local tickets='[]' n obj
+  for n in $issue_nums; do
+    obj=$(gh issue view "$n" -R "$REPO" --json number,title,state,author,labels,body,comments --jq '{
+      number, title, state, kind: "issue",
+      author: (.author.login? // null),
+      labels: [.labels[].name],
+      commenters: ([.comments[] | .author.login? // empty] | unique),
+      refs: ([(.body // "") | scan("#[0-9]+") | ltrimstr("#")] | unique)
+    }')
+    tickets=$(jq -c --argjson t "$obj" '. + [$t]' <<<"$tickets")
+  done
+  for n in $pr_nums; do
+    obj=$(gh pr view "$n" -R "$REPO" --json number,title,state,author,labels,body,comments,reviews,commits --jq '{
+      number, title, state, kind: "pr",
+      author: (.author.login? // null),
+      labels: [.labels[].name],
+      committers: ([.commits[].authors[]? | .login? // empty] | unique),
+      commenters: ([.comments[] | .author.login? // empty] | unique),
+      reviewers: ([.reviews[] | .author.login? // empty] | unique),
+      refs: ([(.body // "") | scan("#[0-9]+") | ltrimstr("#")] | unique)
+    }')
+    tickets=$(jq -c --argjson t "$obj" '. + [$t]' <<<"$tickets")
+  done
+
+  # Cross-references that point outside the milestone: classify issue/pr/discussion
+  # and resolve the author, since discussion numbers (feature requests) are invisible
+  # to `gh issue view` / `gh pr view` and are the easiest contributor to miss.
+  local in_ms extra cross='[]' ref kind author
+  in_ms=$(jq -c '[.[].number | tostring]' <<<"$tickets")
+  extra=$(jq -r --argjson ms "$in_ms" '[.[].refs[]] | unique | map(select(IN($ms[]) | not)) | .[]' <<<"$tickets")
+  local resp
+  for ref in $extra; do
+    if resp=$(gh api "repos/$REPO/issues/$ref" 2>/dev/null); then
+      kind=$(jq -r 'if .pull_request then "pr" else "issue" end' <<<"$resp")
+      author=$(jq -r '.user.login // "null"' <<<"$resp")
+    else
+      kind="discussion"
+      author=$(gh api graphql -f query="{repository(owner:\"${REPO%/*}\",name:\"${REPO#*/}\"){discussion(number:$ref){author{login}}}}" --jq '.data.repository.discussion.author.login // "null"' 2>/dev/null || echo null)
+    fi
+    cross=$(jq -c --arg n "$ref" --arg k "$kind" --arg a "$author" '. + [{number: ($n | tonumber), kind: $k, author: $a}]' <<<"$cross")
+  done
+
+  jq -n --arg ms "$milestone" --argjson tickets "$tickets" --argjson cross "$cross" \
+    '{milestone: $ms, tickets: $tickets, cross_references: $cross}'
+}
+
+verify() {
+  local milestone="${1:?milestone required}" file="${2:-tmp.md}"
+  [ -f "$file" ] || die "file not found: $file"
+  local ms rel
+  ms=$(milestone_numbers "$milestone")
+  rel=$(grep -oE '#[0-9]+' "$file" | tr -d '#' | sort -un)
+  echo "Milestone tickets MISSING from $file (investigate each):"
+  comm -23 <(echo "$ms") <(echo "$rel") | sed 's/^/  #/' || true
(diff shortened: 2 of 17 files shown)


falkoschindler (changes requested):
This review comes in two parts: Part 1 collects my own findings, Part 2 the findings from a Fable review pass. Where the two independently landed on the same spot, the overlap is intentional.

## Part 1 — Falko's findings

**1. docs/module_reference.md — the `drivable` cell is far too long.** The property description blows the table cell wide open, repeated across all three wheels modules. If the property can't be explained briefly, the concept probably isn't crisp enough — or it needs a proper paragraph of explanation below the table instead of one overloaded cell.

**2. odrive_wheels.cpp (line 54) and roboclaw_wheels.cpp (line 75) — why an early `return` instead of just swapping the condition?** The old code guarded the body with `if (enabled) { ... }`. Turning it into `if (!may_drive()) { return; } ...` changes the structure; replacing the condition in place — `if (may_drive()) { ... }` — would be a +1/−1 change and keep the existing shape. What's the reason for the early-exit restructure?

**3. wheels.cpp (`step()`, handbrake block) — why doesn't `step()` use its own `may_drive()`?** The hold condition is open-coded as `enabled && !drivable` instead of going through the new helper. Even if it isn't literally `!may_drive()` (a disabled module shouldn't brake), having two hand-written encodings of the same gate is exactly what the helper was meant to prevent — they can drift.

**4. The "handbrake" metaphor is never cashed out.** The docs lean on "handbrake", but a differential drive has no independent brake the way a car's handbrake is independent of the engine. The motor is commanded to — what, exactly? A zero-velocity setpoint in closed loop? An active holding torque? This still isn't clear from the documentation, and "handbrake" implies a mechanism that isn't there.

**5. wheels.h — if it's "a handbrake", why is the property called `drivable`?** The documentation talks about brakes and handbrakes throughout, but the property is named `drivable`. Pick one: either call it `brake` to match how it's described, or drop the brake language and document it consistently in terms of drivability. The mismatch between the name and the prose is confusing.

**6. wheels.cpp (`Wheels::write_property`) — this looks like a breaking change.** Overriding `write_property` to forward properties to shadows changes base `Module` write behavior, but only for wheels. If it isn't breaking and this is the right behavior for shadowed modules, why not do it in `Module` for all modules instead of just here? And if there's a reason it can't be general, what makes wheels special — and is there a risk it breaks existing shadow setups?

## Part 2 — Fable's findings

**1. main/modules/wheels.cpp (line 41) — The handbrake loop silently undoes `off()`, contradicting the documented contract.** With `enabled = true` and `drivable = false`, calling `wheels.off()` puts the ODrive axes into IDLE (or RoboClaw into duty 0), but `off()` doesn't touch the `enabled` property, so the very next `step()` hits `enabled && !drivable` and calls `do_wheel_speeds(0, 0)` — `ODriveMotor::speed(0)` re-sends `set_mode(8, ...)` and the axes go straight back into closed-loop hold, one cycle (~10 ms) after `off()`. The docs added in this PR explicitly promise "`disable()` and `off()` still switch the motors off regardless" (docs/module_reference.md, line 541) — that currently only holds for `disable()`. The same mechanism also fights direct per-motor commands (e.g. a `left_motor.position(x)` service jog thrashes control modes every cycle while the handbrake is engaged). Either `off()` needs to stick (e.g. the handbrake loop stands down until the next drive/enable event) or the docs need to say that `off()` cannot be used while the handbrake is engaged.

**2. main/modules/wheels.cpp (line 41) — On RoboClawWheels, the handbrake loop turns `step()` into a blocking, throwing path when the serial link is dead.** `RoboClawMotor::speed` retries up to 4 times, each attempt blocking on serial reads (~600 ms worst case per motor), and throws after the final failure. Pre-PR, `Wheels::step()` never touched the serial bus (odometry reads the cached position property), so the wheels module always broadcast its odometry. Now, with the handbrake engaged and an unpowered/flaky RoboClaw — exactly the "host went silent" scenario the feature targets — every cycle blocks for hundreds of ms, throws before reaching `Module::step()`, and the host stops receiving `linear_speed`/`angular_speed` for as long as the hold lasts.

**3. main/modules/wheels.cpp (line 69) — The shadow-attach sync can clobber the wrong module: `shadow_modules.back()` is not the freshly attached shadow when the push was skipped.** `Module::call("shadow")` silently skips the `push_back` for a self-shadow (`this == target`), but the sync only checks `!shadow_modules.empty()`. So after `a.shadow(b)`, then `b.drivable = false` (set directly on the shadow), a later `a.shadow(a)` — a harmless no-op before this PR — overwrites b's `drivable`/`enabled` with a's values, silently releasing b's handbrake (and b's `step()` edge-detect will act on a flipped `enabled`). Sync only when a push actually happened, e.g. compare the list size before/after the `Module::call`, or evaluate the target module and sync it directly when it isn't `this`.

**4. main/modules/wheels.cpp (line 41) — Re-sending `do_wheel_speeds(0, 0)` at the full 100 Hz loop rate is unnecessary bus traffic.** The main loop runs every 10 ms, and none of the backends dedupe: steady-state that's 200 CAN frames/s per ODrive wheels module (~9% of a 250 kbit/s bus), 400 CANopen frames/s on Dunker, and blocking serial transactions competing with encoder polls on RoboClaw — for hours, on a robot that is parked. I found no device watchdog that needs 100 Hz feeding; velocity mode holds its setpoint. Sending on the `drivable` falling edge (mirroring the existing `last_applied_enabled` idiom a few lines above) plus a low-rate refresh (e.g. every 10th–100th cycle, which also keeps the nice re-assert-after-ODrive-reboot property) preserves the stated intent and cuts the traffic by 90–99%. This would also confine finding 2 to an occasional retry instead of every cycle.

**5. main/modules/wheels.cpp (line 70) — Attach-time sync is unconditionally master-wins, which can release a safety hold on the new shadow.** If module a currently holds `drivable = false` (or is disabled) and a script then runs `b.shadow(a)` with b at defaults, the sync sets `a.drivable = a.enabled = true` and a's edge-detect re-enables its motors — the hold on a is cleared as a side effect of attaching it. Since `shadow()` is normally a config-time construct this is unlikely in practice, and master-wins is arguably the intended contract ("a now follows b"), so this may be fine as designed — but if the sync only propagated the restrictive direction (`false`), the attach could never accidentally release a hold. Worth a deliberate decision.

**6. main/modules/wheels.cpp (lines 70 and 84) — The gate-property list lives in two places with two mechanisms.** The write forwarding checks `property_name == "drivable" || property_name == "enabled"` while the attach sync pokes the same two properties via `boolean_value` on `back()`. A small private helper (e.g. `sync_gate_properties(shadow)` copying the current values, called from both sites) would leave exactly one list and one mechanism — and naturally fixes finding 3 if it takes the actual target module. (The bigger question — moving shadow property forwarding into `Module` — was already discussed and deliberately declined for this stacked PR, so I'm not re-raising it.)

**7. main/modules/odrive_wheels.cpp (line 52) and main/modules/roboclaw_wheels.cpp (line 73) — The gated `power` branch is now byte-for-byte identical in both subclasses**, and this PR had to edit both copies in lockstep. Hoisting it into `Wheels::call` next to `speed` (dispatching to a `do_wheel_power` hook, like `do_wheel_speeds`) would keep the gate in one place — though it needs care so DunkerWheels still errors on an unknown `power` method rather than silently no-op'ing, and both subclasses keep their `call` overrides for the differing `off` anyway. Optional; leaving `power` subclass-local may have been a considered scoping choice in the base-class extraction.

**8. main/modules/wheels.cpp (line 78) and main/modules/wheels.h (line 42) — Two added lines exceed the 120-character limit** from CONTRIBUTING.md ("Line length: 120 characters maximum"): the `write_property` definition (126 chars) and its declaration (138 chars). Mitigating: the declaration mirrors the pre-existing 137-char virtual in module.h, and .clang-format doesn't enforce a column limit — but these two are new.

## Conclusion

Across both parts the common thread is that the underlying concept still hasn't come through clearly: the "handbrake" framing, the `drivable` name, and what the motor is actually commanded to do while held all pull in slightly different directions. On top of that there are a fair number of rough edges still to settle — the `off()` interaction, the shadow-sync behavior, the continuous braking loop. None of it looks fundamental, though. I'm convinced there's a clear concept underneath driving all of this; once we pin it down and let it drive the naming, the docs, and the implementation, most of these edges should fall out on their own.

JensOgorek:
Thanks for the thorough two-part review! We went through everything together and pinned down the concept first — and as you suspected, most of the edges fell out on their own once it was clear. Everything below landed in 230bf87.

**The concept.** `drivable` is a **safety interlock**, not a parking brake — that's where the "handbrake" framing went wrong. The real-world scenario: the robot must only drive while the tool is stowed. On the Feldfreund, a rule clears `drivable` while the delta arm is not in its end stops (driving with a deployed tool is how we lost a weeding drill recently). The condition lives entirely in the robot's rule — today "arm in end stops", later perhaps "arm side above some angle" — and changing it is a one-line rule edit, no Lizard change. Lizard only provides the generic gate. So the name `drivable` stays (it describes the intent: permission to drive), the brake language is gone everywhere, and the docs now spell out the mechanism: drive commands are ignored and the wheels are actively held at standstill with a zero-speed setpoint; the motors stay enabled.

**Your findings:**

1. *Docs cell too long* — fixed: the cell is now one short line, with a proper paragraph below each of the three tables explaining the behavior.
2. *Early return vs. swapped condition* — no deeper reason; swapped the condition in place (`if (may_drive())`) in both subclasses.
3. *`step()` not using `may_drive()`* — the hold condition is now derived from the helper: `enabled && !may_drive()`.
4. *"Handbrake" never cashed out* — see concept above; the docs now state explicitly what the motors are commanded to (zero-speed setpoint in closed loop, motors stay enabled).
5. *`drivable` vs. brake naming* — resolved by dropping the brake language, not the name (see concept). Renaming to `brake` would name the mechanism instead of the intent.
6. *`write_property` breaking change* — it stays in `Wheels` deliberately: generalizing property forwarding into `Module` was already discussed and deferred, since it would change behavior for **all** shadowed modules. For shadowed wheels the changed behavior (gate property writes forward) is precisely the feature. And nothing is "breaking" yet anyway — all consumers of this are still in development.

**Fable's findings:**

1. *Hold loop undoes `off()`* — fixed, `off()` wins: it stands the hold down (small `hold_suspended` flag) until the next `enable()` or `drivable` change. The docs promise now actually holds, and per-motor service jogs are no longer fought every cycle.
2. + 4. *RoboClaw blocking / 100 Hz bus traffic* — fixed together as suggested: the hold is sent once on the falling edge of `drivable` and refreshed every 100 cycles (~1 s) instead of every cycle. Keeps the re-assert-after-reboot property, cuts steady-state traffic by ~99 %, and confines the dead-serial RoboClaw case to ~1 Hz retries.
3. *`shadow_modules.back()` clobber on self-shadow* — fixed: the attach sync only runs when `Module::call` actually attached a shadow (size check before/after).
4. *(see 2.)*
5. *Master-wins attach sync* — deliberate decision: attach stays master-wins ("a now follows b"); `shadow()` is a config-time construct, so the accidental-release scenario doesn't occur in practice.
6. *Gate list in two places* — fixed: a single `GATE_PROPERTIES` list now feeds both the write forwarding (`is_gate_property`) and the attach sync (`sync_gate_properties`).
7. *Hoisting the `power` branch* — left as is; keeping `power` subclass-local was a scoping choice of the base-class extraction, and both subclasses keep their `call` overrides for `off` anyway.
8. *Line lengths* — fixed, both new lines wrapped.

Built successfully for esp32 (IDF 5.3) and verified on a bench ESP32 over the serial console (no motors attached): gate-property forwarding to shadows, the self-shadow no-clobber fix, attach sync onto a freshly attached shadow, `enabled` forwarding, and the gated speed/power/off/enable flow — 9/9 checks green.

falkoschindler:
@JensOgorek I think the implementation converged and I can't find any more bugs. But I have a final suggestion: What about renaming `drivable` to `locked`? Here is how I landed there, including the candidates that lost:

**Why not `drivable`:** It names the permission, but hides the defining behavior. The whole point of this feature over `disable()` is that the wheels are *actively held* at standstill — yet `drivable = false` doesn't tell the reader whether the robot coasts, brakes, or holds. It can also be misread as a computed status ("can the robot drive right now?"), which it isn't: it can be `true` while `enabled` is `false`, because it's an independent input, not the combined answer.

**Why not `hold`:** It was the obvious next candidate — it names the mechanism directly and matches the vocabulary the implementation already uses. But it lies in exactly one state, and it's the state this PR carefully engineered: after `off()`, the hold is suspended and the robot can be pushed by hand, while a property literally named `hold` still reads back `true`. For a safety-flavored flag, a name that makes a false promise of physical restraint in precisely the suspended case is the wrong trade. It also reads like a command rather than a state, next to state adjectives like `enabled`.

**Why `locked`:** It keeps the best property of each. Under the everyday input-lock reading (a child lock, a control lock: *inputs are refused*), `locked = true` is a true statement in every state — including after `off()`, where drive commands are still ignored. At the same time it carries the active-restraint connotation that `drivable` lacked, and while the motors are powered it actually delivers on it. It's a state adjective like `enabled`, the asserting rule reads as the intent ("lock the wheels while the arm is up": `wheels.locked = true`), and it's literally the adjective of "interlock" — the term the docs already use for this feature. No collisions either: the only "lock" in the codebase is the internal C++ `InterpreterLock`, invisible at the language level.

**What the rename touches:** default `false` instead of `true`, `may_drive()` becomes `enabled && !locked` (main/modules/wheels.cpp (line 35)), the entry in `get_defaults` and `GATE_PROPERTIES` (main/modules/wheels.cpp (lines 9 and 14)), the doc tables and paragraphs for the three wheels modules, and the PR title/body. The shadow-sync and hold machinery carry over unchanged. The docs should keep one explicit precedence sentence — `locked` only blocks commands; `disable()`/`off()` still switch the motors off, and a locked-but-off robot can be moved by hand — but that now reads as ordinary precedence documentation rather than an apology for the name.

Not a blocker if you disagree — the semantics are the hard part and they're right — but I think `locked` is the name that still reads correctly in every state the state machine can reach.

JensOgorek:
Sold — `locked` it is (8fd8df3). The child-lock reading was what convinced us: it states the truth in every state, including the suspended-hold case after `off()`, and "adjective of interlock" seals it.

Renamed exactly along your list: default flips to `false`, `may_drive()` is now `enabled && !locked`, `get_defaults`/`GATE_PROPERTIES`, the three doc sections (each now with the explicit precedence sentence), and the PR title/body. The hold and shadow-sync machinery is untouched.

Re-verified on the bench ESP32 with the inverted logic: 12/12 checks green, including two new ones (`locked` defaults to `false` after construction; driving resumes cleanly after unlocking).

falkoschindler:
@JensOgorek the rename landed cleanly, and I still can't find anything wrong in the machinery — the gate, the shadow forwarding and the attach sync all held up under another pass. But now that the full contract is in front of us, I want to revisit one decision that came out of my own review: the `off()`-wins suspension (Fable finding 1). My original finding offered two ways out — make `off()` stick, or document that `off()` doesn't stick while held — and we picked the first. Having seen what it costs, I now think the second is the better trade.

## Suggestion: drop the hold suspension entirely

Concretely: remove `hold_suspended`, `suspend_hold()` and the `suspend_hold()` calls in the two subclass `off()` handlers. The contract collapses to one sentence: **while `enabled && locked`, the wheels hold; `disable()` switches them off.**

Why I've changed my mind:

- **The pain that motivated `off()`-wins is mostly gone anyway.** The original finding was written against the 100 Hz hold loop: `off()` was undone ~10 ms later and per-motor service jogs were fought every cycle. The low-rate refresh from findings 2/4 already fixed the aggressive part — without the suspension, re-engagement would now happen within ~1 s, once. What's left of the original problem is thin, and the suspension is a whole extra contract to pay for it.
- **The suspension has its own timing edge.** `hold_suspended` is only cleared when a `step()` observes the unlocked state (main/modules/wheels.cpp, lines 60–62) or in `enable()`. So `wheels.off(); wheels.locked = true` within one cycle leaves the hold down, while the same sequence across two cycles re-engages it — whether the documented "or `locked` changes" happens depends on loop timing. Fixable with explicit edge detection, but that's more machinery for a case the simpler contract doesn't have at all.
- **It restores consistency with the existing `off()` philosophy.** The docs have always said `off()` only keeps the robot pushable _in conjunction with_ `enabled = false`, because any line of code can re-activate the motors. The hold loop is just another such line; the suspension gives `off()` a durability it has nowhere else in Lizard.
- **It's the safer default for an interlock.** A locked robot on a slope should be held; today `off()` leaves it coasting while the interlock condition is still unmet, which is exactly the state the feature exists to prevent.

The behavior change: `off()` while locked no longer sticks — the hold re-engages within ~1 s (`ODriveMotor::speed` re-asserts closed-loop control, so the motors come back). Pushing by hand then requires `disable()` first. That's the "docs option" from the original finding, so it needs the one sentence we skipped back then, e.g. "While `locked` is `true`, `off()` does not stick — the hold re-engages within about a second; call `disable()` to switch the motors off durably." In exchange the docs lose the two hardest-to-explain sentences we currently have ("`off()` also suspends the hold until…").

## Two housekeeping notes

One thing was easier to fix than to describe: the gate-property forwarding — writes to `locked` and `enabled` forward to shadows, while all other property writes don't — wasn't predictable from module_reference.md, so I added the missing sentence to the three `locked` paragraphs directly on the branch.

And a nit for the record: we noticed main/modules/wheels.h is the only file in the codebase using Doxygen-style comments (`///` and a `/** */` block); everything else uses plain `//`. We decided to leave it as is — the triple slash explicitly ties each comment to the declaration below it, and if we ever adopt Doxygen or similar tooling, the file is already in the right shape. Just noting it so the deviation is on record as deliberate.

JensOgorek:
Done — the suspension is gone (17d7bc1). The contract is now the one sentence: **while `enabled && locked`, the wheels hold; `disable()` switches them off.** Your docs sentence went in verbatim, and the slope argument is what sealed it for us: an interlock that stops holding because someone called `off()` fails exactly when it matters.

Re-verified on the bench ESP32: **13/13 checks green**, including the new key one — after `off()` while locked, the hold re-engages within ~1 s instead of standing down.

<details>
<summary>The diff (17d7bc1, −20/+7)</summary>

- `wheels.cpp`/`wheels.h`: `hold_suspended` and `suspend_hold()` removed; the engage branch in `step()` is a plain `else` now; `enable()` no longer touches the flag; class comment updated.
- `odrive_wheels.cpp`, `roboclaw_wheels.cpp`: the `suspend_hold()` calls removed from both `off()` handlers.
- `module_reference.md`: the two suspension sentences replaced by the suggested precedence sentence in the ODrive and RoboClaw sections; the Dunker section (no `off()`) and your shadow-forwarding sentence are untouched.
- PR body updated to match (contract paragraph, implementation bullet, hardware-test note).

</details>

<details>
<summary>Bench verification (13 checks, no motors attached)</summary>

Observability via the TWAI TX queue: `ODriveMotor::speed()` sends its velocity frame unconditionally (`set_mode` is suppressed without an ODrive heartbeat), and without a transceiver the frames stay queued and are countable as `can.msgs_to_tx`. Every hold or drive command is exactly 2 frames; `can.reset()` zeroes the queue between checks.

1. `locked` defaults to `false`
2. `enabled` defaults to `true`
3. `speed()` while unlocked sends 2 frames
4. hold sent once on the rising edge of `locked`
5. hold refreshed at the low rate (~1 s), not per cycle
6. 10× `speed()` while locked adds no frames beyond the ongoing refresh
7. **`off()` while locked does not stick: 4 frames within the next 1.5 s (old behavior: 0)**
8. unlocking stops the hold (no frames for 1.5 s)
9. driving resumes after unlocking
10. `locked` write forwards to an attached shadow
11. gate sync onto a freshly attached shadow (attached while the master is locked → shadow locked)
12. `disable()` stops the hold durably
13. `enabled = false` forwards to shadows

Physical holding force still needs a real motor controller on the bus and was not measured.

</details>

falkoschindler (approved):
@JensOgorek approving — the machinery held up under one more full pass: the gate, the shadow forwarding, the attach sync and the low-rate hold refresh all check out, and the one-sentence contract (_while `enabled && locked`, the wheels hold; `disable()` switches them off_) is now exactly what the code does and the docs say.

While re-checking the hold logic I found one last timing edge, and it was once again easier to fix than to describe, so it's already on the branch (d0daf47): `holding` was only cleared by a `step()` observing the unlocked or disabled state, so `wheels.disable()` followed by `wheels.enable()` within the same 10 ms cycle — e.g. from a single rule — skipped the rising-edge send, and the freshly re-enabled motors sat idle without a setpoint for up to ~1 s until the next refresh. `disable()` now resets the flag itself (main/modules/wheels.cpp (line 128)), i.e. exactly where the hold dies with the motors, so any path back to "enabled and locked" re-engages the hold immediately. The `off()` behavior is untouched and keeps its documented ~1 s re-engage. The one-liner compiles clean; if you get a chance, a quick bench check of the disable→enable-in-one-cycle case would close it out.

One non-blocking note for the release notes: direct writes to `wheels.enabled` now reach shadows (previously only the `disable()` method call did, via `call_with_shadows`), and the attach-time sync overwrites a shadow's pre-set gate values master-wins. Both are the intended feature and well documented on this PR, but existing shadow setups that relied on the old behavior will change — worth a sentence when this ships.

Nice work on this one 👏🏻

JensOgorek:
Bench check of the disable→enable-in-one-cycle edge is done — d0daf47 works, **A/B-verified against the pre-fix firmware**: on 17d7bc1 the rising edge is missed exactly as you described (hold only returns with the ~1 s refresh), with your one-liner it re-engages on the next 10 ms cycle. `off()` keeps its documented ~1 s re-engage, `disable()` alone still stops the hold durably — 10/10 checks on both firmwares (each asserting its expected behavior).

Your release-notes point is now a **"Notes for the release notes (0.13.0)"** section in the PR body (both shadow-visible changes: `enabled` property writes forwarding, master-wins attach sync), so it gets picked up when the notes for the milestone are drafted.

<details>
<summary>Measurement details (bench ESP32, no transceiver)</summary>

Same observability as before: frames counted via the TWAI TX queue (`can.msgs_to_tx`). The trigger `can.reset(); wheels.disable(); wheels.enable()` runs in a single command line, i.e. within one 10 ms cycle, sent right after a hold-refresh burst so the next scheduled refresh is ~1 s away. The frame signature makes the two behaviors cleanly distinguishable:

| after trigger | disable's `stop()` | enable's `reset_motor_error()` | hold rising edge | @150 ms | @1.3 s |
|---|---|---|---|---|---|
| **d0daf47** | 2 | 2 | 2 (next cycle) | **6** | 8 |
| **pre-fix (17d7bc1)** | 2 | 2 | — (misses, waits for refresh) | **4** | 6 |

Measured 3× per firmware, byte-identical results each run. One measurement subtlety worth recording: `msgs_to_tx` is refreshed in `Can::step()`, so the first ~50 ms of readings after `can.reset()` can still show the pre-reset value — the checks therefore sample at fixed offsets (150 ms / 1.3 s) instead of polling for a threshold.

Full check list per firmware: `locked`/`enabled` defaults, hold on rising edge of `locked`, ~1 s refresh, 3× the disable+enable-in-one-cycle case (asserting 6/8 on the fix, 4/6 pre-fix), `off()` re-engaging via refresh only (0 frames at 150 ms, 2 at 1.3 s), `disable()` stopping the hold durably.

</details>

PR #229, opened 2026-07-04
espresso: add --baud and --no-stub flags for flaky USB-UART bridges

### LINKED ISSUE #228: espresso.py: hardcoded flash baud + stub upload block some USB-UART bridges — expose --baud / --no-stub
### Problem
`espresso.py flash` hardcodes `--baud 921600` and relies on esptool's default RAM stub, with no way to override either. On some USB-UART bridges the stub upload corrupts and flashing fails outright — and because espresso exposes no knob, the only workaround is to bypass it and call `esptool` by hand.

### Repro (real board)
- **Board:** [ESP32-S3 N16R8 Devboard (16 MB flash / 8 MB PSRAM)](https://www.otronic.nl/en/esp32-s3-n16r8-devboard-16mb-flash-en-8mb-psram.html) — dual USB-C: a **CH343P** USB-UART bridge (WCH `1a86:55d3`) and a native USB-OTG port. Host: macOS/Apple Silicon, esptool v4.11.0, ESP-IDF v5.3.1.
- Over the **CH343P UART** port, `sudo ./espresso.py flash --chip esp32s3 --device /dev/cu.usbmodem…` connects and detects the chip, then:
  ```
  Uploading stub...
  A fatal error occurred: Failed to write to target RAM (result was 0107: Checksum error)
  ```
- Lowering the rate doesn't help — a direct `esptool … --baud 115200 write_flash …` over the same bridge gives the **same** checksum error. So it's the **stub RAM upload over this bridge**, not just transfer speed.
- Flashing the identical binaries over the board's **native USB-Serial-JTAG** port succeeds cleanly (`Hash of data verified`, ~9 s), and lizard then boots and runs.

### Why espresso can't currently cope
- Baud is hardcoded in `flash()` (`--baud 921600`) — no `--baud` flag.
- No `--no-stub` option — the failing stub upload can't be skipped.
- Net: a board whose bridge can't take the fast stub upload can't be flashed *via espresso* at all; you must drop to raw esptool.

### Proposed
1. **Add `--baud`** (default 921600) — smallest fix; lets users slow flaky bridges.
2. **Add `--no-stub`** passthrough — bypasses the exact step that fails here.
3. Optional: document/prefer the **native USB-Serial-JTAG** port for ESP32-S3 as the robust path.

*Per-board/per-host USB-UART flakiness is a known variable, so this isn't "make 921600 universal" — it's "give espresso the escape hatches so it isn't a dead end on such boards." Surfaced while flashing an S3 during #224 (espresso consolidation); related to #217.*

falkoschindler:
Fixes #228.

`espresso.py flash` hardcoded `--baud 921600` and always used esptool's RAM stub. On some USB-UART bridges (e.g. the CH343P in #228) the stub upload fails with a checksum error regardless of baud rate, so such boards could not be flashed via espresso at all.

This PR adds two escape hatches, both passed through to esptool for `flash` and `erase`:

- `--baud` (default: 921600) to slow down flaky bridges,
- `--no-stub` to skip the RAM stub upload and flash via the ROM loader instead (slower, but avoids the failing step entirely).

Default behavior is unchanged — without the new flags the esptool invocation is byte-identical to before (verified via `--dry-run`). `reset_partition` keeps its conservative hardcoded 115200.

The docs now also recommend the native USB-Serial-JTAG port (e.g. on ESP32-S3) as the most robust flashing path; since #221, espresso falls back to esptool's own reset on non-Jetson hosts, so this already works out of the box.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -13,6 +13,10 @@ sudo ./espresso.py flash <device_path>
 Note that flashing may require root access (hence the sudo).
 The command also does not work while the serial interface is busy communicating with another process.
 
+If flashing fails on your USB-UART bridge, you can lower the baud rate with `--baud` (default: 921600).
+If the stub upload itself fails (e.g. "Failed to write to target RAM"), pass `--no-stub` to flash via the ROM loader instead.
+On boards with a native USB-Serial-JTAG port (e.g. ESP32-S3), prefer that port over an external USB-UART bridge as the most robust way to flash.
+
 ### Robot Brain
 
 The `espresso.py` script can also upload firmware on a [Robot Brain](https://www.zauberzeug.com/product-robot-brain.html)
--- a/espresso.py
+++ b/espresso.py
@@ -226,7 +226,8 @@ def reset_partition() -> None:
         'esptool.py',
         '--chip', args.chip,
         '--port', DEVICE,
-        '--baud', '115200',
+        '--baud', BAUD,
+        *STUB_ARGS,
         'erase_region',
         '0xf000',  # otadata partition offset (default ESP-IDF partition table)
         '0x2000',  # otadata partition size (default ESP-IDF partition table)
--- a/espresso.py
+++ b/espresso.py
@@ -106,6 +106,10 @@ parser.add_argument('--chip', choices=['esp32', 'esp32s3'], default='esp32', hel
 parser.add_argument('--reset-partition', action='store_true', help='Reset to default OTA partition after flashing')
 parser.add_argument('-d', '--dry-run', action='store_true', help='Dry run')
 parser.add_argument('--device', nargs='?', default=DEFAULT_DEVICE, help='Serial device path (auto-detected on Jetson)')
+parser.add_argument('--baud', type=int, default=921600, help='Baud rate for flashing and erasing')
+parser.add_argument('--no-stub', action='store_true',
+                    help='Do not upload esptool\'s RAM stub (slower, but avoids stub upload failures '
+                         'on some USB-UART bridges)')
 
 args = parser.parse_args()
 ON = 1 if args.nand else 0
@@ -114,6 +118,8 @@ DRY_RUN = args.dry_run
 SWAP = args.swap
 CHIP = args.chip
 DEVICE = args.device
+BAUD = str(args.baud)
+STUB_ARGS = ('--no-stub',) if args.no_stub else ()
 FLASH_FREQ = {'esp32': '40m', 'esp32s3': '80m'}.get(CHIP, '40m')
 BOOTLOADER_OFFSET = {'esp32': '0x1000', 'esp32s3': '0x0'}.get(CHIP, '0x1000')
 BOOTLOADER = args.bootloader
@@ -203,7 +209,8 @@ def erase() -> None:
                 'esptool.py',
                 '--chip', CHIP,
                 '--port', DEVICE,
-                '--baud', '921600',
+                '--baud', BAUD,
+                *STUB_ARGS,
                 '--before', 'default_reset',
                 '--after', 'hard_reset',
                 'erase_flash',
@@ -237,7 +244,8 @@ def flash() -> None:
                 'esptool.py',
                 '--chip', CHIP,
                 '--port', DEVICE,
-                '--baud', '921600',
+                '--baud', BAUD,
+                *STUB_ARGS,
                 '--before', 'default_reset',
                 '--after', 'hard_reset',
                 'write_flash',


evnchn (commented):
**Reviewed** (I filed #228, so I flashed this on the actual board). The core change is correct and I'd merge it — one gap worth a decision first.

## TL;DR
- ✅ `--baud` / `--no-stub` passthrough in `flash()` and `erase()` is correct; "byte-identical without the flags" holds; `--no-stub` sits in the right place (global esptool option, before the subcommand).
- ⚠️ **`reset_partition()` doesn't get `--no-stub`** — so `flash --no-stub --reset-partition` still uploads the stub in the reset step and **dead-ends on the exact CH343P bridge this PR is for.** Verified on hardware below.

## The one gap
`flash()` calls `reset_partition()` at line 262–263 when `--reset-partition` is set, on the same connection right after `write_flash`. But `reset_partition()` (line 222) builds its own esptool argv **without** `STUB_ARGS`, so it always uploads the RAM stub. On a bridge whose stub upload fails (the #228 CH343P — and #228 established it fails *even at 115200*, so the conservative baud doesn't save it), the flash succeeds via `--no-stub` and then the reset step fails at stub upload — the very step `--no-stub` exists to skip.

<details><summary><b>Hardware verification — ESP32-S3 N16R8 from #228, CH343P port, esptool 5.3.0</b></summary>

Same board, MAC `3c:84:27:c8:4e:40`, two ports: native USB-Serial/JTAG (`cu.usbmodem101`, stub OK) and the CH343P bridge (`cu.usbmodem585A…`, stub fails). All tests below are on the **CH343P bridge**.

**A — what `reset_partition()` does today (stub, 115200) → fails deterministically (2/2):**
```
$ esptool --chip esp32s3 --port <CH343P> --baud 115200 erase_region 0xf000 0x2000
Uploading stub flasher...
A fatal error occurred: Failed to write to target RAM (result was 0107: Checksum error)
```

**B — the proposed fix (`--no-stub`, ROM loader) → works, reliably (3/3):**
```
$ esptool --chip esp32s3 --port <CH343P> --baud 115200 --no-stub erase_region 0xf000 0x2000
Flash memory region erased successfully in 0.1 seconds.
```

**Is `--no-stub` erase safe on the default `--chip esp32` too?** Yes, on modern esptool. esptool 5.3.0 routes ROM-mode erases through `flash_begin` for all non-ESP8266 chips, so it doesn't need the stub's erase opcode:
```python
# esptool/cmds.py erase_region()
if esp.CHIP_NAME != "ESP8266" and not esp.IS_STUB:
    _set_flash_parameters(esp, flash_size="detect"...)
    esp.flash_begin(size, address, logging=False)   # ROM-mode erase
else:
    esp.erase_region(address, size)
```
(Caveat: #228 originally ran esptool 4.11.0; I verified on 5.3.0.)
</details>

**Suggested fix** — thread the same escape hatches into `reset_partition()` (also `BAUD`, for consistency, so a `--baud` chosen for a flaky bridge applies throughout):
```python
def reset_partition() -> None:
    print_bold('Resetting partition to "ota_0"...')
    success = run(
        'esptool.py',
        '--chip', args.chip,
        '--port', DEVICE,
        '--baud', BAUD,        # was '115200'
        *STUB_ARGS,            # new
        'erase_region', '0xf000', '0x2000',
    )
```
Alternatively, if you'd rather keep `reset_partition()` conservative, a one-line note in `tools.md` that `--reset-partition` needs a bridge that can take the stub would close the doc gap.

Nice small fix. 👍

<details><summary><b>Graveyard — things I checked and deliberately did <i>not</i> flag</b> (so they don't get re-raised; shout if you disagree with any)</summary>

| Candidate | Verdict | Why |
|---|---|---|
| "byte-identical without the flags" claim | ✅ true | `str(921600) == '921600'` and `STUB_ARGS == ()` when `--no-stub` absent, so argv is unchanged. |
| `--no-stub` argv placement | ✅ correct | It's a global esptool option and sits before the `write_flash`/`erase_flash` subcommand token in both `flash()` and `erase()`. |
| `--no-stub` erase might break on the default `--chip esp32` (classic ROM lacks an erase opcode) | ❌ not an issue on modern esptool | esptool 5.3.0 routes ROM-mode erases through `flash_begin` for all non-ESP8266 chips (`cmds.py` `erase_region`), so no stub opcode is needed. Verified on S3; classic-esp32 shares the branch. Only unverified on the 4.x line #228 originally used. |
| No range check on `--baud` (`--baud 0` / `-1`) | ❌ cosmetic | argparse accepts it, esptool then errors — not worth a guard. |

</details>

falkoschindler:
Thanks @evnchn for the hardware-verified review — you're right, and I've applied your suggested fix as-is in 0d6879a: `reset_partition()` now gets `BAUD` and `*STUB_ARGS` like `flash()` and `erase()`.

On the baud part there's even a stronger argument than consistency: `reset_partition()` is not a standalone command — it only ever runs from `flash()`, immediately after a successful `write_flash` on the same port. If the multi-MB flash just succeeded at the chosen baud (and stub mode), the same parameters are proven safe for the 8 KiB `erase_region`. So the hardcoded 115200 wasn't buying any real robustness, and the default-behavior change (115200 → 921600) is risk-free in practice.

Re the esptool 4.x caveat in your graveyard: agreed it's acceptable — `--no-stub` is opt-in, and anyone reaching for it is currently hard-blocked at the stub upload, so they can't end up worse off.

evnchn (approved):
Confirmed — 0d6879a is exactly the fix, and your stronger argument is right: `reset_partition()` only runs from `flash()` right after a successful `write_flash` on the same port, so `BAUD` + stub mode were already exercised successfully by the write that just landed. The hardcoded 115200 wasn't buying robustness there. LGTM — approving. Thanks for the quick turnaround. 👍

PR #236, opened 2026-07-08
Bound string-escape loop in print_to_buffer to prevent stack overflow

### LINKED ISSUE #235: Stack buffer overflow in Expression::print_to_buffer when echoing a quote-heavy string

**TL;DR:** A single serial line can overflow a stack buffer in `Expression::print_to_buffer` and corrupt memory. When Lizard echoes a `str` value, it escapes each `"`/`\` into two chars, writing into `char escaped[buffer_len]` — a VLA sized to the **output** buffer (256 on the echo path), with **no bound check** in the loop. A quote-heavy string writes ~2× its length and runs off the end of the VLA. `csprintf`'s length guard only fires *after* the loop, so it can't save you — the stack is already smashed. Reproduced on real hardware (ESP32-S3); the coredump crashes freeing `0x225c225c` (where `0x22`=`"`, `0x5c`=`\`) — the escaped byte pattern, consistent with the overflow corrupting the string temporary's internal pointer. This is worse than a clean crash: it's an out-of-bounds stack write whose landing spot depends on the input.

## Reproduce

Store a string of ~450 quote characters (each entered as `\"`), then echo it:

```
str b = "\"\"\"…\""      # 450 escaped quotes -> a 450-char string of '"'
b                          # echo re-escapes to ~901 bytes into escaped[256]
```

The board panics and reboots. A short quote string round-trips correctly (`str s = "\"\"\""` → `s` echoes `"\"\"\""`), which is what makes the overflow purely a length problem, not a logic one.

## Root cause

`main/compilation/expression.cpp`, the `string` case of `print_to_buffer`:

```cpp
case string: {
    char escaped[buffer_len];              // sized to the OUTPUT capacity (256 on the echo path)
    int pos = 0;
    for (char c : this->evaluate_string()) {
        if (c == '"' || c == '\\') {
            escaped[pos++] = '\\';         // up to 2 bytes per input char...
        }
        escaped[pos++] = c;                // ...with NO check that pos < buffer_len
    }
    escaped[pos] = '\0';
    return csprintf(buffer, buffer_len, "\"%s\"", escaped);   // guard is here — too late
}
```

The echo path passes `buffer_len = 256` (`main.cpp:200`, `static char buffer[256]`). A stored string of N quote/backslash chars makes the loop write `2N + 1` bytes; once `2N + 1 > 256` (N ≥ 128) it writes past `escaped[]`. `csprintf` does bound-check (it `vsnprintf`s and throws `"buffer too small"`), but only when it runs *after* the loop — by then the out-of-bounds writes have already happened. The same code is reachable over the expander bus (`buffer[512]` in `expander.cpp`), so a larger string overflows there too.

## Evidence — coredump from a live crash

ESP32-S3, read back with the core-dump tooling:

<details>
<summary><b>Symbolized backtrace — freeing a pointer made of escaped bytes</b></summary>

```
Panic reason: assert failed: heap_caps_free heap_caps_base.c:69
              (heap != NULL && "free() target pointer is outside heap areas")

#3  heap_caps_free (ptr=<optimized out>)   at components/heap/heap_caps_base.c:69
#4  free (ptr=0x225c225c)                   at components/newlib/heap.c:39
#5  operator delete (ptr=0x225c225c)
#6  std::__new_allocator<char>::deallocate (__n=576463453, ...)
#8  std::string::_M_destroy (__size=576463452, ...)
#11 std::string::~basic_string (this=0x3fcaf6e0)
#12 Expression::print_to_buffer (buffer=..., buffer_len=256) at main/compilation/expression.cpp:48
```

The crashing free target is `0x225c225c`. `0x22` is `"` and `0x5c` is `\` — the byte pattern the escape loop emits. The crash is in the destructor of the `std::string` temporary that the range-`for` iterates over, freeing a pointer whose value is those escaped bytes (and `__size` is likewise garbage) — consistent with the overflow running past `escaped[]` and corrupting that temporary's metadata. Where the out-of-bounds bytes land is input-dependent; a different length could just as well overwrite a return address rather than this pointer.
</details>

## Suggested fix

Escape into a growable `std::string` instead of a fixed stack VLA, then let `csprintf` do the (already correct) length check on the final output — so an over-long value throws `"buffer too small"` and is echoed as an error rather than smashing the stack:

```cpp
case string: {
    std::string escaped;
    for (char c : this->evaluate_string()) {
        if (c == '"' || c == '\\') {
            escaped += '\\';
        }
        escaped += c;
    }
    return csprintf(buffer, buffer_len, "\"%s\"", escaped.c_str());
}
```

(Equivalently, bound the loop on `buffer_len` and truncate — but the `std::string` version keeps the existing "too long ⇒ runtime_error ⇒ echo" behavior and removes the VLA entirely.) Happy to open a PR with this plus a test.

## Environment

- Hardware: ESP32-S3 (the code path is chip-independent; also reachable over the expander bus)
- Firmware: build off the #227 branch, whose `expression.cpp` is byte-identical to `main` — so the bug is present on `main`. ESP-IDF v5.3.1
- Sibling finding: the unguarded integer divide-by-zero in the same evaluator (#234). Both are single-serial-line crashes in expression handling; this one is memory-unsafety rather than a clean trap.

evnchn:
Fixes #235.

**TL;DR:** The `string` case of `Expression::print_to_buffer` escapes `"`/`\` into a stack VLA `char escaped[buffer_len]`, writing up to 2 bytes per input char with **no bound check** — so a quote-heavy stored string writes ~2× its length and runs off the end, smashing the stack (reproduced on ESP32-S3 in #235). This adds a **single per-iteration bound check** that throws `"buffer too small"` — the *same* `runtime_error` `csprintf` already throws for an over-long result, already caught by the top-level handlers and echoed. **No heap, no `std::string`, VLA retained** — addresses the memory-unsafety without the allocation cost flagged in the issue thread.

```diff
         for (char c : this->evaluate_string()) {
+            if (pos + 3 > (int)buffer_len) // room for a 2-byte escape plus the final '\0' (#235)
+                throw std::runtime_error("buffer too small");
             if (c == '"' || c == '\\') {
                 escaped[pos++] = '\\';
             }
             escaped[pos++] = c;
         }
```

Observable behaviour for valid strings is **unchanged**; over-long values are echoed as `buffer too small` instead of crashing. `+3` covers the worst case for one iteration: up to 2 bytes for an escaped char, plus the terminating `'\0'` written after the loop. As a bonus it also closes the 1-byte overrun a *plain* string of length `buffer_len` would cause on that terminating null — not just the quote-doubling case.

<details>
<summary><b>Why the guard is placed on every char, not just the escape branch</b></summary>

Tempting to only check inside the `if (c == '"' || c == '\\')` branch (the escape chars are what write 2 bytes). But a **plain** string overflows too: it writes 1 byte per char, then `escaped[pos] = '\0'` lands after them, so a plain string of length `buffer_len` overruns by one. Guarding only the escape branch leaves the plain write unchecked — confirmed with AddressSanitizer below. Checking every char is the only single-line placement that covers both.
</details>

<details>
<summary><b>Why this can't reject a valid string (boundary reasoning)</b></summary>

The guard bounds `escaped` to the VLA — `pos` never exceeds `buffer_len-1`, so `escaped[pos]='\0'` is always in-bounds. It does **not** bound the final output; `csprintf` still owns that (`csprintf` accepts iff `strlen(escaped) <= buffer_len-3`, for the two `"` quotes + null). The guard is deliberately *looser* than `csprintf`, so any string that used to echo successfully (which had `strlen(escaped) <= buffer_len-3`, well below the trip point) still echoes; only the out-of-bounds write is removed.
</details>

<details>
<summary><b>Empirical verification — AddressSanitizer, host build (<code>-O0 -fsanitize=address</code>)</b></summary>

Extracted the exact `csprintf` + escape loop into a host harness and ran under ASan (`-O0` so the UB isn't optimized away):

```
OLD (unpatched):
  ==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow ... WRITE of size 1
  SUMMARY: dynamic-stack-buffer-overflow in print_old(...)      # the escaped[pos++]=c write

NEW (this PR):
  [control] 3 quotes -> "\"\"\"" (len 8) OK                     # round-trips identically
  [overflow] 450 quotes -> threw "buffer too small"             # clean, no OOB
  reached end of main cleanly

guard-in-escape-branch-only (rejected alternative):
  ==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow      # plain 256-char string, on the '\0'
```

On real ESP32-S3 the unpatched path crashed freeing `0x225c225c` — the escaped `"\` byte pattern — matching the smashed-stack signature in #235.
</details>

<details>
<summary><b>Scope / known non-goal</b></summary>

A pre-existing, **unreachable** edge remains untouched: `print_to_buffer(buf, 0)` with an *empty* string would write `escaped[0]` on a zero-length VLA. All call sites pass either a fixed size (256/512) or `sizeof(buffer) - pos`, and `csprintf` guarantees each chained write leaves `pos <= size-1` (or throws), so `buffer_len >= 1` always — the `0` case isn't reachable, and it predates this change. Left out to keep the diff to the one line that fixes the reported bug; happy to fold in a `buffer_len < 3` pre-check if you'd prefer defence-in-depth.

No host-side test harness exists in the repo (CI is pre-commit only), so no unit test is added here — the ASan repro above stands as the evidence. Glad to add a test in whatever shape you'd prefer.
</details>

### DIFF
--- a/main/compilation/expression.cpp
+++ b/main/compilation/expression.cpp
@@ -41,6 +41,8 @@ int Expression::print_to_buffer(char *buffer, size_t buffer_len) const {
         char escaped[buffer_len];
         int pos = 0;
         for (char c : this->evaluate_string()) {
+            if (pos + 3 > (int)buffer_len) // room for a 2-byte escape plus the final '\0' (#235)
+                throw std::runtime_error("buffer too small");
             if (c == '"' || c == '\\') {
                 escaped[pos++] = '\\';
             }


falkoschindler (approved):
Thanks @evnchn, great catch and a very well-argued fix! I verified the boundary reasoning against the code: the guard keeps all writes (including the trailing NUL) inside the VLA, and since csprintf in main/utils/string_utils.cpp (26) already required strlen(escaped) <= buffer_len - 3, it can't reject any string that previously echoed successfully. Nice bonus that it also closes the 1-byte NUL overrun for plain strings. Keeping the diff minimal and heap-free is exactly right here.

PR #237, opened 2026-07-08
Reject expander proxies as module arguments (#233)

### LINKED ISSUE #233: get_module_argument silently casts expander Proxy modules, crashing on the first virtual call
## Problem

`get_module_argument<M>` in `main/modules/module_helpers.h` falls back to a `static_pointer_cast<M>` when the named module is a `Proxy`:

```cpp
if (auto typed = std::dynamic_pointer_cast<M>(module)) {
    return typed;
}
if (std::dynamic_pointer_cast<Proxy>(module)) {
    return std::static_pointer_cast<M>(module);
}
```

This is what makes expander-hosted modules usable as arguments at all, but it is unconditional: it also "succeeds" for modules that cannot work as a proxy. Example: declare two `ODriveMotor`s on an expander and pass them to a core-hosted `ODriveWheels`. Construction succeeds, then the first `step()` calls the virtual `Motor::get_position()` on an object that is actually a `Proxy` — a vtable lookup on the wrong class layout, i.e. LoadProhibited crash or garbage odometry.

Before the self-registering refactor (#211), this configuration was cleanly rejected at parse time ("module ... is no ODrive motor"). Since #211 every module that takes typed module arguments shares the hazard.

## Suggestion

`get_module_argument` should only hand out a `Proxy` when the call site can actually use one, i.e. when it only interacts with the module through `Module`'s virtual interface (`call`, `write_property`), which `Proxy` forwards. Call sites that invoke C++ methods on the concrete type (motors passed to wheels modules, `Can` passed to CAN devices, ...) need a way to opt out — e.g. a `get_local_module_argument<M>` that rejects proxies with a proper error message, or an explicit flag on `get_module_argument`.

evnchn:
*Drafted by Claude Code on evnchn's behalf; updated by @falkoschindler after the review discussion.*

Closes #233.

`get_module_argument<M>` had a `Proxy` fallback that returned `static_pointer_cast<M>(proxy)` whenever the argument was a proxy for a module on a port expander. A `Proxy` is not laid out as `M`, so this is undefined behavior: the config is accepted at parse time, and the first virtual call on the concrete type (e.g. `ODriveWheels` reading `get_position()` off a proxied `ODriveMotor`) crashes on the running robot.

The helper is now strict: `dynamic_pointer_cast<M>` or a parse-time error — **module arguments must live on the same microcontroller, with no exception**. In particular, end stops on a different microcontroller than the motor control are deliberately no longer supported: reading a safety-relevant limit switch through the expander's broadcast channel adds latency to every safety check, and before the expander is up a proxied end stop silently reports `active = false`. The restriction is documented in the Proxy section of the module reference.

**Breaking change:** configs that pass a proxy as a module argument (in practice: MotorAxis end stops on a port expander with the motor on the core) are now rejected at parse time with `module "..." is no Input`. The few affected robots keep working on Lizard ≤ 0.13; this PR is scheduled for **0.14.0** (see discussion below).

Verified with a full esp32 firmware build (espressif/idf:v5.3.1); an earlier iteration was also verified with a sanitizer-instrumented host MRE of the #233 crash config (see comments).

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -1113,6 +1113,9 @@ Note that the remote module has to have turned on broadcasting: `x.broadcast()`.
 
 Note that the proxy module forwards all method calls to the remote module.
 
+Proxies cannot be passed as arguments to other module constructors (e.g. as end stops for a motor axis), because the actual module only exists on the remote microcontroller.
+Declare the depending module on the same microcontroller instead.
+
 | Properties | Description                                       | Data type |
 | ---------- | ------------------------------------------------- | --------- |
 | `is_ready` | Whether the remote module has booted and is ready | `bool`    |
--- a/main/modules/module.h
+++ b/main/modules/module.h
@@ -37,7 +37,6 @@ protected:
     bool broadcast = false;
 
 public:
-    static inline constexpr const char *TYPE = "Module";
     static bool broadcast_paused;
     const std::string name;
 
--- a/main/modules/module.h
+++ b/main/modules/module.h
@@ -37,6 +37,7 @@ protected:
     bool broadcast = false;
 
 public:
+    static inline constexpr const char *TYPE = "Module";
     static bool broadcast_paused;
     const std::string name;
 
--- a/main/modules/module_helpers.h
+++ b/main/modules/module_helpers.h
@@ -28,6 +28,8 @@ inline std::shared_ptr<Held> get_module_argument(const ConstExpression_ptr &arg)
     // A proxy is not an M in the C++ type system, so casting it to M is UB (#233).
     // But when the caller holds it only as a base (Held, e.g. Module), a proxy standing
     // for a remote M is safe: validate it by the remote type name, return it as that base.
+    // Name equality cannot model inheritance: M must be a registered type name like
+    // "Input", not a base of other registered types like "Motor".
     if constexpr (!std::is_same_v<Held, M>) {
         if (const auto proxy = std::dynamic_pointer_cast<Proxy>(module)) {
             if (proxy->module_type == M::TYPE) {
--- a/main/modules/module_helpers.h
+++ b/main/modules/module_helpers.h
@@ -4,9 +4,11 @@
 #include "../global.h"
 #include "driver/gpio.h"
 #include "module.h"
+#include "proxy.h"
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <type_traits>
 
 #ifdef CONFIG_IDF_TARGET_ESP32S3
 #define DEFAULT_I2C_SDA_PIN GPIO_NUM_8
@@ -16,14 +18,22 @@
 #define DEFAULT_I2C_SCL_PIN GPIO_NUM_22
 #endif
 
-template <typename M>
-inline std::shared_ptr<M> get_module_argument(const ConstExpression_ptr &arg) {
+template <typename M, typename Held = M>
+inline std::shared_ptr<Held> get_module_argument(const ConstExpression_ptr &arg) {
     const std::string name = arg->evaluate_identifier();
     const Module_ptr module = Global::get_module(name);
-    if (auto typed = std::dynamic_pointer_cast<M>(module)) {
+    if (const auto typed = std::dynamic_pointer_cast<M>(module)) {
         return typed;
     }
-    // No Proxy fallback: static_pointer_cast<M> of a proxy is UB (#233). A proxy is
-    // usable only as a base Module, so use get_module_argument<Module> for those.
+    // A proxy is not an M in the C++ type system, so casting it to M is UB (#233).
+    // But when the caller holds it only as a base (Held, e.g. Module), a proxy standing
+    // for a remote M is safe: validate it by the remote type name, return it as that base.
+    if constexpr (!std::is_same_v<Held, M>) {
+        if (const auto proxy = std::dynamic_pointer_cast<Proxy>(module)) {
+            if (proxy->module_type == M::TYPE) {
+                return proxy;
+            }
+        }
+    }
     throw std::runtime_error("module \"" + name + "\" is no " + M::TYPE);
 }
--- a/main/modules/module_helpers.h
+++ b/main/modules/module_helpers.h
@@ -4,11 +4,9 @@
 #include "../global.h"
 #include "driver/gpio.h"
 #include "module.h"
-#include "proxy.h"
 #include <memory>
 #include <stdexcept>
 #include <string>
-#include <type_traits>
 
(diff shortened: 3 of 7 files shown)


evnchn:

Unrelated to this PR, but two tooling things I hit while build-verifying locally, in case they're useful: `compile.sh` runs bare `idf.py build` without `SDKCONFIG_DEFAULTS`, so it builds with C++ exceptions off and fails on every `throw` (the working path is `build.py`); and CI (`check.yml`) is `pre-commit`-only, so nothing actually compiles the firmware on a PR. Happy to file and/or fix either separately if you'd like — just flagging, not asking anything of this PR.

falkoschindler (changes requested):
Thanks @evnchn, this looks great — the strict helper is the right call, and the verification (sanitizer MRE with the exact #233 repro, call-site audit, full ESP32 build) is much appreciated.

One request before merging: please do fold in the follow-up you offered, storing the `module_type` that `Proxy` already receives in its constructor and letting `get_module_argument` validate proxies by type name. Without it, `motor_axis` accepts *any* module as an end-stop input, and a wrong module only fails at runtime: `can_move` throws `unknown property "active"`, which in `step()` (main/modules/motor_axis.cpp (56–58)) propagates before `motor->stop()` is reached and is then merely echoed every cycle by `run_step`. Since the motor can also be commanded directly, a misconfigured axis ends up running without end-stop protection instead of being rejected at parse time. For a safety-oriented module I'd rather keep that validation in this PR than defer it — and it would also let MotorAxis keep requesting `Input` semantically.

Happy to merge once that's in.

evnchn:
*Drafted by Claude Code on evnchn's behalf.*

**TL;DR:**

**Done — folded the validation in as 47b61d3.**

`Proxy` now stores the `module_type` it already received, and `get_module_argument` validates proxies by that remote type name — so a mis-wired `motor_axis` end stop is rejected at parse time again (your safety point), while a proxied end switch still works and the #233 UB stays gone.

One design note worth your eyes: I could not make `get_module_argument<Input>` literally return an `Input_ptr` for a proxy — that return type is exactly what forces the `static_pointer_cast<Input>(proxy)` this PR removes. So MotorAxis requests `get_module_argument<Input, Module>`: validate as Input, hold as base Module. Details below — happy to reshape if you'd prefer.

<details>
<summary><b>The design — how a proxy is validated without being cast</b></summary>

`get_module_argument<M, Held = M>` — the second param is the type the caller *holds* the result as.

- **Strict callers stay strict, zero behavior change.** For `<Motor>`, `<Can>`, `<const Serial>`, … `Held == M`, so `if constexpr (!std::is_same_v<Held, M>)` compiles the proxy branch out entirely and a proxy is rejected exactly as in the current PR. All 24 concrete call sites are untouched.
- **MotorAxis calls `get_module_argument<Input, Module>(...)`:**
  - a **local** module must `dynamic_pointer_cast<Input>` — a real type check, so a non-Input local is now rejected at parse time;
  - a **Proxy** is accepted iff `proxy->module_type == Input::TYPE`, and returned only as `shared_ptr<Module>` (a genuine base of `Proxy`). It is **never** cast to `Input`, so no UB. MotorAxis keeps holding `Module_ptr` and only ever calls the non-virtual `Module::get_property("active")`.

"Request Input semantically, hold as base" is carried by the two type params; the proxy is only ever upcast `Proxy → Module`.
</details>

<details>
<summary><b>Empirical verification — host MRE under ASan/UBSan, 6/6</b></summary>

Faithful port of the `Module ← Proxy` / `Module ← Input ← GpioInput` / `Motor ← LocalMotor` hierarchy with the post-follow-up helper body, built `-O0 -ftrivial-auto-var-init=pattern -fsanitize=address,undefined` (so a would-be UB read isn't optimized into a false green):

```
A strict <LocalMotor> on local motor -> get_position()==42              [OK]
B strict <LocalMotor> on proxy (#233 crash config)                      [rejected: module "p_input" is no LocalMotor]
C <Input,Module> on local Input -> reads get_property(active)           [OK]
D <Input,Module> on proxy(Input) -> accepted, reads get_property(active) [OK]
E <Input,Module> on local Can (wrong module) -> parse-time reject        [rejected: module "m_can" is no Input]
F <Input,Module> on proxy(Can) (wrong remote type) -> parse-time reject  [rejected: module "p_can" is no Input]
ALL CHECKS PASSED (failures=0), exit=0
```

E and F are the new safety cases: a wrong end stop — local, or a proxy of the wrong remote type — is rejected at parse time instead of surfacing later as `unknown property "active"` on a motor that can still be commanded directly. B confirms the original #233 crash config is still rejected. Clean under UBSan — no reintroduced UB.
</details>

<details>
<summary><b>Target build + independent second-model review</b></summary>

- **ESP32 firmware:** `python3 build.py esp32` in `espressif/idf:v5.3.1` → `Build completed successfully for esp32`, `build/lizard.bin` 1,290,704 bytes. `module.cpp`, `motor_axis.cpp`, `proxy.cpp`, `odrive_wheels.cpp` all compiled.
- **Adversarial second-model pass** (Codex, "assume it's broken"): verdict **SHIP**, no blockers. It independently confirmed `Input` is the *only* registered module type — `GpioInput`/`McpInput` are not separately registered — so `expander.Input(...)` proxies carry `"Input"` and the exact-string match cannot false-reject a working config. Its one nit was an optional `static_assert` making the `Held`-is-a-base contract explicit; I left it out to keep the diff minimal.
- **CI gate** (`check.yml` = pre-commit: clang-format / whitespace / EOF): passes locally; watching the run on the pushed commit.
</details>

Diff is +24/−10 across `module_helpers.h`, `motor_axis.cpp`, `proxy.h`, `proxy.cpp`. Open to the `static_assert`, a clearer second-param name than `Held`, or anything else you'd want before merge — just say the word.

falkoschindler:
We just discussed the special treatment of motor axes: Only four of our robots use proxies for end-stop inputs, and they run on Lizard 0.7.3 and 0.10.0. This is not only extra code, but also a safety and performance issue: And end stop should take effect immediately and avoid additional latency caused by our network of port expanders.

Therefore we decided to postpone this PR to 0.14.0 before we remove the special case for motor axes. This gives such old robots one more Lizard version to work with before discontinuing this feature.

I'll update the PR accordingly.

falkoschindler (approved):
As announced above, I've now removed the special treatment for motor axes again: the `Held` mechanism in `get_module_argument`, the `module_type` storage in `Proxy` and the `Module_ptr` end stops in `MotorAxis` are reverted, and the restriction is documented in the Proxy section of the module reference. The PR is back to the minimal strict helper from your first commit, @evnchn — sorry for the detour with the follow-up I asked for; the investigation it triggered is what surfaced the latency and availability concerns in the first place. Thanks for the clean work on both iterations.

PR #238, opened 2026-07-14
Guard integer division/modulo by zero in expressions

### LINKED ISSUE #234: Dividing two integers by zero (e.g. `1 / 0`) panics and reboots the ESP32

**TL;DR:** A single serial line that divides two integers by zero — `1 / 0` — panics the ESP32 with an unhandled `IntegerDivideByZero` hardware trap and **reboots the whole controller.** `DivideExpression::evaluate_integer` (and the modulo / floor-divide siblings) do a raw `int64` `/` with no divisor check, so the trap fires *below* the C++ level and sails past the `try/catch (std::runtime_error)` that turns normal expression errors into a graceful `error: …` echo. On a robot this is an unexpected reboot from one typo in a script or a bad expander value — safety-relevant. Reproduced on real hardware (ESP32-S3); coredump + fix below.

## Reproduce

Connect to a running Lizard over serial and send:

```
1 / 0
```

The board panics and reboots. Any two-integer division/modulo by zero does it (`5 % 0`, `x // 0` with integer `x`, …). Note the *float* path is unaffected — `5.0 / 0` returns `inf`, because only the integer path hits the CPU divide instruction.

## Root cause

`main/compilation/expressions.cpp` — the integer operators divide with no zero guard:

```cpp
int64_t DivideExpression::evaluate_integer() const {
    return this->left->evaluate_integer() / this->right->evaluate_integer();   // line 197
}
```

`ModuloExpression::evaluate_integer` (line 208) and `FloorDivideExpression::evaluate_integer` (line 220) have the same shape. `get_common_number_type` returns the integer type for two integer operands, so `1 / 0` takes this path and the Xtensa integer-divide instruction traps.

Why it isn't caught: Lizard already funnels expression errors through `std::runtime_error` — e.g. `expressions.cpp:24` throws `"invalid type for arithmetic operation"`, and `main.cpp:462-465` wraps `process_uart()` in `try { … } catch (const std::runtime_error &e) { echo("error processing uart0: %s", e.what()); }`. A hardware divide-by-zero trap is **not** a C++ exception, so it never reaches that handler — it goes straight to the panic handler and reboots.

## Evidence — coredump from a live crash

Captured on an ESP32-S3 (16 MB), read back with the core-dump tooling (the expression code in the build is unchanged vs `main`):

<details>
<summary><b>Panic + symbolized backtrace</b></summary>

```
Guru Meditation Error: Core 0 panic'ed (IntegerDivideByZero). Exception was unhandled.
I esp_core_dump_flash: Save core dump to flash...
I esp_core_dump_flash: Core dump has been saved to flash.

Crashed task handle: 0x3fcb0028, name: 'main'
exccause        0x6 (IntegerDivideByZeroCause)

#1  0x4200edc0 in DivideExpression::evaluate_integer (this=0x3fcb57bc) at main/compilation/expressions.cpp:197
#2  0x4200ec10 in Expression::print_to_buffer (...)               at main/compilation/expression.cpp:37
#3  0x42018971 in process_tree (...)                              at main/main.cpp:201
#4  0x4201a335 in process_lizard (...)
#5  0x4201a4b0 in process_line (...)                              at main/main.cpp:380
#6  0x4201a54a in process_uart ()                                 at main/main.cpp:398
#7  0x4201a7c8 in app_main ()                                     at main/main.cpp:463
```

The trap fires in `evaluate_integer` (frame #1), unwinds through `process_uart` (frame #6) — the exact function whose `catch (std::runtime_error)` would otherwise have echoed the error — and reboots instead of echoing.
</details>

## Suggested fix

Guard the divisor and throw `std::runtime_error`, matching how this file already reports arithmetic errors, so it's caught and echoed instead of trapping:

```cpp
int64_t DivideExpression::evaluate_integer() const {
    const int64_t divisor = this->right->evaluate_integer();
    if (divisor == 0) {
        throw std::runtime_error("division by zero");
    }
    return this->left->evaluate_integer() / divisor;
}
```

…and the analogous guard in `ModuloExpression` and `FloorDivideExpression`. With this, `1 / 0` echoes `error processing uart0: division by zero` and the controller keeps running. Happy to open a PR with the guards + a test if that'd be welcome.

## Environment

- Hardware: ESP32-S3 (also applies to ESP32 classic — the code path is chip-independent)
- Firmware: build off the #227 branch, whose `expressions.cpp` is byte-identical to `main` — so the bug is present on `main`. ESP-IDF v5.3.1
- The coredump above was read using the tooling from #227; the bug itself is independent of that PR and present on `main`.

falkoschindler:
Fixes #234.

Sending a single line like `1 / 0` over serial panicked the ESP32 with an unhandled `IntegerDivideByZero` hardware trap and rebooted the whole controller. `DivideExpression::evaluate_integer`, `ModuloExpression::evaluate_integer`, and `FloorDivideExpression::evaluate_integer` performed a raw integer division with no divisor check, so the trap fired below the C++ level and bypassed the `try/catch (std::runtime_error)` in `process_uart()` that normally turns expression errors into a graceful `error: ...` echo.

This PR guards the divisor in all three integer operators and throws `std::runtime_error` instead, matching how the file already reports arithmetic errors (e.g. "invalid type for arithmetic operation"). With this change, `1 / 0` echoes `error processing uart0: division by zero` and the controller keeps running. The float path is unaffected (it already yields `inf`/`nan` without trapping).

### DIFF
--- a/main/compilation/expressions.cpp
+++ b/main/compilation/expressions.cpp
@@ -194,7 +194,11 @@ DivideExpression::DivideExpression(const ConstExpression_ptr left, const ConstEx
 }
 
 int64_t DivideExpression::evaluate_integer() const {
-    return this->left->evaluate_integer() / this->right->evaluate_integer();
+    const int64_t divisor = this->right->evaluate_integer();
+    if (divisor == 0) {
+        throw std::runtime_error("division by zero");
+    }
+    return this->left->evaluate_integer() / divisor;
 }
 
 double DivideExpression::evaluate_number() const {
@@ -206,7 +210,11 @@ ModuloExpression::ModuloExpression(const ConstExpression_ptr left, const ConstEx
 }
 
 int64_t ModuloExpression::evaluate_integer() const {
-    return this->left->evaluate_integer() % this->right->evaluate_integer();
+    const int64_t divisor = this->right->evaluate_integer();
+    if (divisor == 0) {
+        throw std::runtime_error("modulo by zero");
+    }
+    return this->left->evaluate_integer() % divisor;
 }
 
 double ModuloExpression::evaluate_number() const {
@@ -218,7 +226,11 @@ FloorDivideExpression::FloorDivideExpression(const ConstExpression_ptr left, con
 }
 
 int64_t FloorDivideExpression::evaluate_integer() const {
-    return this->left->evaluate_integer() / this->right->evaluate_integer();
+    const int64_t divisor = this->right->evaluate_integer();
+    if (divisor == 0) {
+        throw std::runtime_error("division by zero");
+    }
+    return this->left->evaluate_integer() / divisor;
 }
 
 double FloorDivideExpression::evaluate_number() const {


JensOgorek (approved):
Reviewed and **verified on real hardware** — the fix works: all three integer operators now report a graceful error instead of panicking, and the controller keeps running. LGTM. :shipit:

One tiny nit, non-blocking: the commit title ends in "(#234)", which mimics GitHub's squash-merge suffix — after squash-merging, main history will read "… (#234) (#238)". Consider dropping it from the title; "Fixes #234" in the body already links the issue.

<details>
<summary>Review details</summary>

Review by Claude Code (Fable 5).

- **Exception is caught on every runtime path, not just serial input.** All evaluation sites in `main.cpp` are wrapped in `catch (const std::runtime_error &)`: `process_uart()`, module `step()` via `run_step()`, rule conditions, and routine steps. A division by zero inside a rule or routine echoes an error and the loop continues.
- **Float path correctly left alone** — `evaluate_number()` yields `inf`/`nan` without trapping.
- **Distinct messages** ("division by zero" vs. "modulo by zero") help debugging scripts over serial.
- The guard evaluates `right` before `left`, changing evaluation order — not observable, since expression evaluation is side-effect-free.
- `INT64_MIN / -1` remains UB, but on Xtensa 64-bit division goes through libgcc soft division (negate + unsigned divide), which wraps rather than traps — can't cause the panic this PR fixes. Not worth guarding.
- Pre-existing, out of scope: `FloorDivideExpression::evaluate_integer()` uses C++ truncating division, so `-7 // 2` yields `-3`, not `-4`. If `//` is meant to have Python floor semantics, that's a separate issue.

</details>

<details>
<summary>Hardware verification</summary>

Built the PR branch (target `esp32`, IDF v5.3.1) and flashed an ESP32-D0WD-V3 via `espresso.py flash`, then sent test lines over UART0 at 115200 baud:

| Input | Response |
|---|---|
| `1 / 0` | `error processing uart0: division by zero` |
| `1 % 0` | `error processing uart0: modulo by zero` |
| `1 // 0` | `error processing uart0: division by zero` |
| `1.0 / 0` | `inf` |
| `17 / 5` | `3` |
| `core.millis` | `4726` |

`core.millis` was still at ~4.7 s after all three error cases, i.e. **no reboot occurred** — before this fix, the first `1 / 0` already triggered an `IntegerDivideByZero` panic and restarted the controller.

</details>

PR #239, opened 2026-07-15
Handle owl allocation failure instead of aborting

JensOgorek:
### Motivation

`process_lizard()` handles every `owl_tree_get_error` code except `ERROR_ALLOCATION_FAILURE`, which falls through to the `default` branch and calls `process_tree()`. owl's accessors call `exit(-1)` on a failed tree — `exit` is not implemented on ESP-IDF, so the firmware aborts and the whole brain reboots.

This showed up as a hard boot loop (8 reboots in 25 s) on a robot with an I/O expander: garbled expander lines arriving while the heap was tight produced trees with `ERROR_ALLOCATION_FAILURE`, and every such line took the brain down. Backtrace: `Expander::handle_messages → process_lizard → process_tree → owl_tree_get_parsed_statements → check_for_error → exit → abort`.

### Implementation

Handle the error code explicitly like all other parse errors: echo `error: allocation failure while parsing` and drop the line, so a failed allocation can never reach owl's aborting accessors.

### Progress

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -339,6 +339,10 @@ void process_lizard(const char *line, bool trigger_keep_alive, bool from_expande
     case ERROR_MORE_INPUT_NEEDED:
         echo("error: more input needed at range %zu %zu", range.start, range.end);
         break;
+    case ERROR_ALLOCATION_FAILURE:
+        // Must not reach process_tree: owl's accessors exit() on a failed tree, which aborts on ESP-IDF.
+        echo("error: allocation failure while parsing");
+        break;
     default:
         if (debug) {
             owl_tree_print(tree.get());
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -320,6 +320,10 @@ void process_lizard(const char *line, bool trigger_keep_alive, bool from_expande
     if (debug) {
         toc("Tree creation");
     }
+    if (!tree) {
+        echo("error: allocation failure while parsing");
+        return;
+    }
     struct source_range range;
     switch (owl_tree_get_error(tree.get(), &range)) {
     case ERROR_INVALID_FILE:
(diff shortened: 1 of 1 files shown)


JensOgorek:
Verified on hardware (ESP32 with NAND gates and an I/O expander): without the fix the brain was in a hard boot loop (8 reboots in 25 s), with the fix it boots stably and every failed parse is just an echoed error line.

<details>
<summary>Test evidence and root-cause trail</summary>

- Firmware built from this branch with a new motor driver setup, flashed via `espresso.py flash --nand` from the Jetson.
- Crash decoded with addr2line: `Expander::handle_messages → process_lizard → process_tree → owl_tree_get_parsed_statements → check_for_error → exit → abort` (`parser.h:1308`).
- Trigger: garbled expander lines during the expander's boot window while free heap was extremely low (~2 KB, caused by stale NVS content on that device; a full flash erase brought it back to ~150 KB virgin / ~32 KB with the full startup script).
- With the fix and the tight heap, the firmware kept running and reported `error: allocation failure while parsing` instead of rebooting — which is exactly what made the real problem diagnosable.

</details>

evnchn:
image

want to help look for other `exit()` failures but can't

falkoschindler (approved):
Nice find, @JensOgorek! Claude traced the failure path through the generated parser and can confirm the diagnosis: `check_for_error` in main/parser.h (1282-1309) calls `exit(-1)` for any non-`ERROR_NONE` tree and is invoked by every `owl_tree_get_parsed_*` accessor that `process_tree` uses. The tokenizer sets `ERROR_ALLOCATION_FAILURE` and still returns a valid tree object (main/parser.h (5181-5184)), so handling the error code explicitly is exactly right — the line is dropped, the `unique_ptr` deleter still frees the tree, and the ESP stays up.

I pushed 5bca351 with two small hardenings in the same spirit:

- A NULL guard after `owl_tree_create_from_string`: under full heap exhaustion the tree allocation itself is a bare `calloc` and can return NULL, which would have crashed in `owl_tree_get_error`.
- The switch is inverted so that only `ERROR_NONE` reaches `process_tree` and unknown error codes fall into an error echo — if a future owl regeneration adds a new error code, it fails safe instead of hitting the aborting accessors.

Worth knowing (out of scope here): owl has further unconditional `abort()` calls on OOM inside tree construction itself, e.g. `construct_node_alloc` (main/parser.h (2227-2230)), which we cannot catch from main.cpp. So an extremely tight heap can still reboot the ESP, but this PR eliminates the reported path.

Approved. 👍🏻

PR #241, opened 2026-07-21
Guard parsing against low-heap crashes in owl's parser

### LINKED ISSUE #240: Parser can still abort/crash under low heap on inputs #239 doesn't cover
`process_lizard()` builds the tree with `owl_tree_create_from_string()` and only checks `owl_tree_get_error()` afterwards. #239 handled the case where owl *returns* `ERROR_ALLOCATION_FAILURE`. But under low heap, owl's generated `parser.h` can `abort()` or dereference a failed allocation **while building the tree** — before that check runs — so the brain still reboots.

Two kinds of site, all reached from `owl_tree_create_from_string()`:

- **`abort()` on OOM during construction** — e.g. `parser.h` `2228`/`2245`/`376`/`5125`, and the token-run realloc at `7899`.
- **Unchecked `calloc` → null-deref** — `parser.h` `5166` (and `5144`), worse than a clean abort.

owl already fails gracefully for the **tokenizer** (`allocation_failed → ERROR_ALLOCATION_FAILURE`), and #239 already handles that error — the construction phase just doesn't use the same mechanism. The complete fix is in `ianh/owl` (make construction propagate the error too); `parser.h` is generated, so it can't be patched here.

A pragmatic lizard-side mitigation that covers realistic lines is in the linked PR.

<details>
<summary>Host MRE (no hardware) — 26 of 27 allocations crash on one valid line</summary>

Fail the *k*-th allocation while parsing `x = 1 + 2`, driving the public API as `process_lizard()` does:

```c
// clang -O0 -fsanitize=address,undefined -I main mre.c -o mre && ./mre <k>
#include <stdlib.h>
#include <stdio.h>
static long at=-1, n=0;
static void*fa(size_t s){ if(at>=0&&n==at){n++;return 0;} n++; return malloc(s); }
static void*fc(size_t a,size_t b){ void*p=fa(a*b); if(p)memset(p,0,a*b); return p; }
static void*fr(void*p,size_t s){ if(at>=0&&n==at){n++;return 0;} n++; return realloc(p,s); }
#define malloc(s) fa(s)
#define calloc(a,b) fc(a,b)
#define realloc(p,s) fr(p,s)
#define OWL_PARSER_IMPLEMENTATION
#include "parser.h"
int main(int c,char**v){ at=c>1?atol(v[1]):-1; struct owl_tree*t=owl_tree_create_from_string("x = 1 + 2");
  printf("k=%ld err=%d\n", at, (int)owl_tree_get_error(t,0)); owl_tree_destroy(t); return 0; }
```

Sweeping `k = 0..26`: `k=0,1` → SEGV null-deref (`5155`, `5167`), `k=2` → graceful `ERROR_ALLOCATION_FAILURE`, `k=3..26` → `abort()`. Only 1 of 27 is handled.

</details>

evnchn:
### Motivation

Under low heap, owl's generated parser can `abort()` or dereference a failed allocation **during** `owl_tree_create_from_string()` — before `process_lizard()` checks `owl_tree_get_error()` — so the brain reboots. #239 fixed the one path where owl *returns* `ERROR_ALLOCATION_FAILURE`; the construction-phase sites still crash (details + host MRE in #240). `parser.h` is generated, so they can't be fixed in-tree; the complete fix is graceful allocation failure upstream in `ianh/owl`.

### Implementation

Skip parsing and drop the line with an error when free heap is below what the parse needs, so owl's allocator is never entered too low. owl's peak transient allocation tracks the line's token count, so the floor scales with length:

```cpp
if (xPortGetFreeHeapSize() < PARSE_HEAP_BASE + PARSE_HEAP_PER_CHAR * strlen(line)) {
    echo("error: not enough free memory to parse");
    return;
}
```

`xPortGetFreeHeapSize()` is the same API already reported as `core.heap`. The constants are conservative starting points measured against `OWL_TOKEN_RUN_LENGTH=256`; since `strlen` is only a proxy for token count, a pathological ultra-dense line can still exceed the estimate (fully solved only by the owl change in #240).

### Progress

- [ ] Tested on hardware — builds for `esp32s3` (idf v5.3.1); on-device threshold tuning still to confirm.

### DIFF
--- a/docs/machine_safety.md
+++ b/docs/machine_safety.md
@@ -35,6 +35,13 @@ so that no stale commands fire after the host stopped:
 when core.last_message_age > 500 then motor.stop(); core.clear_schedule() end
 ```
 
+## Low memory
+
+Parsing a line requires a certain amount of free heap memory, which scales with the length of the line.
+When there is not enough free or contiguous memory, Lizard drops the line and responds with `error: not enough free memory to parse (<free> free, <required> required)` instead of risking a crash and reboot.
+Like a line with an incorrect checksum, the command was not executed;
+the host system should react accordingly, for example by re-sending the command later or stopping the machine.
+
 ## Expander watchdog
 
 The `expander` module provides a watchdog feature that restarts the port expander when it gets stuck and does not send messages anymore.
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -9,6 +9,8 @@
 #include "compilation/rule.h"
 #include "compilation/variable.h"
 #include "compilation/variable_assignment.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
 #include "global.h"
 #include "modules/bluetooth.h"
 #include "modules/core.h"
@@ -305,10 +307,21 @@ void process_tree(owl_tree *const tree, bool from_expander) {
     }
 }
 
+// owl's peak transient allocation scales with the token count of the line, so the
+// required-heap floor scales with its length (base covers a short line's fixed cost).
+static constexpr size_t PARSE_HEAP_BASE = 4096;
+static constexpr size_t PARSE_HEAP_PER_CHAR = 64;
+
 void process_lizard(const char *line, bool trigger_keep_alive, bool from_expander) {
     InterpreterLock lock;
     if (trigger_keep_alive) {
         core_module->keep_alive();
+    }
+
+    // owl's generated parser can abort()/deref under low heap instead of returning an error; drop the line rather than reboot.
+    if (xPortGetFreeHeapSize() < PARSE_HEAP_BASE + PARSE_HEAP_PER_CHAR * strlen(line)) {
+        echo("error: not enough free memory to parse");
+        return;
     }
 
     const bool debug = core_module->get_property("debug")->boolean_value;
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -9,8 +9,8 @@
 #include "compilation/rule.h"
 #include "compilation/variable.h"
 #include "compilation/variable_assignment.h"
+#include "esp_heap_caps.h"
 #include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
 #include "global.h"
 #include "modules/bluetooth.h"
 #include "modules/core.h"
@@ -318,9 +318,13 @@ void process_lizard(const char *line, bool trigger_keep_alive, bool from_expande
         core_module->keep_alive();
     }
 
-    // owl's generated parser can abort()/deref under low heap instead of returning an error; drop the line rather than reboot.
-    if (xPortGetFreeHeapSize() < PARSE_HEAP_BASE + PARSE_HEAP_PER_CHAR * strlen(line)) {
-        echo("error: not enough free memory to parse");
+    // owl's generated parser can abort()/deref on a failed allocation instead of returning an error, so require both
+    // enough total heap (scaled by the line's length) and a large enough contiguous block before parsing. Other tasks
+    // may allocate between the check and the parse; the padded constants absorb that. Drop the line rather than reboot.
+    const size_t required_heap = PARSE_HEAP_BASE + PARSE_HEAP_PER_CHAR * strlen(line);
+    const size_t free_heap = xPortGetFreeHeapSize();
+    if (free_heap < required_heap || heap_caps_get_largest_free_block(MALLOC_CAP_8BIT) < PARSE_HEAP_BASE) {
+        echo("error: not enough free memory to parse (%u free, %u required)", (unsigned)free_heap, (unsigned)required_heap);
         return;
     }
 


falkoschindler (changes requested):
Thanks @evnchn — nice find, and the PR body is admirably clear about what this is (a conservative stopgap until the upstream owl fix in #240) and what it isn't. The guard sits on the single choke point all parse paths go through (UART, expander, startup script), it runs after `keep_alive()` so the watchdog stays fed, and the constants are consistent with `OWL_TOKEN_RUN_LENGTH=256` from main/CMakeLists.txt (line 34). One change I'd like to see before merging, plus a few nice-to-haves:

**Requested: also check the largest free block.** `xPortGetFreeHeapSize()` reports *total* free bytes, not the largest contiguous block. On a fragmented heap the guard can pass while owl's individual allocations (~2.6 KB token runs, plus the `parse_tree` realloc with 1.5× growth in main/parser.h (line 360)) still fail — which lands us right back in `abort()` and a reboot, the exact thing this PR is meant to prevent. Since the failure mode being guarded is an allocation failure, the largest free block is the more faithful proxy. Something like:

```cpp
if (xPortGetFreeHeapSize() < required ||
    heap_caps_get_largest_free_block(MALLOC_CAP_8BIT) < PARSE_HEAP_BASE) {
```

**Nice to have: include the numbers in the error message.** Threshold tuning on hardware is still open, and once this is deployed the echo is the only signal we get. Reporting free and required bytes (e.g. `error: not enough free memory to parse (%u free, %u required)`) would let us tune `PARSE_HEAP_BASE` / `PARSE_HEAP_PER_CHAR` from field logs instead of instrumented builds.

**Nice to have: a word on the false-positive tradeoff.** With `PARSE_HEAP_PER_CHAR = 64`, a full-length line (~250 chars) requires ~20 KB free, so lines that would previously have parsed fine in less can now be dropped. That's still strictly better than a reboot, but hosts should treat the new error as "command not executed" — worth a sentence in the changelog or docs.

**Nice to have: note the TOCTOU in the comment.** Other tasks (Wi-Fi, BLE) allocate concurrently, so the heap snapshot can decay before owl allocates; the padded constants absorb this. A half-sentence in the comment would keep a future tuner from shaving the margins too thin.

**Nit:** `freertos/task.h` looks unused — `xPortGetFreeHeapSize()` comes in via `freertos/FreeRTOS.h`. It mirrors the include block in core.cpp, so keeping it is defensible too.

evnchn:

**All applied** — the largest-free-block check (the blocking one), numbers in the error message, TOCTOU note in the comment, and dropped the unused `task.h`. Thanks @falkoschindler, the fragmentation catch is exactly right.

<details><summary><b>What changed</b></summary>

- **Largest contiguous block (requested):** now also guards `heap_caps_get_largest_free_block(MALLOC_CAP_8BIT) < PARSE_HEAP_BASE`, so a fragmented heap that passes the total-free check but can't fit an individual owl allocation (~2.6 KB token run, or the growing `parse_tree` buffer) no longer slips through to `abort()`.
- **Field-tunable error:** now `error: not enough free memory to parse (%u free, %u required)`.
- **TOCTOU:** comment notes other tasks may allocate between the check and the parse, absorbed by the padded constants.
- **Nit:** dropped `freertos/task.h` (unused — `xPortGetFreeHeapSize` arrives via `FreeRTOS.h`), added `esp_heap_caps.h`.
- Builds green for `esp32s3` (idf v5.3.1); on-device threshold tuning still open.
- **False-positive tradeoff (your doc nice-to-have):** agreed — the new error means "command not executed", not a reboot. There's no CHANGELOG in the repo; happy to add a sentence to `docs/` wherever you'd prefer it to live.

</details>

falkoschindler (approved):
Verified, all applied cleanly — thanks @evnchn! I took the liberty of pushing the doc note directly (1d8926a): a short "Low memory" section in docs/machine_safety.md telling hosts to treat the new error like a failed checksum — the command was not executed. Approving; the on-device threshold tuning can happen after merge now that the error reports the numbers.

falkoschindler:
Heads-up: we are reverting this PR before 0.14.0 ships (#278). Not because the analysis in #240 is wrong — the construction-phase `abort()` sites are real — but because a length-based heap estimate turned out to be the wrong tool for it.

The guard rejected an ordinary 3906-character rosys startup script (`159072 free, 254080 required`) and with it a whole robot configuration, see #272 by @ftilde. A review of that PR showed that no linear constant can work: tuned for real configs it admits expression-dense lines that still hit owl's `abort()`, tuned for those it rejects real configs again. So the guard trades a hypothetical reboot for a real "machine ignores its startup script".

The risk window the guard was meant to close is also narrower than it looks. owl's first allocation is the 2568-byte token run; when the heap is really exhausted, that one fails first and owl reports it as `ERROR_ALLOCATION_FAILURE`, which #239 already handles. The abort sites are only reachable in the few-KB band where the token run still fits but a later small allocation does not, i.e. on a device that is no longer operable anyway.

The proper fix is the one @evnchn named in #240: make owl's construction phase propagate allocation failures the same way its tokenizer already does. We will take that upstream first: ianh already added graceful failure for the token-run allocation in response to ianh/owl#45 and noted that the remaining allocations still lack error handling, so this is a follow-up he has anticipated. A zauberzeug fork or a patch step in `gen_parser.sh` stays as a fallback if upstream stalls. #240 is reopened and scheduled for the next milestone to track that. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Substance%20from%20the%20conversation%20with%20the%20maintainers%2C%20wording%20from%20the%20model.&a=claude-code "claude-fable-5: Substance from the conversation with the maintainers, wording from the model.")

PR #243, opened 2026-07-22
Guard checksum parsing against unparseable suffixes

### LINKED ISSUE #242: Unparseable checksum suffix (e.g. `foo@zz`) escapes as std::invalid_argument and reboots the ESP32
A UART0 line whose checksum suffix `std::stoi` cannot parse at all — e.g. `foo@zz` — throws `std::invalid_argument` out of `check()`. That's a `std::logic_error`, but every enclosing handler catches only `std::runtime_error`, so it escapes to `std::terminate()` → `abort()` → **reboot**.

The nearest non-crashing input differs by two characters:

```python
import serial, time
s = serial.Serial('/dev/ttyACM0', 115200, timeout=0.3)
time.sleep(0.6); s.reset_input_buffer()

s.write(b"foo@00\n")   # control: parseable suffix, wrong checksum -> graceful warning
time.sleep(2); print(s.read(4000).decode('latin-1'))

s.write(b"foo@zz\n")   # bug: unparseable suffix -> abort() -> reboot
time.sleep(6); print(s.read(4000).decode('latin-1'))
```

**Actual output** (ESP32-S3, clean `main` @ `ba866e6`, ESP-IDF v5.3.1):

```
foo@00  ->  warning: Checksum mismatch while processing UART0@0f     <- fine
foo@zz  ->  abort() was called at PC 0x4208c6ff on core 0
            ... Core dump has been saved to flash. Rebooting...
```

<details>
<summary><b>Mechanism + symbolized backtrace</b></summary>

`check()` (`main/utils/uart.cpp:55-70`) validates the **position** of the `@` but never that the suffix parses:

```cpp
const std::string hex_number(&buffer[len - 2], 2);
if (std::stoi(hex_number, 0, 16) != checksum) {   // <- uart.cpp:64, unguarded
```

`std::invalid_argument` derives from `std::logic_error`, not `std::runtime_error`, so the handler around `process_uart()` (`main.cpp:492`) doesn't catch it — nor do the other sites (`main.cpp:435/469/477/511`), which are all `runtime_error`-only.

`addr2line` on the panic backtrace, outermost first:

```
0x40375aee: panic_abort at components/esp_system/panic.c:463
0x403812f9: esp_system_abort at components/esp_system/port/esp_system_chip.c:92
0x403863e9: abort at components/newlib/abort.c:38
0x4208c6ff: __cxxabiv1::__terminate(void (*)()) at libsupc++/eh_terminate.cc:48
0x4208c734: std::terminate() at libsupc++/eh_terminate.cc:58
0x4208c80f: __cxa_throw at libsupc++/eh_throw.cc:98
0x4208d64b: std::__throw_invalid_argument(char const*) at c++11/functexcept.cc:78
0x4203c10d: __gnu_cxx::__stoa<long, int, char, int>(...) at ext/string_conversions.h:85
0x4204dd6d: std::__cxx11::stoi(...) at bits/basic_string.h:4109
 (inlined by) check(char*, int, bool*) at /project/main/utils/uart.cpp:64
0x4201a654: process_uart() at /project/main/main.cpp:422
0x4201a8e9: app_main at /project/main/main.cpp:492
0x420b0b2a: main_task at components/freertos/app_startup.c:208
0x4038166e: vPortTaskWrapper at .../portable/xtensa/port.c:134
```

`check()` is shared by a few other receive paths; I only reproduced UART0 on hardware.

</details>

<details>
<summary><b>Scope: only wholly unparseable suffixes crash</b></summary>

`std::stoi` accepts partial parses and leading whitespace/signs, so several malformed suffixes are *accepted* rather than throwing. Measured on the same board:

| line | result |
|---|---|
| `foo@zz` | **abort → reboot** |
| `foo@1z` | warning (parses as `0x1`) |
| `foo@ f` | warning (parses as `0xf`) |
| `foo@+f` | warning (parses as `0xf`) |
| `foo@00` | warning |

So this is specifically "the suffix has no leading parseable digit", not "the suffix isn't two hex digits".

</details>

<details>
<summary><b>Suggested fix</b></summary>

Minimal crash fix — treat a parse failure as what it already looks like, a checksum mismatch, matching the guard `SerialBus::parse_message` already uses on its own `stoi` (`serial_bus.cpp:220-229`):

```cpp
try {
    if (std::stoi(hex_number, 0, 16) != checksum) {
        ok = false;
    } else {
        len -= 3;
    }
} catch (...) {
    ok = false;
}
```

`foo@zz` then echoes the checksum-mismatch warning instead of rebooting; valid-checksum behaviour is unchanged.

If you'd also want strict two-hex-digit validation (rejecting the `1z` / ` f` / `+f` cases in the table above), that's a separate behaviour change — happy to include it, but I've kept it out of the crash fix. Happy to open a PR in whichever shape you prefer.

</details>

<details>
<summary><b>Repro environment</b></summary>

- Board: ESP32-S3 (chip rev v0.2, 16 MB flash), USB-CDC on `/dev/ttyACM0`
- Firmware: clean clone of `origin/main` @ `ba866e6` (`v0.13.0-3-gba866e6`), built with `python3 build.py esp32s3` in `espressif/idf:v5.3.1`, flashed via `espresso.py flash --chip esp32s3`
- Still reproduces after the recent parser/crash hardening — this path was untouched by it.

</details>

evnchn:
Fixes #242.

`check()` calls `std::stoi` on the two-character checksum suffix with no guard. A suffix it can't parse — e.g. the line `foo@zz` — throws `std::invalid_argument`, which is a `std::logic_error`, so it slips past the `runtime_error`-only handlers → `std::terminate()` → **reboot**. Four bytes of line noise reboot the controller.

This treats a parse failure as what it already looks like — a checksum mismatch — matching the guard `SerialBus::parse_message` already uses on its own `std::stoi`.

**With whitespace hidden ([`?w=1`](https://github.com/zauberzeug/lizard/pull/243/files?w=1)) the change is 4 added lines and nothing removed:**

```diff
         const std::string hex_number(&buffer[len - 2], 2);
+        try {
             if (std::stoi(hex_number, 0, 16) != checksum) {
                 ok = false;
             } else {
                 len -= 3;
             }
+        } catch (...) {
+            ok = false;
+        }
```

No new locals, no comments in the code, no behaviour change for any input that didn't already throw.

<details>
<summary><b>Before / after on hardware — full matrix</b></summary>

Same ESP32-S3, same build path (`python3 build.py esp32s3` in `espressif/idf:v5.3.1`, flashed via `espresso.py`), pristine `main` vs this branch.

**Before** (`ba866e6`): `foo@zz` → `abort()` at PC `0x4208c6ff`, coredump saved, REBOOT.

**After:**

| input | result | note |
|---|---|---|
| `foo@zz` | `warning: Checksum mismatch while processing UART0@0f` | **was: abort → reboot** |
| `foo@1z` | `warning: Checksum mismatch …` | unchanged (partial-parses to `0x1`) |
| `foo@ f` | `warning: Checksum mismatch …` | unchanged |
| `foo@+f` | `warning: Checksum mismatch …` | unchanged |
| `foo@00` | `warning: Checksum mismatch …` | unchanged |
| `1 + 1@2b` | `2@32` | **valid checksum still executes — `len -= 3` path intact** |
| `1 + 1` | `2@32` | unchecksummed control |
| `bogus_prop@2e` | `error processing uart0: unknown variable "bogus_prop"@0b` | valid checksum, suffix stripped correctly |

Observed failing without the patch and passing with it on the same board.

</details>

<details>
<summary><b>Why <code>catch (...)</code> rather than validating hex</b></summary>

Strict two-hex-digit validation would *also* reject `1z` / ` f` / `+f`, which `std::stoi` currently accepts via partial parse. That's a grammar change rather than a crash fix, so it's deliberately left out — happy to add it separately if you want that behaviour.

`catch (...)` is scoped to the `stoi` call and its branch body only. `std::string hex_number(...)` stays **outside** the `try`, so a `bad_alloc` from constructing it still escapes instead of being silently reported as a checksum mismatch.

`len -= 3` can't leave inconsistent state: the only throwing expression in the `try` is `stoi` itself, which runs before any mutation.

</details>

<details>
<summary><b>Checks</b></summary>

- `clang-format` 22.1.5 (the pinned `mirrors-clang-format` rev): clean, zero diff.
- Builds green for `esp32s3`; run locally, since CI here is pre-commit-only.
- No test added — there's no host test harness for `uart.cpp` in the repo. The hardware matrix above is the evidence.

</details>

### DIFF
--- a/main/utils/uart.cpp
+++ b/main/utils/uart.cpp
@@ -61,10 +61,14 @@ int check(char *buffer, int len, bool *checksum_ok) {
             checksum ^= buffer[i];
         }
         const std::string hex_number(&buffer[len - 2], 2);
-        if (std::stoi(hex_number, 0, 16) != checksum) {
+        try {
+            if (std::stoi(hex_number, 0, 16) != checksum) {
+                ok = false;
+            } else {
+                len -= 3;
+            }
+        } catch (...) {
             ok = false;
-        } else {
-            len -= 3;
         }
     }
     buffer[len] = 0;


falkoschindler (approved):
Thanks @evnchn, great catch and a very well-documented fix! Verified the mechanism (all handlers in main.cpp are `runtime_error`-only, so the `std::invalid_argument` really did escape to a reboot) and the guard itself: the only throwing expression inside the `try` is `stoi`, `len -= 3` still only runs on a successful match, and it mirrors the existing pattern in SerialBus::parse_message. The hardware matrix is much appreciated. LGTM!

PR #244, opened 2026-07-22
Consolidate serial device detection across the host tools

pascalzauberzeug:
### Motivation

The host tools answered the question which serial device to use three different ways: `monitor.py` globbed a hard-coded path list (often the wrong adapter, never an ESP32-S3 on macOS), `espresso.py` resolved a platform default, and `otb_update.py` assumed `/dev/ttyUSB0`.

### Implementation

Rebuilt along the review below: glob the well-known device nodes instead of enumerating with pyserial.

- Add `serial_devices.py` as the single answer to which serial device a host tool talks to, used by `monitor.py`, `espresso.py`, `otb_update.py` and `configure.py`
- Glob the well-known USB-serial nodes; on a Jetson, derive the UART from the L4T version instead (R35 → `ttyTHS0`, R36 → `ttyTHS1`)
- Add `resolve_device()`: connects straight away on a unique match, otherwise lists the candidates and asks
- Replace `requirements-macos.txt` with a `sys_platform` marker on `gpiod`

<details>
<summary>Design decisions and behavior changes</summary>

- **`resolve_device()` is cached with `lru_cache`,** and `Config.port` in `espresso.py` reads it lazily: the pin-only commands (`enable`, `disable`, `reset`, `release_pins`) never ask about a port they do not open, and nothing needs pyserial for detection — `sudo ./espresso.py flash` works without it being installed for root.
- **Asking rather than guessing.** The first candidate is not necessarily the Lizard, and the callers go on to flash whatever comes back. The exchange happens on stderr so a redirected stdout cannot hide the prompt; without a terminal it raises naming the candidates, `Ctrl+C` exits 130.
- **A dry run without an adapter fails with `No serial device found`** instead of printing a placeholder — the honest answer, as settled in the review.
- **`monitor_ssh.sh` sends `monitor.py` and `serial_devices.py` along** (and `espresso.py --host` sends `serial_devices.py`), so the age of the target's checkout no longer decides which UART is used.
- **Other changes:** `otb_update.py --port` and `configure.py`'s device path are now optional; a failed connect reports as `Error: <reason>` instead of a traceback; every tool still accepts an explicit path, which is what non-interactive callers should use.

</details>

### Progress

- [ ] Tested on hardware: a Robot Brain run (both L4T UARTs, `monitor_ssh.sh`) and two adapters attached at once are still open.

<details>
<summary>What was checked by hand</summary>

On a macOS host: all pin-only commands run to completion with stdin closed; an explicit `--device` reaches the esptool call; `--host` rsyncs `serial_devices.py` along; the multi-device case without a terminal names the candidates and asks for an explicit path; a unique match connects immediately and is cached across reads.

</details>

---
⬛ claude-fable-5 (rework after review) · claude-opus-4-8 (initial commit)


JensOgorek:
This got fairly big for what it does. Most of `serial_devices.py` follows from one decision — enumerating with `pyserial`'s `list_ports` instead of globbing the device nodes — and each of its consequences pulled in the next: the `vid is None` filter, the extra glob for the macOS nodes `list_ports` misses, the `realpath` dedup between those two sources, the deferred import plus the long "install pyserial for root" error and doc paragraph, and finally the `uses_serial` declaration, `Config.port` and the `ask=`/`allow_missing=` block in `espresso.py` that exist because resolving is now expensive enough to want to skip.

Globbing the well-known nodes — what one types by hand anyway — drops that whole chain. A version that does is ~60 lines instead of 158, and shrinks the `espresso.py` diff from 83 lines to roughly 25. Worth a rebuild before this lands, I think.

<details>
<summary>Sketch and trade-offs</summary>

```python
PATTERNS = ['/dev/ttyUSB*', '/dev/ttyACM*', '/dev/cu.usbserial*', '/dev/cu.usbmodem*',
            '/dev/cu.wchusbserial*', '/dev/cu.SLAB_USBtoUART*']


def find_devices() -> List[str]:
    if IS_JETSON:  # L4T 35 -> ttyTHS0, 36 -> ttyTHS1, unchanged from this PR
        ...
    return sorted(path for pattern in PATTERNS for path in glob.glob(pattern))


@functools.lru_cache(maxsize=None)  # so a command reading the device twice asks at most once
def resolve_device(path: Optional[str] = None) -> str:
    if path:
        return path
    devices = find_devices()
    if not devices:
        raise RuntimeError('No serial device found')
    if len(devices) == 1:
        return devices[0]
    if not sys.stdin.isatty():
        raise RuntimeError(f'Multiple serial devices found: {", ".join(devices)}. Pass the device path explicitly.')
    ...  # list them on stderr and ask, as this PR already does
```

**What the glob removes for free.** Built-in and virtual ports (`/dev/ttyS*`, `cu.Bluetooth-*`, `cu.debug-console`) never match a pattern, so the `vid is None` filter is unnecessary rather than merely correct. There is one source of paths instead of two, so nothing has to be deduplicated. And nothing imports `pyserial`, so `sudo ./espresso.py flash` works without it being installed for root — which takes the deferred import, its error message, the doc paragraph about it, and the `requirements-macos.txt` entry in `zip_content.txt` with it.

**`lru_cache` replaces `uses_serial`.** With resolution cached, `Config.port` can be `return self.device or resolve_device()` and resolve lazily on first read. A pin-only command (`enable`, `disable`, `reset`, `release_pins`) then never asks because it never reads the port — no declaration to keep in sync, and the reverse mistake the `Config.port` error message guards against cannot happen at all. That removes the `uses_serial` field, its five entries in `COMMANDS`, and the resolution block in `main()`.

**What is given up:**

- *Descriptions in the prompt.* With two adapters attached the list is `[0] /dev/ttyUSB0` / `[1] /dev/ttyUSB1` rather than naming the bridges. On Linux the `/dev/serial/by-id/*` symlinks carry the name in the path and could bring it back in a few lines; on macOS the node names (`cu.SLAB_USBtoUART`, `cu.wchusbserial*`) already identify the chip.
- *Adapters whose node matches no known pattern.* `list_ports` finds those without being told; a glob has to be. The one plausible case is a Prolific PL2303 under macOS, whose vendor driver names its node `/dev/cu.PL2303G-USBtoUART*` — one more pattern, if it ever shows up. `--device` covers everything else, and the CP210x/CH34x/FTDI/CDC-ACM nodes we actually use are all in the list above.
- *The `--dry-run` placeholders.* Lazy resolution makes a dry run without an adapter attached fail with `No serial device found` instead of printing `<no serial device attached>`. Arguably the honest answer, but it is a behavior change against what this PR describes, so worth deciding rather than inheriting.

The Jetson half is unaffected: several platform UARTs exist on a Robot Brain, so the L4T version really is the only way to pick the right one, and that part should stay as it is.

</details>

falkoschindler (changes requested):
@pascalzauberzeug I agree with @JensOgorek's analysis, and I'd like this PR rebuilt along his sketch before it lands. The problem itself is real — we shouldn't close this — but the solution has gotten too heavy.

The fixes are worth keeping: monitor.py silently picks the wrong adapter when several are attached and can't find an ESP32-S3 on macOS at all (native USB, cu.usbmodem*); otb_update.py hard-codes /dev/ttyUSB0; the L4T mapping (R35 → ttyTHS0, R36 → ttyTHS1) only lives in espresso.py while monitor_ssh.sh passes ttyTHS1 unconditionally — the wrong UART on an L4T 35 Robot Brain.

But almost all of the size follows from a single decision: enumerating with `list_ports` instead of globbing the well-known device nodes. That pulls in the `vid is None` filter, the extra glob for the macOS nodes, the realpath dedup between the two sources, the deferred import with the "pyserial under sudo" error and doc paragraph, and finally `uses_serial`/`Config.port` in espresso.py, because resolving became expensive enough to want to skip. The only real gain is the device descriptions in the selection prompt — nice, but the node names (cu.SLAB_USBtoUART, cu.wchusbserial*) already identify the chip on macOS, and `--device` covers exotic adapters.

Concretely:

- Shrink serial_devices.py to Jens's version (~60 instead of 158 lines): glob over the known patterns, the Jetson L4T logic unchanged, `resolve_device` with `lru_cache`, the selection prompt on stderr as it is now. No pyserial for detection → the whole sudo problem, the requirements-macos.txt entry in the release zip and the doc paragraph about it all disappear.
- Let `Config.port` in espresso.py resolve lazily (`self.device or resolve_device()`). The pin-only commands then never ask because they never read the port — `uses_serial`, its five entries in `COMMANDS` and the resolution block in `main()` all go away. The espresso.py diff shrinks from 83 to ~25 lines.
- The small callers (monitor.py, configure.py, otb_update.py) stay almost as they are — they're already minimal. Keep the monitor_ssh.sh rsync too: without sending monitor.py along, an old checkout on the target decides wrong again.
- The one open behavior question from Jens's comment I'd settle in favor of simplicity: `espresso.py --dry-run` with no adapter attached may fail with "No serial device found" instead of printing a placeholder. That's the more honest answer, and it removes the `ask=`/`allow_missing=` machinery along with the two placeholder constants.

That should land the PR at roughly +150/−80 instead of +295/−79, with what remains being almost entirely the actual fix. I wouldn't go for the minimal alternative (fixing only monitor.py) either: that would leave the Jetson logic duplicated, otb_update.py broken on macOS and monitor_ssh.sh wrong on L4T 35 — exactly the motivation of this PR.

pascalzauberzeug:
I reworked the PR with your reviews in mind.


<details>
<summary>Claude had some more notes that are not implemented</summary>

Remarks (all minor, no blockers)

1. monitor_ssh.sh can no longer pass an explicit device path through. Previously ttyTHS1 was hard-coded (the actual bug); now there is no way to give the remote monitor a specific device. On a Robot Brain, detection is the right default; anyone who does need an explicit path has to ssh in manually. Worth noting, but acceptable.
2. find_devices() does not check that the UART node exists on a Jetson — the failure only surfaces when opening the port, as a SerialException. The message is still clear (Error: could not open port …), so this is fine.
3. The hardware test is still open according to the PR checklist (Robot Brain with both L4T UARTs, monitor_ssh.sh, two adapters attached at once). The L4T 35/36 distinction is exactly the part only hardware can verify — this should happen before merging, as the unchecked box itself says.
4. upload_ssh.py is now just a deprecation shim with no device logic of its own — correctly left untouched.

And the assessment of whether to act on them: 1 and 2 should deliberately not be built in — adding a pass-through option would reintroduce exactly the kind of speculative feature Jens and Falko reviewed out of the PR, and an existence check would merely swap one clear error message for another. 3 is the only real open item, and it's testing, not code. 4 requires nothing.

</details>

falkoschindler (changes requested):
Hi @pascalzauberzeug, I like where this lands — one detection path instead of three fall-through lists, and the reasoning in the docstrings (why only the `cu.` nodes, why an unreadable L4T version raises instead of falling back) makes the tricky parts reviewable. I checked the head at `9b5fdf6`, not the older revisions of the branch.

**Up front:** these findings are not mine. I had Claude Code (Opus 5) go through the branch, and everything below — the analysis, the severity calls and the suggested patches — is its output, lightly checked by me. Please treat it as a starting point rather than a verdict, and push back where it is wrong.

Two of the five look blocking to me (1 and 4); the rest are decisions to make explicitly rather than defects.

### 1. The device is resolved after the ESP is already in the bootloader — blocking

espresso.py (133): `Config.port` is lazy, and in `flash()` and `erase()` the first read happens inside the esptool argument list, i.e. inside `with _flash_mode(config)`. By then the pins have been toggled, so the "which device?" prompt appears after the microcontroller has been driven into the download bootloader — and with no tty, or with nothing attached, the pins are driven and the command then aborts. A cron/CI invocation knocks a running ESP out of its program for a command that was never going to work.

The laziness itself is right, and the docstring's reason is good: a pin-only command should not resolve a device it never opens. Only the commands that do open one need to resolve it before touching hardware. Cheapest fix is one line at the top of each such handler, since `lru_cache` makes the later reads free:

```python
def flash(config: Config) -> None:
    """Flash the microcontroller."""
    print_bold('Flashing...')
    port = config.port  # resolve (and possibly ask) before the pins move
    with _pin_config(config):
        ...
```

and then pass `port` into the esptool arguments, which also documents that both invocations use the same device. Same for `erase()`; `reset_partition()` needs nothing extra as long as it is only reached from `flash()`.

The alternative is resolving in `main()` behind a `uses_serial` flag on `Command`, the way the earlier revision of this branch did — but that reintroduces the flag `e4f050f` deliberately removed, so I would only go back to it if more commands grow a serial dependency.

### 2. `--dry-run` now needs hardware, and that is undocumented

espresso.py (190, 481): `flash -d`, `erase -d` and `coredump -d` exit 1 with "No serial device found" on a machine with nothing attached, where they used to print the full esptool line. The PR body says this is deliberate, and it is defensible — the printed command should be the one that would actually run. But `--dry-run` is not mentioned in `docs/tools.md` at all, and its help is just `'Dry run'`, so the first person to hit this has nothing to go on.

```python
parser.add_argument('-d', '--dry-run', action='store_true',
                    help='Print the commands instead of running them (still resolves the serial device)')
```

plus a sentence in `docs/tools.md` next to the auto-detection note: "`--dry-run` prints the commands instead of running them. It still resolves the serial device, so it needs the adapter attached (or an explicit `--device`)."

Fixing 1 helps here too: with the port resolved up front, a dry run fails before printing half a run rather than in the middle of it.

### 3. `otb_update.py` loses its scripted callers

otb_update.py (17): dropping the `/dev/ttyUSB0` default for `--port` is right interactively, but a cron/CI caller that worked before now dies with "Cannot ask which of several serial devices to use" as soon as a second node is present — and `/dev/ttyACM*` is globbed now, so a modem or a second dev board is enough.

I would not restore the default; a silent `/dev/ttyUSB0` was its own trap, and the error already names the candidates and says what to do. But it deserves a release-note line: `otb_update.py --port` no longer defaults to `/dev/ttyUSB0`, and non-interactive callers must pass it explicitly. Your call whether that is enough or whether the compatible variant (keep auto-detection, prefer `/dev/ttyUSB0` when it is among the candidates) is worth the wart.

### 4. A Jetson with a USB adapter becomes undetectable — blocking

serial_devices.py (54): `find_devices()` returns the L4T-derived UART on any Jetson without checking that the node exists, and never falls back to the USB patterns. A Jetson with the ESP on a USB bridge — a dev setup, or a Robot Brain with an external adapter — is unreachable by auto-detection, where the old `monitor.py` fall-through list found it.

```python
def find_devices() -> List[str]:
    """Return the serial devices that could be a microcontroller."""
    uart = jetson_uart()
    usb = sorted(path for pattern in PATTERNS for path in glob.glob(pattern))
    if uart is not None and Path(uart).exists():
        return [uart] + usb
    return usb
```

The Jetson UART stays first, so the ordinary Robot Brain case is still a single candidate and never prompts. A Jetson with an extra adapter attached now asks instead of silently picking the pin header, which is the right question — those are genuinely two different microcontrollers.

One thing to decide explicitly: with this change, a Jetson whose UART node is missing falls back to the USB patterns instead of failing. The argument in `jetson_uart()` — better to raise than to hide the cause — still holds for an *unreadable L4T version*; this only relaxes the case where the version is known but the node is not there.

### 5. `Config.port` is the only place that says "port" instead of "device"

espresso.py (133): everything else this PR touches speaks "device" — `serial_devices.py`, `find_devices()`, `resolve_device()`, the "No serial device found" error, espresso's own `--device` flag, monitor.py's positional `device`, and `docs/tools.md` throughout. `Config.port` is the one place that switches vocabulary.

The word is also already taken: in monitor.py (93) `port` is the *open* `serial.Serial` object while `device` is the path string, which is a useful distinction. In `Config` it is a path string again, so the same word now means two things across two files this PR just unified.

And within `Config` the split does not track what a reader expects. `device` and `port` look like two different things, but they are the same thing at two stages — the raw `--device` argument and the resolved result. `return self.device or resolve_device()` gives that away immediately.

Suggestion: keep "device" as the vocabulary and let "port" survive only where it is someone else's spelling (esptool's `--port` flag):

```python
@dataclass(frozen=True)
class Config:
    device_option: Optional[str] = None  # an explicit --device; otherwise .device detects one on first read

    @property
    def device(self) -> str:
        """The serial device, detected on first read. ..."""
        return self.device_option or resolve_device()
```

Call sites then read `'--port', config.device` — our concept, their flag name. It is a mechanical rename of one field and one property.

Related, and possibly out of scope: `otb_update.py` calls its flag `--port` while `espresso.py` calls it `--device`, and `docs/tools.md` documents both spellings for the same concept. Since point 3 already changes that flag's behaviour, this might be the moment to rename it to `--device` and keep `--port` as an alias — or to decide explicitly that it stays as it is.

PR #245, opened 2026-07-22
Compute serial checksums over UTF-8 bytes in the Python tools

falkoschindler:
### Motivation

Counterpart to zauberzeug/rosys#440: the Python tools in this repo XOR Unicode **code points** (`ord(c)`), while the firmware XORs the raw **bytes** it receives over the UTF-8 serial transport. For any non-ASCII line the two sides disagree, and above U+00FF the `'{:02x}'` format overflows to three hex digits, so the line does not even round-trip with itself. For ASCII the byte and code-point computations are identical, so existing traffic is unaffected.

### Implementation

All four Python files now compute checksums over `line.encode()` instead of `ord(c)`:

- configure.py (31, 48) — the per-line XOR in send() and the sum-based startup checksum, matching what `core.startup_checksum()` computes over the stored bytes.
- monitor.py (54, 66) — receive verification and send. The send loop is rewritten as a plain split('\n') loop; the old index-juggling variant prefixed every segment after the first with a stray `\n` when multi-line text was pasted.
- examples/trajectory_planner/serial_connection.py (25, 42)
- examples/ros/scripts/lizard.py (14, 47)

Riding along, the same guard as in the rosys PR and #242: the two examples parsed `int(line[-2:], 16)` unguarded, so a corrupted checksum suffix (e.g. `foo@zz`) raised `ValueError` out of the read loop. They now treat unparseable suffixes as a failed check. (The ROS example also no longer crashes on `line.split('@')` when the payload itself contains an `@`, and the `reduce(ixor, ...)` calls got a `0` initializer so an empty line no longer raises `TypeError`.)

docs/machine_safety.md now says "XOR of all bytes of the UTF-8 encoded line" instead of "of all characters" — the ASCII example table is unchanged.

Verified against the test vectors pinned in zauberzeug/rosys#440 (`grün@04`, `café@0e`, `ß@5c`, `日本@02`, `1 + 2@28`) plus the byte-sum startup checksum (`grün` → `02c6`); pre-commit passes on all changed files.

### Progress

- [x] Tested on hardware.

### DIFF
--- a/configure.py
+++ b/configure.py
@@ -27,8 +27,8 @@ def send(payload: str) -> None:
         line_ = payload
     print(f'Sending: {line_}')
     checksum_ = 0
-    for c in line_:
-        checksum_ ^= ord(c)
+    for byte in line_.encode():
+        checksum_ ^= byte
     port.write((f'{line_}@{checksum_:02x}\n').encode())
 
 
@@ -45,7 +45,7 @@ def read(*, timeout: float) -> Iterator[str]:
 
 with serial.Serial(args.device_path, baudrate=args.baud, timeout=1.0) as port:
     startup = Path(args.config_file).read_text('utf-8') + '\n'
-    checksum = sum(ord(c) for c in startup) % 0x10000
+    checksum = sum(startup.encode()) % 0x10000
 
     send('!-')
     for line in startup.splitlines():
--- a/docs/machine_safety.md
+++ b/docs/machine_safety.md
@@ -8,7 +8,7 @@ Each line sent via the command-line interface can and should be followed by a ch
 Lizard will omit any lines with incorrect checksums.
 Any output is as well sent with a checksum.
 
-The 8-bit checksum is computed as the bitwise XOR of all characters excluding the newline character and written as a two-digit hex number (with leading zeros) separated with an `@` character, for example:
+The 8-bit checksum is computed as the bitwise XOR of all bytes of the UTF-8 encoded line excluding the newline character and written as a two-digit hex number (with leading zeros) separated with an `@` character, for example:
 
 | Line    | Bitwise XOR                             | Result     |
 | ------- | --------------------------------------- | ---------- |
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -25,23 +25,23 @@ It is automatically created right after the boot sequence.
 | `core.heap`             | Free heap memory (bytes)                                        | `int`     |
 | `core.last_message_age` | Time since last input message was received and interpreted (ms) | `int`     |
 
-| Methods                          | Description                                        | Arguments    |
-| -------------------------------- | -------------------------------------------------- | ------------ |
-| `core.restart()`                 | Restart the microcontroller                        |              |
-| `core.version()`                 | Show lizard version                                |              |
-| `core.info()`                    | Show lizard version, compile time and IDF version  |              |
-| `core.print(...)`                | Print arbitrary arguments to the command line      | arbitrary    |
-| `core.output(format)`            | Define the output format                           | `str`        |
-| `core.startup_checksum()`        | Show 16-bit checksum of the startup script         |              |
-| `core.get_pin_status(pin)`       | Print the status of the chosen pin                 | `int`        |
-| `core.set_pin_level(pin, value)` | Turns the pin into an output and sets its level    | `int`, `int` |
-| `core.get_pin_strapping(pin)`    | Print value of the pin from the strapping register | `int`        |
-| `core.forget_serial_bus()`       | Remove the saved SerialBus configuration from NVS  |              |
-| `core.set_baudrate(baud)`        | Persist UART0 baud rate (applied after restart)    | `int`        |
-| `core.pause_broadcasts()`        | Pause property broadcasts (all modules)            |              |
-| `core.resume_broadcasts()`       | Resume property broadcasts                         |              |
-| `core.clear_schedule()`          | Discard all pending scheduled blocks               |              |
-| `core.keep_alive()`              | Reset `last_message_age` without producing output  |              |
+| Methods                          | Description                                                         | Arguments    |
+| -------------------------------- | ------------------------------------------------------------------- | ------------ |
+| `core.restart()`                 | Restart the microcontroller                                         |              |
+| `core.version()`                 | Show lizard version                                                 |              |
+| `core.info()`                    | Show lizard version, compile time and IDF version                   |              |
+| `core.print(...)`                | Print arbitrary arguments to the command line                       | arbitrary    |
+| `core.output(format)`            | Define the output format                                            | `str`        |
+| `core.startup_checksum()`        | Show 16-bit checksum of the startup script (sum of its UTF-8 bytes) |              |
+| `core.get_pin_status(pin)`       | Print the status of the chosen pin                                  | `int`        |
+| `core.set_pin_level(pin, value)` | Turns the pin into an output and sets its level                     | `int`, `int` |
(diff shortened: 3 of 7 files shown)


JensOgorek:
Reviewed with **Claude Opus 5 (1M context)** via `/review`, and verified on hardware: built this branch for esp32s3, flashed a bench ESP32-S3, and reproduced every test vector from the description against real firmware — plus the ASCII no-op case. Everything checks out, so the *Tested on hardware* box can be ticked. No blocking findings; one optional firmware hardening and a few nits below.

<details>
<summary><b>Hardware verification</b> — what was actually run</summary>

Firmware built from this branch (`v0.13.0-3-g248f4ed`), flashed to an ESP32-S3 on `/dev/ttyUSB0`. Round-trips use the `!"` echo control command, so the firmware both validates the incoming checksum and emits its own over the echoed payload.

| Sent | Firmware reaction |
| --- | --- |
| `!"grün@07` (bytes) | echoes `grün@04` ✔ — the vector from the description |
| `!"grün@84` (old, `ord`) | `warning: Checksum mismatch while processing UART0` |
| `!"café@0d` / `@8e` | accepted / rejected |
| `!"日本@01` (bytes) | echoes `日本@02` ✔ |
| `!"日本@2ca` (old, 3-digit) | no `@` at position −3, so **`日本@2ca` reaches the interpreter as an unchecked payload** |
| `!"1 + 2@2b` | identical either way (ASCII unaffected) |

The 3-digit overflow case is slightly worse than "does not round-trip with itself": the line isn't merely rejected, it bypasses validation entirely and is parsed as-is.

Startup checksum via `configure.py` with a `.liz` containing `# grün: Ümlaut-Test mit 日本`:

```
Sending: core.startup_checksum()
Checksum matches.        →  device:            checksum: 1510
                            new (bytes):       0x1510  ✔
                            old (code points): 0xdccb  ✘
```

An ASCII-only config also reports `Checksum matches.`, so no regression for existing setups.

The `monitor.py` send rewrite was diffed offline against the old loop and both byte streams were written to the device: old `b'core.info()@3a\n\ncore.millis@2b\n'`, new `b'core.info()@3a\ncore.millis@2b\n'`. The stray `\n` was harmless for Lizard (the firmware consumes it as a line separator), but the new loop is clearly cleaner. Empty prompt input yields `@00\n` in both versions — no behavior change.

</details>

<details>
<summary><b>Findings</b> — one optional hardening, four nits</summary>

1. **Implicit dependency on `char` signedness** (optional, firmware side). `main/modules/core.cpp:79-82` sums `for (char const &c : Storage::startup) checksum += c;`. That `sum(startup.encode())` matches it relies on `char` being unsigned in the toolchain (`xtensa-esp32s3-elf-gcc -dM -E` reports `__CHAR_UNSIGNED__ 1`). With signed `char` the sum would be off by 256 per byte ≥ 0x80 — the test above would have produced `0x0810` instead of `0x1510`. A `checksum += static_cast<uint8_t>(c);` would pin this down. Nothing is broken today, so a follow-up issue is fine.
1. **Inconsistent encode argument.** `monitor.py` uses `.encode('utf-8')`, the other three files use `.encode()`. Identical in effect, but a PR about exactly this topic could unify it.
1. **`split('\n')` keeps a trailing `\r`.** For a CRLF paste the `\r` ends up inside the segment and thus inside the checksum. Firmware and Python agree (the checksum passes), but the interpreter then sees `foo\r` and reports a parse error. Pre-existing, not a regression; `segment.rstrip('\r')` would cover it.
1. **Adjacent pre-existing bug in `examples/trajectory_planner/serial_connection.py:22-23`.** The guard tests `if '\n' in self.buffer` but the split is on `'\r\n'`, so a bare `\n` raises `ValueError`. Same block this PR touches — worth taking along, otherwise out of scope.
1. **Docs.** `docs/machine_safety.md` is correctly updated. `docs/module_reference.md:35` still describes `core.startup_checksum()` only as "16-bit checksum of the startup script"; since `configure.py` has to reproduce it, a "sum of the UTF-8 bytes" clause would be consistent.

Not raised as findings: the four duplicated XOR loops (reasonable for four standalone scripts that must not import from the repo root), and the missing automated tests (the repo has no Python test infrastructure, CI runs pre-commit only — the hardware run above covers the vectors).

</details>

<details>
<summary><b>Scope for the human reviewer</b></summary>

Correctness, hardware behavior, and the ASCII-compatibility claim are covered — no need to redo those mechanically. What still benefits from a human eye is a **superficial pass on two judgment calls**: whether finding 1 (`static_cast<uint8_t>`) should go into this PR or a follow-up, and whether the nits 2–5 are worth the extra churn here. No deep bug hunt needed; the diff is small and byte-for-byte compatible for ASCII traffic.

Hardware state: run on a bench ESP32-S3 (not a robot), firmware built from this branch at `248f4ed`, no special feldfreund software state involved.

</details>

falkoschindler:
@JensOgorek Ok, I agreed on all 5 findings and updated the code accordingly. Ready for your approval! 😉

JensOgorek (approved):
Re-tested on hardware at `6f77e5f` with **Claude Opus 5 (1M context)**: all five findings are cleanly addressed, no regressions, and the byte-checksum behavior is confirmed against real firmware. Approving. 👍

<details>
<summary><b>Hardware re-verification</b> — bench ESP32-S3, firmware built from this branch</summary>

Firmware rebuilt from this branch (`v0.13.0-8-g6f77e5f`, target `esp32s3`, `--clean`) and flashed to a bench ESP32-S3 on `/dev/ttyUSB0`. Round-trips use the `!"` echo control command, so the firmware both validates the incoming checksum and emits its own over the echoed payload.

| Sent | Firmware reaction |
| --- | --- |
| `!"grün@07` | echoes `grün@04` ✔ |
| `!"café@0d` | echoes `café@0e` ✔ |
| `!"ß@5f` | echoes `ß@5c` ✔ |
| `!"日本@01` | echoes `日本@02` ✔ |
| `!"1 + 2@2b` | echoes `1 + 2@28` ✔ (byte == code point, ASCII unaffected) |

The old code-point checksums (`@84`, `@8e`, `@dc`) are all rejected with `Checksum mismatch`, as expected. The 3-digit overflow case `日本@2ca` still bypasses validation entirely (no `@` at position −3, so it reaches the interpreter as an unchecked payload) — unchanged from before, and after this PR nothing produces such a line anymore.

**Startup checksum** (finding 1, `static_cast<uint8_t>` in `d89277c`). A `.liz` containing `# grün: Ümlaut-Test mit 日本` was applied via `configure.py`:

```
device:              checksum: 1828
byte sum (new):      0x1828  ✔
code points (old):   0xdfe3  ✘
signed char:         0x0e28  ✘
```

`configure.py` reports `Checksum matches.`, and the value is unambiguously the byte sum — the cast is a no-op on this toolchain (`char` is unsigned there) but pins the behavior down. The stored script round-trips intact via `!?` (`# grün: Ümlaut-Test mit 日本@4a`).

</details>

<details>
<summary><b>The four Python fixes</b> — offline edge cases</summary>

`serial_connection.py` (findings 3+4) now accepts bare-LF and CRLF alike, and no longer raises on an unparseable suffix (`foo@zz`) or an `@` inside the payload (`a@b@43`); the ROS guard behaves identically, including the empty line `@00`.

The rewritten `monitor.py` send loop (finding 3) produces byte-identical output to the old one for a single line and for empty input, drops the stray `\n` on multi-line pastes (`b'core.info()@3a\ncore.millis@2b\n'` instead of `...@3a\n\ncore...`), and strips the `\r`. That last one matters in practice: with the `\r` inside the payload the firmware answers `error: invalid token at range 11 12 "\r"`, without it the line executes.

Docs (finding 5) and the unified `.encode()` (finding 2) check out as described.

</details>

PR #248, opened 2026-07-28
Fix I2C bus setup on ESP32-S3

### LINKED ISSUE #247: ESP32-S3: no I²C module can be constructed — DEFAULT_TIMEOUT_TICKS exceeds the S3 timeout register

**TL;DR: on ESP32-S3, no I²C module can be constructed at all** — `Imu`, `ImuBno085` and `Mcp23017` all throw before touching the bus, whatever is (or isn't) wired up.

`DEFAULT_TIMEOUT_TICKS = 1048575` in `main/utils/i2c_bus.cpp` is the ESP32's maximum timeout value; the ESP32-S3's timeout register is five bits wide, so `i2c_set_timeout()` rejects it.

A second, smaller bug sits right behind it: when `ensure()` throws there, the already-installed driver is never unwound, so every *subsequent* I²C construction fails with a different and misleading error.

Repro is one line typed at the Lizard prompt — no I²C device, no wiring, no startup script needed.

## Reproduction

Flash `main` @ `020b32d` to any ESP32-S3 with nothing attached to the I²C pins, and type:

```
imu = Imu()
```

**Actual (ESP32-S3):**

```
E (11110) i2c: i2c_set_timeout(948): i2c timing value error
error processing uart0: could not set i2c timeout
```

**Expected** — what the identical command does on an ESP32, also with nothing attached:

```
error processing uart0: imu setup failed: I2CError: Check your wiring.
```

That is, on ESP32 the bus initializes and the *device* is correctly reported as absent. On ESP32-S3 the bus never initializes, so an S3 can't talk to a BNO055, a BNO085 or an MCP23017 even when one is correctly wired.

<details>
<summary><b>Cause</b></summary>

[`main/utils/i2c_bus.cpp:8`](https://github.com/zauberzeug/lizard/blob/020b32d/main/utils/i2c_bus.cpp#L8):

```cpp
constexpr uint32_t DEFAULT_TIMEOUT_TICKS = 1048575;
```

`1048575` is `0xFFFFF`. In ESP-IDF v5.3.1:

- ESP32 — [`I2C_TIME_OUT_REG_V`](https://github.com/espressif/esp-idf/blob/v5.3.1/components/soc/esp32/include/soc/i2c_reg.h#L163) is `0xFFFFF`, and [`I2C_LL_MAX_TIMEOUT`](https://github.com/espressif/esp-idf/blob/v5.3.1/components/hal/esp32/include/hal/i2c_ll.h#L840) maps to it. `1048575` is exactly that limit, so it is accepted.
- ESP32-S3 — [`I2C_TIME_OUT_VALUE_V`](https://github.com/espressif/esp-idf/blob/v5.3.1/components/soc/esp32s3/include/soc/i2c_reg.h#L248) is `0x0000001F`, and [`I2C_LL_MAX_TIMEOUT`](https://github.com/espressif/esp-idf/blob/v5.3.1/components/hal/esp32s3/include/hal/i2c_ll.h#L951) maps to that. `i2c_set_timeout()` range-checks against it and returns an error.

So the constant is silently ESP32-specific. Everything funnels through the same call — `imu.cpp:55`, `imu_bno085.cpp:134` and `mcp23017.cpp:30` all call `I2cBusManager::ensure(...)`, which throws at [`i2c_bus.cpp:37`](https://github.com/zauberzeug/lizard/blob/020b32d/main/utils/i2c_bus.cpp#L37) before any device I/O. All three were exercised on the S3 and all three fail identically.

I haven't proposed a fix here because the right one is a judgement call: clamp to `I2C_LL_MAX_TIMEOUT`, pick a per-target constant, or drop the explicit `i2c_set_timeout()` and take the driver default. Happy to send a PR once you've said which you'd prefer.

</details>

<details>
<summary><b>Second bug: the successful driver install is never unwound</b></summary>

In `ensure()`, `i2c_driver_install()` succeeds and then `i2c_set_timeout()` throws. The function leaves without calling `i2c_driver_delete()`, and `config.initialized` is never set — so the port is left with a live driver that the manager believes it has not installed.

The consequence is order-dependent and deterministic. Whichever I²C module is constructed **first** reports the real error; every later one reports a misleading one:

```
>>> mcp = Mcp23017()
error processing uart0: could not set i2c timeout          ← the true cause

>>> imu = Imu()
error processing uart0: could not install i2c driver       ← misleading
```

Reversing the order swaps which module reports which error, confirming it is position and not module identity:

```
>>> imu = Imu()
error processing uart0: could not set i2c timeout

>>> mcp = Mcp23017()
error processing uart0: could not install i2c driver
```

This is latent on ESP32 rather than fixed — `ensure()` only leaks when `i2c_set_timeout()` is the call that fails, which on ESP32 it currently cannot. Worth unwinding regardless, so a future failure between install and success doesn't poison the port.

</details>

<details>
<summary><b>What was run</b></summary>

- Source: `zauberzeug/lizard` @ `020b32d`, clean clone with submodules, ESP-IDF v5.3.1, `sdkconfig.defaults.secret` from the shipped template.
- **ESP32-S3** — QFN56 rev v0.2, 16 MB flash, built with `./build.py esp32s3`, stock `partitions.csv`, nothing attached to the I²C pins. This is the board the bug is about.
- **ESP32** (the control) — ESP32-D0WD-V3, 4 MB flash, also with nothing attached. Two build-config deviations were needed to fit a 4 MB dev board, neither touching this code path: `coredump` dropped from `partitions.csv`, and `CONFIG_ESPTOOLPY_FLASHSIZE` 8MB → 4MB. Hence `-dirty` in its `core.info()`.
- Every block of output above is an interactive UART transcript, pasted verbatim: the board was booted with an **empty startup script**, then the module constructor was typed at the prompt at 115200. That is why the errors read `error processing uart0: ...` rather than `error while loading startup script: ...` — the same failure reaches both paths, but the repro deliberately uses the interactive one so there is no startup script to configure.
- All three I²C modules (`Imu`, `ImuBno085`, `Mcp23017`) were constructed on the S3; each fails identically when it is the first one called.
- Deterministic — no timing dependence, fails identically on every boot.
- Not tested: an S3 with an I²C device actually present. The failure happens before any device I/O, so it should be unaffected, but I haven't confirmed that on hardware.

</details>

evnchn:
*Opened by Claude Code on evnchn's behalf.*

**TL;DR: 3-line fix — no I²C module can be constructed on ESP32-S3.** Fixes #247.

`DEFAULT_TIMEOUT_TICKS` was the ESP32's `I2C_LL_MAX_TIMEOUT` written out as a literal (`1048575`). The S3's timeout register is five bits wide, so `i2c_set_timeout()` rejected it and `Imu`, `ImuBno085` and `Mcp23017` all threw before touching the bus. Using the constant the driver actually range-checks against makes it correct on both targets.

The second hunk deletes the driver on that failure path. It had been installed successfully by then but was never unwound, so the *first* failing module reported the real cause and every one after it reported a misleading `could not install i2c driver`.

```diff
+#include "hal/i2c_ll.h"
+// I2C_LL_MAX_TIMEOUT is a cycle count on ESP32 (~13 ms) but an exponent on ESP32-S3 (~54 s)
-constexpr uint32_t DEFAULT_TIMEOUT_TICKS = 1048575;
+constexpr uint32_t DEFAULT_TIMEOUT_TICKS = I2C_LL_MAX_TIMEOUT;

     if (i2c_set_timeout(port, DEFAULT_TIMEOUT_TICKS) != ESP_OK) {
+        i2c_driver_delete(port);
         throw std::runtime_error("could not set i2c timeout");
     }
```

<details>
<summary><b>Before / after on hardware, both chips</b></summary>

One line typed at the prompt, nothing attached to the I²C pins:

| `imu = Imu()` | before | after |
|---|---|---|
| **ESP32-S3** | `could not set i2c timeout` | `imu setup failed: I2CError: Check your wiring.` |
| **ESP32** | `imu setup failed: I2CError: Check your wiring.` | *unchanged* |

After the fix the S3 behaves exactly like the ESP32 — the bus initializes and the *device* is correctly reported absent. `Mcp23017()` likewise reaches device I/O (`unable to send i2c command`) rather than failing at bus setup.

The ESP32 column is a pure regression check; behaviour is identical either way.

NOTE: `core.info()`'s `compile time` is not a reliable freshness signal on an incremental build — it comes from a translation unit ninja may not recompile. Since ESP32 behaviour is unchanged by this fix, I confirmed the board was running the patched binary from object/link timestamps rather than from `core.info()`.

</details>

<details>
<summary><b>The unwind hunk, shown failing without it</b></summary>

With the timeout fixed the throw is unreachable, so verifying the second hunk needs the failure forced. Built the S3 with `DEFAULT_TIMEOUT_TICKS = I2C_LL_MAX_TIMEOUT + 1` to guarantee `i2c_set_timeout()` fails, then constructed all three I²C modules in sequence:

**Without `i2c_driver_delete` (current code):**

```
>>> imu = Imu()          → could not set i2c timeout        ← true cause
>>> mcp = Mcp23017()     → could not install i2c driver     ← misleading
>>> b = ImuBno085()      → could not install i2c driver     ← misleading
```

**With `i2c_driver_delete` (this PR):**

```
>>> imu = Imu()          → could not set i2c timeout
>>> mcp = Mcp23017()     → could not set i2c timeout
>>> b = ImuBno085()      → could not set i2c timeout
```

Every module reports the real cause regardless of construction order. The temporary `+ 1` was reverted and the tree re-verified before committing.

</details>

<details>
<summary><b>On the <code>hal/i2c_ll.h</code> include</b></summary>

Two things I checked before proposing it, since it is a private ESP-IDF header:

- **It is required.** Removing the include fails the build: `error: 'I2C_LL_MAX_TIMEOUT' was not declared in this scope`. It is not reachable transitively through `driver/i2c.h`.
- **It matches existing practice here.** `main/modules/core.cpp` already includes `hal/gpio_hal.h`, plus `soc/io_mux_reg.h` and `soc/soc.h`; `main/modules/stepper_motor.cpp` includes `soc/gpio_sig_map.h`.

I chose `I2C_LL_MAX_TIMEOUT` because it is exactly the constant `i2c_set_timeout()` validates against, so the value cannot drift out of agreement with the check when a new target is added. The alternatives, if you would rather not take the dependency:

1. Per-target constants behind `#if CONFIG_IDF_TARGET_ESP32 / _ESP32S3` — public, but duplicates a value that already exists and needs maintaining per new target.
2. Drop the explicit `i2c_set_timeout()` and take the driver default from `i2c_param_config()` — smallest diff, but silently changes ESP32 behaviour. The long timeout looks deliberate — 88687e8 ("fix i2c timeouts", 2022) raised it from `30000` to `1048575`, i.e. ~0.4 ms to ~13 ms for the BNO055 — so I did not.

Happy to switch to either.

</details>

<details>
<summary><b>Against the CONTRIBUTING checklist</b></summary>

- **Compiles without warnings** — clean build for both targets, no warnings from `i2c_bus.cpp`.
- **Follows style guidelines** — `pre-commit run clang-format` passes.
- **Tested on target hardware** — ESP32-S3 (QFN56 rev v0.2, 16 MB) and ESP32 (D0WD-V3, 4 MB), both with an empty I²C bus. ESP-IDF v5.3.1.
- **No debug prints left in code.**
- **Documentation** — no user-facing behaviour change to document; the DSL surface is unchanged.

Not tested: an S3 with an I²C device actually present. Bus setup now succeeds, so device I/O should proceed normally, but I have not confirmed that on hardware. The 4 MB ESP32 test rig needed two local build-config deviations to fit the board (`coredump` dropped from `partitions.csv`, `CONFIG_ESPTOOLPY_FLASHSIZE` 8MB → 4MB); neither touches this code path and neither is part of this branch.

</details>

### DIFF
--- a/main/utils/i2c_bus.cpp
+++ b/main/utils/i2c_bus.cpp
@@ -1,11 +1,12 @@
 #include "i2c_bus.h"
 
-#include "hal/i2c_ll.h"
 #include <stdexcept>
+#include "hal/i2c_ll.h"
 
 namespace {
 constexpr int I2C_MASTER_TX_BUF_DISABLE = 0;
 constexpr int I2C_MASTER_RX_BUF_DISABLE = 0;
+// I2C_LL_MAX_TIMEOUT is a cycle count on ESP32 (~13 ms) but an exponent on ESP32-S3 (~54 s)
 constexpr uint32_t DEFAULT_TIMEOUT_TICKS = I2C_LL_MAX_TIMEOUT;
 } // namespace
 
--- a/main/utils/i2c_bus.cpp
+++ b/main/utils/i2c_bus.cpp
@@ -1,11 +1,12 @@
 #include "i2c_bus.h"
 
+#include "hal/i2c_ll.h"
 #include <stdexcept>
 
 namespace {
 constexpr int I2C_MASTER_TX_BUF_DISABLE = 0;
 constexpr int I2C_MASTER_RX_BUF_DISABLE = 0;
-constexpr uint32_t DEFAULT_TIMEOUT_TICKS = 1048575;
+constexpr uint32_t DEFAULT_TIMEOUT_TICKS = I2C_LL_MAX_TIMEOUT;
 } // namespace
 
 std::map<i2c_port_t, I2cBusManager::BusConfig> I2cBusManager::configs;
@@ -35,6 +36,7 @@ void I2cBusManager::ensure(i2c_port_t port, gpio_num_t sda_pin, gpio_num_t scl_p
         throw std::runtime_error("could not install i2c driver");
     }
     if (i2c_set_timeout(port, DEFAULT_TIMEOUT_TICKS) != ESP_OK) {
+        i2c_driver_delete(port);
         throw std::runtime_error("could not set i2c timeout");
     }
 


JensOgorek:
Independently verified on hardware with **Claude Opus 5 (1M context)**: the diagnosis holds, before/after reproduced on an ESP32-S3, and the ESP32 path is a provable no-op. No blocking findings. Two style nits, one nuance worth a code comment, and one separate pre-existing bug this fix makes reachable — that last one probably deserves its own issue rather than being pulled into this PR.

<details>
<summary><b>Root cause confirmed statically</b>, plus the ESP32 no-op argument</summary>

`I2C_LL_MAX_TIMEOUT` is `0xFFFFF = 1048575` on ESP32 — bit-for-bit the old literal, so the ESP32 column of your table is not just "unchanged in testing" but unchanged by construction — and `0x1F = 31` on the S3, which is exactly what `i2c_set_timeout()` range-checks against (`driver/i2c/i2c.c:948`).

The unwind hunk is sound: `configs[port]` is value-initialized, so `initialized` stays `false` and a later `ensure()` re-runs the full sequence against a freshly deleted driver. The other throw path (`could not install i2c driver`) correctly needs no unwind — nothing is installed at that point.

</details>

<details>
<summary><b>Before / after on an ESP32-S3</b>, empty bus</summary>

**Before** (firmware without the fix), three modules in sequence:

```
imu = Imu()        → i2c timing value error / could not set i2c timeout   ← true cause
mcp = Mcp23017()   → could not install i2c driver                          ← misleading
bno = ImuBno085()  → could not install i2c driver                          ← misleading
```

Constructed in the opposite order the true message moves to whichever module runs first, confirming the order-dependence you describe.

**After** (this branch, built and flashed): bus setup succeeds and every module reaches device I/O — `imu setup failed: I2CError: Check your wiring.` and `unable to send i2c command`.

The ESP32 target builds cleanly, no warnings from `i2c_bus.cpp`. Not flashed — only an S3 was on the bench — but see the no-op argument above.

</details>

<details>
<summary><b>One nuance: "maximum" means something very different per target</b></summary>

The two timeout registers are not encoded alike. `i2c_ll_calculate_timeout_us_to_reg_val()` is linear on ESP32 (`clk_cycle_num_per_us * timeout_us`) but logarithmic on the S3 (`31 - __builtin_clz(...)`, i.e. the field is an exponent). In wall-clock terms:

| | register value | default source clock | effective timeout |
| --- | --- | --- | --- |
| ESP32 | 1048575 cycles | APB 80 MHz | **≈ 13 ms** |
| ESP32-S3 | 2³¹ cycles | XTAL 40 MHz | **≈ 54 s** |

Worth knowing because `31` looks alarmingly small at first glance — it is the opposite. It also puts the original intent in context: the value dates to `88687e8` ("fix i2c timeouts", 2022), which raised `30000` → `1048575`, i.e. ~0.4 ms → ~13 ms for the BNO055. So a concrete duration was meant, not "whatever the register tops out at" — and on the S3 that now inherits a hardware timeout on the order of a minute. (Minor: the PR description attributes the value to #199; `88687e8` is where it comes from.)

In practice this is defused by `i2c_master_cmd_begin()` having its own FreeRTOS-level `ticks_to_wait` — errors came back promptly in testing. I would keep `I2C_LL_MAX_TIMEOUT` for exactly the drift argument you give, and just note the per-target encoding in a comment. Deriving the register value from a fixed microsecond duration would be the target-independent alternative, but it leans even harder on the private HAL header.

</details>

<details>
<summary><b>Two nits</b></summary>

1. **Include order contradicts CONTRIBUTING.md.** Lines 42-43 there prescribe `<standard_library>` before `"project_header.h"`; the new `#include "hal/i2c_ll.h"` sits *above* `<stdexcept>`. `i2c_bus.h` itself follows the convention (`<map>`, then `"driver/i2c.h"`). Note that green pre-commit says nothing here — `.clang-format` has `SortIncludes: false`.
1. **`DEFAULT_TIMEOUT_TICKS` is now a misnomer**, since on the S3 the value is an exponent rather than a tick count. Something like `MAX_TIMEOUT_REG_VALUE` would carry the nuance above without needing a comment at all.

Not raised as findings: the private-header dependency (your reasoning and the `hal/gpio_hal.h` precedent in `core.cpp` both hold up — it is the constant the driver validates against, so it cannot drift), and the absence of automated tests (the firmware has none; the CONTRIBUTING hardware checklist is satisfied, and forcing `MAX + 1` to reach the otherwise-unreachable unwind path is the right way to cover that hunk).

</details>

<details>
<summary><b>Separate finding</b> — <code>ImuBno085()</code> hard-resets the S3 with default arguments</summary>

Out of scope here, but recording it since this fix is what makes the path reachable — I'd suggest a separate issue rather than widening this PR.

On the patched firmware, `bno = ImuBno085()` with defaults resets the chip after ~1.4 s, 3 out of 3 attempts on an empty bus:

```
rst:0x8 (TG1WDT_SYS_RST),boot:0x8 (SPI_FAST_FLASH_BOOT)
W (46) boot.esp32s3: PRO CPU has been reset by WDT.
```

`imu_bno085.cpp:17-18` hardcodes `int_pin = GPIO_NUM_26` and `rst_pin = GPIO_NUM_32`. Both are ESP32 pin numbers, and unlike SDA/SCL — target-aware via `DEFAULT_I2C_SDA_PIN` / `DEFAULT_I2C_SCL_PIN` in `module_helpers.h:12-16` — they are not conditional on the target. On the S3 those two are `SPICS1` and `SPID` (`soc/esp32s3/include/soc/io_mux_reg.h:116-122`), i.e. SPI flash pins, so the reset line takes the flash down with it.

Control: `bno = ImuBno085(0, 8, 9, 4, 5)` fails cleanly with `BNO085 initialization failed` and no reset — bus and module logic are fine, only the defaults are wrong for this target. Before #248 the module threw at bus setup, well before touching any GPIO, so the S3 never got there.

</details>

evnchn:
Thank you — the 88687e8 catch is a straight correction and the per-target encoding is exactly the thing that wanted saying out loud. Pushed as 792a749: include moved, encoding noted in a one-line comment, PR description fixed. Declined the rename (below). Your separate BNO085 finding is filed as #250, with the hardware evidence attributed to you.

Back over to you, @JensOgorek — I would reassign it myself but I only have read access here, so could you take the assignee back off me?

<details>
<summary><b>Include order</b> — moved as asked, but the tree disagrees with CONTRIBUTING here</summary>

Moved, since it costs nothing:

```diff
-#include "hal/i2c_ll.h"
 #include <stdexcept>
+#include "hal/i2c_ll.h"
```

One thing worth putting on the record before you accept it, though. Of the 37 `.cpp` files under `main/` whose leading include block mixes both kinds, **36 put project headers before standard ones** — and after this change `i2c_bus.cpp` is the only one that does not:

```
quoted-before-angle : 36
angle-before-quoted : 1   -> main/utils/i2c_bus.cpp
```

`core.cpp`, `imu.cpp`, `mcp23017.cpp` and `stepper_motor.cpp` all end their block with `<stdexcept>` / `<memory>` / `<algorithm>` after the quoted ones. The CONTRIBUTING snippet you cite is under the **`### Header Files`** heading, next to `#pragma once` — and `i2c_bus.h` does follow it, as you say; `module_helpers.h` does not.

So this is either a doc that undersells what the `.cpp` files actually do, or 36 files quietly drifting — not something this PR should decide. Happy either way: say the word and I revert the hunk, or leave it and `i2c_bus.cpp` is the first file following the letter of the doc.

</details>

<details>
<summary><b>The comment</b>, and why I kept <code>DEFAULT_TIMEOUT_TICKS</code></summary>

```cpp
// I2C_LL_MAX_TIMEOUT is a cycle count on ESP32 (~13 ms) but an exponent on ESP32-S3 (~54 s)
constexpr uint32_t DEFAULT_TIMEOUT_TICKS = I2C_LL_MAX_TIMEOUT;
```

On the rename: `MAX_TIMEOUT_REG_VALUE` is the better name, and I would take it in a fresh file. I am leaving it because renaming touches every use site for no behaviour change, and a bigger diff on a fix that wants to be obviously-safe reads as more expensive than it is. If you would rather have the name than the comment, it is a one-liner and I will do it.

I confirmed your numbers against my own IDF v5.3.1 checkout rather than taking them on trust — all four hold:

| claim | source | value |
| --- | --- | --- |
| ESP32 `I2C_LL_MAX_TIMEOUT` | `soc/esp32/include/soc/i2c_reg.h:163` | `I2C_TIME_OUT_REG_V` = `0xFFFFF` = 1048575 |
| S3 `I2C_LL_MAX_TIMEOUT` | `soc/esp32s3/include/soc/i2c_reg.h:248` | `I2C_TIME_OUT_VALUE_V` = `0x1F` = 31 |
| ESP32 encoding | `hal/esp32/include/hal/i2c_ll.h` | `clk_cycle_num_per_us * timeout_us` — linear |
| S3 encoding | `hal/esp32s3/include/hal/i2c_ll.h` | `31 - __builtin_clz(...)` — exponent |

Default source clocks likewise: `I2C_CLK_SRC_DEFAULT` is `SOC_MOD_CLK_APB` on ESP32 and `SOC_MOD_CLK_XTAL` on the S3 (`clk_tree_defs.h:301` / `:319`), which is where your 80 MHz / 40 MHz — and so ≈13 ms vs 2³¹/40 MHz ≈ 53.7 s — come from. The old literal being bit-for-bit `0xFFFFF` is a nice way to put the ESP32 no-op beyond doubt; I had only argued it from behaviour.

</details>

<details>
<summary><b>Rebuilt both targets</b> after the change</summary>

Clean builds from a wiped `build/` on ESP-IDF v5.3.1, no warnings from `i2c_bus.cpp` on either:

```
```

Include reordering cannot change behaviour here, but "it still compiles on both" is cheap to actually check rather than assert. No hardware re-test — the change is a comment and an include line; the flashed behaviour in the table above is unchanged.

</details>

JensOgorek (approved):
Both my nits dissolve. Re-verified `792a749` on an ESP32-S3 — nothing blocking, approving. Thanks for filing #250.

**Include order:** my nit was wrong, withdrawn — leave the include wherever you prefer, not worth a round trip. **Rename:** dropping it; your comment states the per-target encoding, which was the part that needed saying.

<details>
<summary>Why the nit was wrong</summary>

Counted independently: of the `.cpp` files under `main/` mixing both kinds, **37 of 37** put project headers first, so your 36:1 holds. And the doc snippet sits under `### Header Files` next to `#pragma once` — it describes headers, not `.cpp` files. My error, not a drift in your code. Whether the doc should be widened isn't this PR's question.

</details>

<details>
<summary>Hardware re-verification</summary>

Clean esp32s3 build, no warnings from `i2c_bus.cpp`, flashed, firmware reports `v0.13.0-6-g792a749`.

| | `792a749` | `02550ca` |
| --- | --- | --- |
| `imu = Imu()` | `imu setup failed: I2CError: Check your wiring.` | identical |
| `mcp = Mcp23017()` | `unable to send i2c command` | identical |
| `bno = ImuBno085()` after `Imu()` | `i2c bus already initialized with different parameters` | identical |
| reverse order, fresh boot | both reach device I/O | identical |

Row three is the parameter guard doing its job — `ImuBno085` defaults to 400 kHz against 100 kHz — pre-existing, unrelated. #250 still reproduces here: defaults reset the chip, `ImuBno085(0, 8, 9, 4, 5)` fails cleanly.

</details>

@falkoschindler — what do you think? The one judgement call worth a second opinion is taking a dependency on the private `hal/i2c_ll.h` for `I2C_LL_MAX_TIMEOUT`; the alternatives are laid out in the PR description. Correctness and hardware behaviour are covered.

falkoschindler:
@evnchn was closing this intentional? It was approved after the hardware re-verification, but the close carries no comment and coincides to the second with the deletion of the head branch — which looks like branch cleanup auto-closing the PR rather than a deliberate withdrawal.

As it stands the fix has been dropped on the floor: main still has the old literal in main/utils/i2c_bus.cpp (line 8), #247 is still open, and #251 deliberately doesn't touch the I²C bus setup — it even assumes this fix exists ("Before #248 the module threw at bus setup…").

Should we re-open?

evnchn:
@falkoschindler this was NOT intentional. Sorry...😞

PR #249, opened 2026-07-28
Report which module failed in the startup script, and make it queryable

### LINKED ISSUE #246: A failing I²C device (e.g. wrong/absent IMU) aborts the rest of the startup script — while the Core still reports `Ready.`
## Summary

If a module constructor throws while the startup script is being applied, every statement **after** the failing one is silently dropped, yet the Core still prints `Ready.` and keeps answering on UART. A single mis-configured or unresponsive I²C device — for example a `BNO055` IMU expected at `0x28` when the board actually carries a `BNO085` at `0x4A`, or no IMU at all — therefore leaves the Core in an ordering-dependent, partially-configured state that looks alive from the outside.

We hit this during RB70 bring-up: the wrong IMU driver kept the board from coming up in a usable state.

Note: whether a missing device *should* stop the Core is a legitimate design question (a safety-relevant sensor may well be a hard requirement). The point of this issue is that the **current** behavior is neither a clean fail-safe nor a clean fail-open — it's an accidental middle state. This needs to be discussed and decided here before implementing anything.

## Root cause

`process_tree()` in [`main/main.cpp`](main/main.cpp) iterates the startup script's statements in a single loop **without per-statement isolation**:

```cpp
for (struct owl_ref r = statements.statement; !r.empty; r = owl_next(r)) {
    ...
    const Module_ptr module = Module::create(module_type, module_name, arguments, process_lizard);
    ...
}
```

When a constructor throws (e.g. `Imu::Imu` → `BNO055::begin()` → `throw BNO055ChipNotDetected()`, re-wrapped as `"imu setup failed: ..."`), the exception propagates out of `process_tree` → out of `process_lizard`, and is only caught one level up in `app_main`:

```cpp
try {
    Storage::init();
    process_lizard(Storage::startup.c_str());          // whole startup script in one call
    echo("error while loading startup script: %s", e.what());   // <-- remaining statements never run
}
...
printf("\nReady.\n");                                   // still prints "Ready."
```

Consequences:

- **Ordering-dependent.** Only statements *after* the failing one are lost. A safety-relevant module declared *before* the bad device stays up; one declared *after* it is never created. What survives depends on the position in the startup script, not on any policy.
- **Masked by `Ready.`** The Core signals ready and stays responsive regardless, so a host cannot tell from the ready signal that the config is incomplete.
- **Free-text-only diagnosis.** The failing device is reported only as a log string; there is no structured way to learn which module/address failed after boot.

A secondary observation on time-bounding: constructors on the I²C path can block (e.g. `BNO055::reset()` does `vTaskDelay(700ms)`, and I²C reads rely on a per-transaction timeout). The observed failure mode here is the abort-the-rest behavior above, not an infinite hang — but it's worth deciding whether individual device init should be explicitly time-bounded too.

## Reproduction

1. Put a startup script on a Robot Brain that declares an `imu` on an address where no BNO055 responds (wrong address, wrong chip, or nothing connected), followed by further module declarations (e.g. `wheels`, `expander`, `bluetooth`).
2. Boot.
3. Observe: `error while loading startup script: imu setup failed: ...` followed by `Ready.`; the modules declared after the `imu` line do not exist, yet the Core reports ready.

## To discuss / decide

The core policy question: **what should happen when a required device fails at startup, and how does a device get classified as required vs optional?** A few directions, not mutually exclusive — to be weighed here:

- **Explicit fail-safe.** A device failure puts the Core into a defined, diagnosable error state and does **not** signal `Ready.`, instead of the current ordering-dependent partial config. Makes "don't run without this sensor" an intentional behavior rather than a side effect.
- **required vs optional per module.** Let a module be marked optional so its failure is skipped and the rest of the config still builds, while a *required* module's failure triggers the fail-safe halt above. This keeps the "some sensors are mandatory" case explicit.
- **Where the policy lives.** Whether a given sensor is safety-critical is arguably a host/integration decision (RoSys and the robot's safety layer), so Lizard may only need to provide the mechanism — a clean fault report, a required/optional distinction, and an honest ready/health signal — and leave the decision to the host.
- **Structured, queryable fault reporting.** Regardless of the above: report the affected module and, for I²C devices, the bus/port and address in a machine-parseable form (e.g. `error: module "imu" init failed on i2c0 @0x28: chip not detected`), and consider a queryable Core/module health state so a host can ask which startup modules failed after `Ready.`.

---

_Note: drafted by Claude for Falko._

evnchn:
*Opened by Claude Code on evnchn's behalf.*

**Draft on purpose — this is a discussion artifact for #246, not a merge request.** That issue says the policy question needs deciding before anything is implemented, and I agree. So this deliberately implements *only* the part of it that holds under every outcome of that decision, and leaves the decision alone.

**What it does (14 lines):**

1. The startup-script error names the failing module and its type — `module "imu" (Imu): imu setup failed: ...` instead of just `imu setup failed: ...`.
2. That failure is recorded and readable at any time via `core.startup_error()`, so a host connecting after `Ready.` can discover the configuration is incomplete.

**What it deliberately does NOT do:** change which statements run after a failure, or whether `Ready.` is printed. Those are the fail-safe and required-vs-optional questions, and they stay open.

## Why this scope

Of the four directions in #246, three are policy. The fourth is prefixed *"Regardless of the above"* — report the affected module in a machine-parseable form, and offer a queryable health state so a host can ask what failed after `Ready.`. That is the part that is useful under every outcome, so it is the only part built here.

Not implemented from that fourth bullet: the bus/port and address in the error (`on i2c0 @0x28`). That needs plumbing through each I²C module, and the right shape depends on whether the answer becomes a list — better done once the format is agreed.

<details>
<summary><b>Hardware verification (ESP32, absent BNO055)</b></summary>

Startup script `before = Output(4)` / `imu = Imu()` / `after = Output(5)`, empty I²C bus.

**Boot — the error now identifies the module:**

```
error while loading startup script: module "imu" (Imu): imu setup failed: I2CError: Check your wiring.

Ready.
```

**Post-boot — the failure is queryable, which it previously was not:**

```
>>> core.startup_error()
startup_error: module "imu" (Imu): imu setup failed: I2CError: Check your wiring.
```

**Negative control — a clean startup script reports empty:**

```
>>> core.startup_error()
startup_error:
```

**Behaviour deliberately unchanged** — the module after the failing line is still dropped, exactly as before:

```
>>> after.on()
error processing uart0: unknown module "after"
```

**The name is the instance, not a hardcoded string.** This matters because the existing message hardcodes `"imu setup failed"` whatever you name the module:

```
>>> myimu = Imu()
error processing uart0: module "myimu" (Imu): imu setup failed: I2CError: Check your wiring.
```

</details>

<details>
<summary><b>Open questions — the reason this is a draft</b></summary>

1. **Is `core.startup_error()` the right surface?** A single string is the smallest thing that answers "did my config apply?". But if the policy lands on per-statement isolation there could be *several* failures, and this wants to be a list. Building the list now would presuppose that outcome.
2. **Is the message format right?** `module "name" (Type): ...` is parseable but ad-hoc. If a more structured form is wanted, that is better decided once — together with the bus/address plumbing — than changed twice.
3. **Redundancy with existing messages.** `module "imu" (Imu): imu setup failed: ...` double-reports the name because `imu.cpp` hardcodes its own prefix. Cleaning that up across modules is a wider sweep; it would bloat this diff, so I left it.
4. **Scope of the module-name context.** Only constructor failures get it. A failure in a method call or rule during the startup script takes a different path through `process_tree` and is unchanged.

</details>

<details>
<summary><b>Against the CONTRIBUTING checklist</b></summary>

- **Compiles without warnings** — clean for both `esp32` and `esp32s3`, no warnings from the touched files.
- **Follows style guidelines** — `pre-commit run clang-format` passes.
- **Tested on target hardware** — ESP32 (D0WD-V3, 4 MB), empty I²C bus, ESP-IDF v5.3.1. Behaviour above is pasted verbatim. ESP32-S3 was compile-verified but the runtime behaviour above was not re-run on it.
- **No debug prints left in code.**
- **Documentation** — `core.startup_error()` is a new DSL method and would need a docs entry before this leaves draft. Deliberately not written yet, since the surface is question 1 above.

Tested in isolation: the ESP32-S3 I²C fix (#248) was reverted out of the tree first, so none of the above is confounded by it. The 4 MB test rig needed two local build-config deviations that are not part of this branch (`coredump` dropped from `partitions.csv`, `CONFIG_ESPTOOLPY_FLASHSIZE` 8MB → 4MB).

</details>

Happy to take this in a different direction entirely, or close it, once the policy question in #246 is settled — the point is to make that discussion concrete, not to pre-empt it.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -24,6 +24,7 @@ It is automatically created right after the boot sequence.
 | `core.millis`           | Time since booting the microcontroller (ms)                     | `int`     |
 | `core.heap`             | Free heap memory (bytes)                                        | `int`     |
 | `core.last_message_age` | Time since last input message was received and interpreted (ms) | `int`     |
+| `core.startup_error`    | Why the startup script did not run completely (empty if it did) | `str`     |
 
 | Methods                          | Description                                        | Arguments    |
 | -------------------------------- | -------------------------------------------------- | ------------ |
@@ -48,6 +49,11 @@ The `precision` is an optional integer specifying the number of decimal places f
 For example, the format `"core.millis input.level motor.position:3"` might yield an output like `"92456 1 12.789"`.
 
 `core.get_pin_status(pin)` reads the pin's voltage, not the output state directly.
+
+**Startup errors:**
+A failing startup script is reported once during boot, before `Ready.`, which a host connecting later never sees.
+`core.startup_error` keeps that message available afterwards, so a host can check whether the persisted configuration was applied at all.
+It names the module that could not be created, e.g. `"module \"imu\" (Imu): imu setup failed: I2CError: Check your wiring."`, and is also set by a syntax error, which drops the whole script.
 
 **UART baud rate:**
 The console (UART0) defaults to 115200 baud.
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -330,7 +330,13 @@ struct StartupLoadScope {
 static void report_parse_error(const std::string &message) {
     echo("error: %s", message.c_str());
     if (loading_startup) {
-        core_module->get_property("startup_error")->string_value = message;
+        std::string value = message;
+        for (char &c : value) {
+            if (static_cast<unsigned char>(c) < ' ') {
+                c = ' '; // the offending token can be a newline, which would split the property over two lines
+            }
+        }
+        core_module->get_property("startup_error")->string_value = value;
     }
 }
 
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -211,7 +211,12 @@ void process_tree(owl_tree *const tree, bool from_expander) {
                 }
                 const std::string module_type = identifier_to_string(constructor.module_type);
                 const std::vector<ConstExpression_ptr> arguments = compile_arguments(constructor.argument);
-                const Module_ptr module = Module::create(module_type, module_name, arguments, process_lizard);
+                Module_ptr module;
+                try {
+                    module = Module::create(module_type, module_name, arguments, process_lizard);
+                } catch (const std::exception &e) {
+                    throw std::runtime_error("module \"" + module_name + "\" (" + module_type + "): " + e.what());
+                }
                 Global::add_module(module_name, module);
             } else {
                 const std::string module_name = identifier_to_string(constructor.module_name);
@@ -475,6 +480,7 @@ void app_main() {
         Storage::init();
         process_lizard(Storage::startup.c_str());
     } catch (const std::runtime_error &e) {
+        Core::startup_error = e.what();
         echo("error while loading startup script: %s", e.what());
     }
 
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -317,6 +317,23 @@ void process_tree(owl_tree *const tree, bool from_expander) {
 static constexpr size_t PARSE_HEAP_BASE = 4096;
 static constexpr size_t PARSE_HEAP_PER_CHAR = 64;
 
+static bool loading_startup = false;
+
+struct StartupLoadScope {
+    StartupLoadScope() { loading_startup = true; }
+    ~StartupLoadScope() { loading_startup = false; }
+};
+
+// A rejected line is dropped instead of throwing, so these paths have to feed core.startup_error
+// themselves. The out-of-memory branches above/below deliberately don't: building a message
+// allocates, and they exist to drop the line rather than reboot.
+static void report_parse_error(const std::string &message) {
+    echo("error: %s", message.c_str());
+    if (loading_startup) {
+        core_module->get_property("startup_error")->string_value = message;
+    }
+}
+
 void process_lizard(const char *line, bool trigger_keep_alive, bool from_expander) {
     InterpreterLock lock;
     if (trigger_keep_alive) {
@@ -349,21 +366,21 @@ void process_lizard(const char *line, bool trigger_keep_alive, bool from_expande
     struct source_range range;
     switch (owl_tree_get_error(tree.get(), &range)) {
     case ERROR_INVALID_FILE:
-        echo("error: invalid file");
+        report_parse_error("invalid file");
         break;
     case ERROR_INVALID_OPTIONS:
-        echo("error: invalid options");
+        report_parse_error("invalid options");
         break;
     case ERROR_INVALID_TOKEN:
-        echo("error: invalid token at range %zu %zu \"%s\"", range.start, range.end,
-             std::string(line, range.start, range.end - range.start).c_str());
+        report_parse_error("invalid token at range " + std::to_string(range.start) + " " + std::to_string(range.end) +
+                           " \"" + std::string(line, range.start, range.end - range.start) + "\"");
         break;
     case ERROR_UNEXPECTED_TOKEN:
-        echo("error: unexpected token at range %zu %zu \"%s\"", range.start, range.end,
-             std::string(line, range.start, range.end - range.start).c_str());
+        report_parse_error("unexpected token at range " + std::to_string(range.start) + " " + std::to_string(range.end) +
+                           " \"" + std::string(line, range.start, range.end - range.start) + "\"");
         break;
     case ERROR_MORE_INPUT_NEEDED:
-        echo("error: more input needed at range %zu %zu", range.start, range.end);
+        report_parse_error("more input needed at range " + std::to_string(range.start) + " " + std::to_string(range.end));
         break;
     case ERROR_ALLOCATION_FAILURE:
         echo("error: allocation failure while parsing");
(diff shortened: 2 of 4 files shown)


JensOgorek:
### Verified on an ESP32-S3 — works as claimed; good as a first increment once two boundary cases are fixed

I ran the PR (merged with current main, so the S3 gets past #248's I2C fix) on a bench ESP32-S3 — the runtime verification the PR body still lists as missing. Everything behaves as advertised: the boot error names module and type, `core.startup_error()` returns it after `Ready.`, a clean script reports empty, ordering behavior unchanged.

Two boundary cases, both reproduced on the same board, currently make the query misleading and should be fixed within this PR's scope: a factory-fresh (or NVS-erased) board reports a bogus storage error, and a syntax error reports **nothing** although it drops the *entire* script. Details below. Also consider exposing the value as a `StringVariable` property instead of a method — and it needs a docs entry either way.

The direction itself is decided over in #246 — this fits as the first increment.

<details>
<summary><b>Hardware verification — the PR scenario on an ESP32-S3</b></summary>

Bench S3 (QFN56 v0.2, 8 MB), PR merged with main @ `9e4b39a`, empty I²C bus, startup script `before = Output(1)` / `imu = Imu()` / `after = Output(2)`:

| check | result |
|---|---|
| boot error line | `error while loading startup script: module "imu" (Imu): imu setup failed: I2CError: Check your wiring.` |
| `core.startup_error()` | same message, retrievable post-boot |
| `before.on()` / `after.on()` | `before` exists, `after` unknown — behavior unchanged |
| clean script control | `startup_error: ` (empty) |

First runtime run on an S3 (the PR body had only compile-verified that target).

</details>

<details>
<summary><b>Boundary case 1 — fresh board reports a bogus startup error</b></summary>

First boot after `erase-flash`, no startup script ever saved:

```
>>> core.startup_error()
startup_error: could not peek storage storage.num_chunks (ESP_ERR_NVS_NOT_FOUND)
```

`Storage::init()` sits inside the same `try` in `app_main`, so its missing-key throw lands in the new state. Tooling using "non-empty ⇒ startup failed" flags every fresh or NVS-erased board (also after `expander.flash()`) as broken.

</details>

<details>
<summary><b>Boundary case 2 — syntax error drops the whole script, reports nothing</b></summary>

Startup script `before = Output(1)` / `garbage(((`:

```
error: unexpected token at range 29 30 "
Ready.
>>> core.startup_error()
startup_error:
>>> before.on()
error processing uart0: unknown module "before"
```

The parser rejects the whole tree — even the line *before* the typo never ran — yet `startup_error` stays empty, because the parse-error branches in `process_lizard` echo and return without throwing. Those branches (and the low-heap parse guard) need to feed the same state.

</details>

<details>
<summary><b>Surface suggestion — StringVariable property instead of a method</b></summary>

`this->properties["startup_error"] = std::make_shared<StringVariable>();` next to `core.debug`/`core.millis`, written as `core_module->properties.at("startup_error")->string_value = e.what();` in the `app_main` catch. Same mechanism as Can's `state` property; replaces both the public static and the hand-written `call()` branch, is readable as a bare `core.startup_error` expression, works in `core.output(...)`, and is opt-in broadcastable — which is what makes it forward-compatible with the error-code state discussed in #246. One caveat from checking the interpreter: string properties can't be used in `when`-rule conditions (the language has no string comparisons), so a boolean aggregate (e.g. `core.has_error`) would be the rule-facing signal later.

</details>

Reviewed and tested with Claude Fable 5.

evnchn:
Thanks for the run — both boundary cases are fixed and `startup_error` is now a read-only property (b116fad), plus one follow-up commit the board round surfaced (c59b1ff).

The board round (real ESP32-D0WD-V3) showed that a stored syntax error carries owl's raw newline, so the property arrived as **two** checksummed lines. c59b1ff sanitises `\n` and `\r` to spaces in the stored value — property reads are a single well-formed line, and the boot-time echo is untouched (diffed against `main`'s echo for the same stored script: byte-identical, checksums included).

<details>
<summary>Pre-existing issue found while verifying, not addressed here: <code>core.output</code> emits string values unescaped</summary>

Of the three read paths, two escape correctly and one does not:

| Read path | Wire output for a value containing `"` |
|---|---|
| bare expression `core.startup_error` | `"a\"b"@7d` — escaped |
| `core.print(core.startup_error)` | `"a\"b"@7d` — escaped |
| `core.output("core.startup_error")` | `core "a"b"@1a` — **unescaped** |

Feeding lizard's own broadcast line back into the parser fails: `error: invalid token at range 9 10 """`. This predates this PR — any string property with a quote hits it, and the parser itself accepts `\"` in string literals fine — so it affects the `output` path generally, not just `startup_error`. Reproduced on the same board. Happy to file it separately if you want it tracked.
</details>

PR #252, opened 2026-08-05
Add opt-in time synchronization to SerialBus

JensOgorek:
This adds `bus.enable_time_sync()` to the SerialBus module so the coordinator can estimate each peer's clock offset and publish it as a property `offset_<id>` (milliseconds). Motivation: timestamping sensor data that arrives over the bus — concretely wheel odometry from plexus peers, where the 10–35 ms delivery delay otherwise distorts pose estimation. With the offset, a peer timestamp `t_peer` maps to coordinator time as `t_peer - offset_<id>`, so the *measurement* time is known precisely no matter how late the data arrives.

The feature is strictly opt-in: without `enable_time_sync()` the wire protocol and polling behavior are byte-identical to today. A small preparatory commit guards the SerialBus frame reader against oversized garbage lines (e.g. cross-baud boot chatter), which previously terminated the firmware via an unhandled exception in the communication task.

**Hardware-validated** on a Robot Brain (core ESP32 as coordinator, P0 as bus peer, 300 s, 23,740 timestamped events): offset locks after ~1.6 s, wanders only **1.03 ms over 5 minutes** while tracking ~3 ppm crystal drift, and produced **zero acausal timestamps**. Roughly 8× tighter than the best host-side estimation on the same link (±4 ms).

<details>
<summary>Design and validation details</summary>

### Mechanism

- Each peer with sync enabled stamps its local `esp_timer` clock (µs) into its DONE response: `__DONE__<us>`. The poll cycle that runs anyway becomes a free NTP-style measurement channel at communication-task level (1 ms tick) — *below* the 10 ms main-loop raster that limits any script- or host-level approach.
- The coordinator subtracts the DONE frame's deterministic wire airtime from the receive time and feeds `peer_us - (t_receive - airtime)` into a windowed-maximum filter (`SYNC_WINDOW = 128`): the least delayed samples carry the least queueing latency, so the window maximum approximates the true offset from below the noise.
- After the first full window the offset locks; each subsequent window moves the published value halfway toward the new maximum — follows crystal drift (tens of µs per window) while damping outliers.
- Published from the main task in `step()` as `offset_<id>` (NumberVariable, ms); the comm-task/main-task handoff is guarded by a spinlock.

### Compatibility

- DONE parsing prefix-matches, so a plain `__DONE__` from an old peer is still accepted (no offset update). An old coordinator would mis-parse a *stamped* DONE — that is why stamping only happens after an explicit `enable_time_sync()` on the peer, controlled by the same deployment that upgrades the coordinator.
- `make_coordinator()` after `enable_time_sync()` (or vice versa) both work; per-peer state is (re)built on either call.

### Hardware validation (F31/rb67, 2026-08-05)

Setup: coordinator on the core ESP32 (runtime modules), peer on the P0 over the internal UART link at 115,200 baud; the peer pushes its `core.millis` every 10 ms via a rule, the coordinator outputs `(core.millis, peer_stamp, offset_2)` triples at 100 Hz to the host for analysis. 300 s, 23,740 events at 79 updates/s.

| Metric | Result |
|---|---|
| Lock time | ~1.6 s (128 samples at 79 polls/s) |
| Offset wander over 5 min | 1.03 ms total, tracking ~3 ppm drift |
| Corrected delivery latency | min 4.31 / median 12.31 / max 19.40 ms — never negative |
| Causality violations | 0 of 23,740 |
| Independent host-side cross-check | constant gap ≈ minimum delivery path (4–7 ms), stable across the run |

The corrected-latency distribution matches the sum of the two 10 ms main-loop quantizations plus the poll cycle — the estimator itself contributes well under 1 ms. Caveats: single peer, plain UART (no RS485 transceiver), multi-peer arbitration and concurrent OTB sessions untested. A full reproducible test protocol (scripts + raw JSON) exists outside the repo and can be shared.

### Known review points

- `std::strtoll` relies on a transitive `<cstdlib>` include (compiles in the validation build).
- The `sync_ready` handshake is a `volatile bool` set from the main task only; a reviewer may prefer an explicit atomic.

</details>


evnchn:

**TL;DR — one real memory-safety bug (rare trigger, cheap fix), three silent-wrong-value issues, two doc fixes. The design and your hardware numbers hold up.**

The blocking findings are about states your 300 s single-peer run never entered: **runtime reconfiguration, peer loss, and the pre-lock window.** I have also left two lower-priority protocol and diagnostic notes below.

1. **Use-after-free — `ensure_peer_clocks()` can reallocate `peer_clocks` while the comm task holds a `PeerClock &` into it** (`serial_bus.cpp:152` vs `:163`). I built a repro: forced interleave reproduces it 100%, free-running it fires unaided at roughly 1 in 300,000 reconfigure events. So: real UB, rare trigger, worth fixing. Your "Known review points" flags `sync_ready` — the thing to know is that **promoting it to an atomic would not fix this**, because the stale reference is acquired *after* the check.

2. **`offset_<id>` reads a plausible number when it is not one.** Three states publish `0.0` or a frozen last-known value, and a script cannot tell any of them from a genuinely measured zero: not-yet-converged, peer dropped by a later `make_coordinator()`, peer rebooted mid-window.

3. **Docs** — `t_peer - offset_<id>` needs its units stated (the property is ms, `esp_timer` is µs), and "about 2 seconds" is really 128 poll *rounds*, so it scales with peer count.

**What I would fix before merge:** the use-after-free and the `poll_index` clamp beside it — both are cheap and both corrupt state rather than merely degrade it. The rest can be follow-ups. For `offset_<id>`, I would at least choose the validity contract now, because it shapes how a host consumes the property.

**Scope:** static review against `ececaad`, plus one host repro. Nothing ran on an ESP32 — I am not disputing any of your measured numbers, only pointing at paths the single-peer run did not cover.

<details>
<summary><b>Use-after-free: reference invalidation across the main task / comm task boundary</b></summary>

`ensure_peer_clocks()` runs on the **main task** (via `call()` → `make_coordinator` / `enable_time_sync`) and can grow the vector:

```cpp
this->sync_ready = false; // comm task must not iterate while we rebuild
for (const uint8_t id : this->peer_ids) {
    ...
    this->peer_clocks.push_back(clock);   // serial_bus.cpp:152 — may reallocate
    ...
}
this->sync_ready = true;
```

`update_peer_offset()` runs on the **comm task** (`xTaskCreatePinnedToCore(... communication_loop ... core 1)` → `process_uart()` → `:280`), checks the flag exactly once, then holds a reference across the rest of the function:

```cpp
if (!this->sync_ready) { return; }        // serial_bus.cpp:160 — checked once
for (auto &clock : this->peer_clocks) {   // serial_bus.cpp:163 — reference into the vector
    ...
    clock.window_max_us = ...;            // :167 — after the main task may have reallocated
```

**Sequence:** comm task passes the `sync_ready` check and is preempted while holding `&clock`. Main task processes a `make_coordinator(1, 2, 3)` that introduces a new peer id, `push_back` reallocates, old block freed. Comm task resumes and reads/writes through the stale reference — corrupting heap/state and, on ESP32, typically taking the coordinator down with it.

The `volatile bool` → `std::atomic<bool>` change you flag in the PR body closes a visibility gap but **not this one**: the danger is the pointer the comm task already dereferenced, not the freshness of the flag. Some options:

- `peer_clocks.reserve(...)` is *not* sufficient on its own — a later reconfigure can still exceed it.
- Build the new peer-clock state off-lock, then publish it with one short critical section (swap, don't mutate in place).
- Or use a task-level mutex around both `ensure_peer_clocks()` and `update_peer_offset()`.

Whichever way: **don't extend `sync_mux` over the rebuild.** `portMUX_TYPE` critical sections should stay short and non-allocating, and the rebuild allocates on both lines — `peer_clocks.push_back(clock)` and `properties[...] = std::make_shared<...>()` (`:152-153`).

**Same trigger, second symptom** (`:270`): the new guard proves the vector is non-empty but not that the index is in range —

```cpp
if (!this->peer_ids.empty() && message.sender == this->peer_ids[this->poll_index]) {
```

`poll_index` is a persistent member. After `make_coordinator(1, 2, 3)` advances it to 2, a subsequent `make_coordinator(4)` leaves `peer_ids.size() == 1` with `poll_index == 2`. An in-flight poll can then read past the vector on **two** paths — the DONE match above, and the timeout warning at `:200`, which indexes `peer_ids[poll_index]` with no guard at all. That is UB rather than a predictable symptom, so I would not try to characterise what it does; resetting/clamping `poll_index` and clearing `is_polling` wherever `peer_ids` is assigned removes both.

</details>

<details>
<summary><b>Repro for the use-after-free (self-contained, no ESP-IDF needed)</b></summary>

| Mode | Result |
|---|---|
| `./uaf_repro forced` — deterministic interleave | ASan `heap-use-after-free`, **every run**. Freed by `push_back` (`:152`), read back through the stale `clock` reference (`:167`). |
| `./uaf_repro free 20000` — free-running, no forced preemption | **2 and 3 hits across two independent sweeps of 800,000 reconfigure events each** — order 1-in-300,000 on this host. |

**Proves:** the `sync_ready` handshake does not close the window. The comm task checks it once at `:160` and then holds `&clock` across `:167`; a `push_back` in between frees the block underneath it. That is a property of the control flow, so it holds regardless of platform — and it is why making `sync_ready` atomic would not help.

**Does not prove:** anything quantitative about the ESP32. Host clang + two pthreads on an M-series Mac is not FreeRTOS with the comm task pinned to core 1 at priority 5. The 1-in-300,000 figure is a host number and does not transfer — please don't quote it as the device's odds. The real device has a *genuinely parallel* window (two cores, not merely preemption), while the triggering event — a `make_coordinator` introducing a new peer id, mid-operation, with time sync on — is rarer than my loop makes it.

**Not a verbatim copy of your code.** The `peer_clocks` control flow and the `sync_ready` handshake are as written; deliberately different are `portMUX` → `std::mutex`, FreeRTOS tasks → `std::thread`, `time_sync_enabled` defaulting true so the path is reachable without an `enable_time_sync()` call, the dropped `properties["offset_<id>"]` publication at `:153` (a second allocation, not part of the dangling reference), and the `hook` that forces the deterministic interleave. All listed in the file header. If you think I mis-transcribed something, that is the first thing to check.

```bash
clang++ -std=c++17 -fsanitize=address -g uaf_repro.cpp -o uaf_repro
./uaf_repro forced        # fires every time
./uaf_repro free 20000    # ~1 hit per 300k events, so re-run it
```

```cpp
// Host repro for zauberzeug/lizard PR 252: reference invalidation across the
// main-task / comm-task boundary in SerialBus time sync.
//
// ensure_peer_clocks() and update_peer_offset() are transcribed from
// main/modules/serial_bus.cpp @ ececaad for the peer_clocks control flow that
// the race lives in. The sync_ready handshake and the loop/reference structure
// are kept exactly as written. Deliberate differences, none of which touch the
// race: portMUX -> std::mutex, FreeRTOS tasks -> std::thread, the
// properties["offset_<id>"] publication at :153 is dropped (it is a second
// allocation, not part of the dangling reference), time_sync_enabled defaults
// true here so the scenario is reachable without an enable_time_sync() call,
// and `hook` is added purely to force the interleave in the deterministic mode.
//
// Build: clang++ -std=c++17 -fsanitize=address -g uaf_repro.cpp -o uaf_repro
// Run:   ./uaf_repro forced     (deterministic interleave)
//        ./uaf_repro free       (free-running race, no forced sleeps)

#include <algorithm>
#include <atomic>
#include <cstdint>
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <mutex>
#include <string>
#include <thread>
#include <vector>

// ---- from serial_bus.h ------------------------------------------
struct PeerClock {
    uint8_t peer_id = 0;
    bool locked = false;
    int64_t window_max_us = INT64_MIN;
    size_t window_count = 0;
    int64_t offset_us = 0; // published value, guarded by sync_mux
};
static constexpr size_t SYNC_WINDOW = 128;

struct SerialBus {
    bool time_sync_enabled = true;
    volatile bool sync_ready = false; // peer_clocks fully built, safe for the comm task
    std::vector<PeerClock> peer_clocks;
    std::vector<uint8_t> peer_ids;
    std::mutex sync_mux; // stands in for portMUX_TYPE

    // ---- from serial_bus.cpp:141-156 ----------------------------
    void ensure_peer_clocks() {
        if (!this->time_sync_enabled || this->peer_ids.empty()) {
            return;
        }
        this->sync_ready = false; // comm task must not iterate while we rebuild
        for (const uint8_t id : this->peer_ids) {
            const bool known = std::any_of(this->peer_clocks.begin(), this->peer_clocks.end(),
                                           [id](const PeerClock &clock) { return clock.peer_id == id; });
            if (!known) {
                PeerClock clock;
                clock.peer_id = id;
                this->peer_clocks.push_back(clock); // :152 -- may reallocate
            }
        }
        this->sync_ready = true;
    }

    // ---- from serial_bus.cpp:159-183 ----------------------------
    // `hook` is NOT in the original; it only forces the interleave the claim
    // describes, at the point where the comm task is preempted.
    void update_peer_offset(const uint8_t sender, const int64_t raw_offset_us,
                            void (*hook)() = nullptr) {
        if (!this->sync_ready) { // :160 -- checked ONCE
            return;
        }
        for (auto &clock : this->peer_clocks) { // :163 -- reference into the vector
            if (clock.peer_id != sender) {
                continue;
            }
            if (hook) {
                hook(); // <-- "comm task preempted here"
            }
            clock.window_max_us = std::max(clock.window_max_us, raw_offset_us); // :167
            if (++clock.window_count >= SYNC_WINDOW) {
                {
                    // scoped to match portENTER/portEXIT_CRITICAL at :169/:178
                    std::lock_guard<std::mutex> lock(this->sync_mux);
                    if (!clock.locked) {
                        clock.offset_us = clock.window_max_us;
                        clock.locked = true;
                    } else {
                        clock.offset_us += (clock.window_max_us - clock.offset_us) / 2;
                    }
                }
                clock.window_max_us = INT64_MIN; // :179 -- outside the lock, as in the source
                clock.window_count = 0;
            }
            return;
        }
    }
};

static SerialBus bus;

// The "main task" reconfigure that the operator triggers with
// `bus.make_coordinator(1, 2, 3)` at runtime.
static void reconfigure(std::initializer_list<uint8_t> ids) {
    bus.peer_ids.assign(ids);
    bus.ensure_peer_clocks();
}

static int run_forced() {
    std::printf("[forced] steady state: make_coordinator(1, 2)\n");
    reconfigure({1, 2});
    std::printf("[forced] peer_clocks.size()=%zu capacity=%zu\n",
                bus.peer_clocks.size(), bus.peer_clocks.capacity());

    std::printf("[forced] comm task enters update_peer_offset(1, ...), passes the\n"
                "         sync_ready check, then is preempted while holding &clock\n");
    bus.update_peer_offset(1, 4242, [] {
        std::printf("[forced] main task runs make_coordinator(1, 2, 3) -> push_back reallocates\n");
        reconfigure({1, 2, 3});
        std::printf("[forced] peer_clocks.size()=%zu capacity=%zu (old block freed)\n",
                    bus.peer_clocks.size(), bus.peer_clocks.capacity());
        std::printf("[forced] comm task resumes and writes through the stale reference...\n");
    });

    std::printf("[forced] NO ASan report -- claim would be WRONG\n");
    return 1;
}

// Faithful free-running variant: each ROUND models exactly ONE runtime
// reconfigure event -- a coordinator already polling peers 1,2 when the
// operator issues make_coordinator(1, 2, 3). ensure_peer_clocks() stays
// add-only, as in the source; nothing clears or shrinks the vector.
// A fresh SerialBus per round, because the real vector only grows.
static int run_free(int rounds) {
    std::printf("[free] %d rounds; each = one make_coordinator(1,2,3) issued while\n"
                "       the comm task polls peer 1. No forced sleeps anywhere.\n", rounds);
    for (int r = 0; r < rounds; r++) {
        SerialBus local;
        local.peer_ids.assign({1, 2});
        local.ensure_peer_clocks();

        std::atomic<bool> go{false}, done{false};
        std::thread comm([&] {
            while (!go.load(std::memory_order_acquire)) {
            }
            while (!done.load(std::memory_order_relaxed)) {
                local.update_peer_offset(1, 4242); // the DONE-frame path
            }
        });
        go.store(true, std::memory_order_release);
        local.peer_ids.assign({1, 2, 3});
        local.ensure_peer_clocks(); // the single reallocating reconfigure
        done.store(true, std::memory_order_relaxed);
        comm.join();
    }
    std::printf("[free] %d reconfigure events completed with no ASan report\n", rounds);
    return 1;
}

int main(int argc, char **argv) {
    const std::string mode = argc > 1 ? argv[1] : "forced";
    const int rounds = argc > 2 ? std::atoi(argv[2]) : 2000;
    return mode == "free" ? run_free(rounds) : run_forced();
}
```

</details>

<details>
<summary><b><code>offset_&lt;id&gt;</code> is indistinguishable from a real zero in three states</b></summary>

`step()` only publishes locked clocks (`:72-79`), but `ensure_peer_clocks()` creates the property for every configured peer up front (`:153`) as a default-constructed `NumberVariable` — and `NumberVariable(const double value = 0.0)` (`main/compilation/variable.h:41`) means that default really is `0.0`, not an indeterminate value. So the property always exists and always reads as a plausible number.

| State | What a script reads | Why it is wrong |
|---|---|---|
| Not yet converged | `0.0` | `clock.locked` false → `step()` skips it, the default stands. With N peers this lasts 128×N poll rounds, not a fixed ~1.6 s. |
| Peer dropped by a later `make_coordinator()` | last locked value, forever | `ensure_peer_clocks()` is add-only — the `PeerClock` survives with `locked == true`, so `step()` republishes it every cycle while the dead peer's crystal drifts away from it. If that id is later reassigned to a different board, the new board inherits the old offset (`known` is true, the window never restarts). |
| Peer rebooted mid-window | a value up to hours off | Nothing resets `window_max_us` / `window_count` on the poll timeout that `communication_loop` already detects. The pre-reboot maximum wins the max filter for that window; the halving convergence then needs ~32 windows to walk it back. |

The concrete cost lands in exactly your motivating use case: a host that timestamps wheel odometry with `t_peer - offset_<id>` applies a zero (or stale) correction and silently mis-timestamps — the failure the feature exists to remove. Some way to say "no valid offset" fixes all three: NaN until locked, a companion `offset_<id>_valid`, or not creating the property until first lock.

Related and cheap: reset the peer's window on the existing `poll to %u timed out` path, and prune `peer_clocks` / the `offset_` properties for ids no longer in `peer_ids`.

</details>

<details>
<summary><b>Protocol: the prefix match now swallows user payloads</b></summary>

`:268` relaxes the DONE test from exact to prefix:

```cpp
// before: if (std::strcmp(message.payload, DONE_CMD) == 0) {
if (std::strncmp(message.payload, DONE_CMD, sizeof(DONE_CMD) - 1) == 0) {
```

With `strcmp`, only the literal token was reserved, so `bus.send(1, "__DONE__%d", status)` was a previously deliverable bus payload — `send` formats and enqueues arbitrary strings (`:96-113`). It now matches, hits `continue`, and never reaches `inbound_queue` — silently dropped. Worse, if that sender happens to equal `peer_ids[poll_index]`, `is_polling` clears early and the coordinator starts polling the next peer while the real one is still transmitting.

The check also runs unconditionally, with no `time_sync_enabled` guard, so it applies to installations that never opt in. Validating the remainder (all digits, or empty) before treating it as DONE keeps the compatibility win without reserving the whole prefix.

</details>

<details>
<summary><b>Diagnostics: the swallowed <code>read_line</code> exception hides a destructive flush</b></summary>

```cpp
try {
    len = this->serial->read_line(buffer, sizeof(buffer));   // :240
    continue; // oversized garbage line (e.g. cross-baud boot chatter); skip it   // :241
}
```

Agreed the unhandled exception had to go. The concern is that one path behind it — `Serial::read_line`'s `available() < pos` branch — calls `uart_flush_input()` *before* throwing (`main/modules/serial.cpp:136`), discarding **every other pending frame** in the ring buffer, not just the oversized line. That now happens with no console output at all, while every other failure path in `process_uart()` (checksum, parse, enqueue) still emits a `print_to_incoming_queue` warning. Symptom for a user: a peer that intermittently stops responding, and nothing in the log points at the flush.

A rate-limited warning in the catch keeps the fix and the diagnostic.

</details>

<details>
<summary><b>Docs, and one thing I am <i>not</i> asking you to change</b></summary>

- **Units in the conversion recipe** (`docs/module_reference.md:111`). `offset_<id>` is published in milliseconds (`offset_us / 1000.0`, `:79`) while `t_peer` as stamped into DONE is an `esp_timer` microsecond value. Read literally, `t_peer - offset_<id>` is off by 1000×. Your validation used `core.millis` on both sides, which is self-consistent — but the doc never says which unit `t_peer` is in, so a user reading the `esp_timer` stamp directly gets it wrong. Worth stating both units explicitly.
- **Convergence time** (`:110`). "Converges within about 2 seconds" is true for your one-peer bench; each peer gets one sample per full poll round, so 128 samples take 128×N rounds. "128 poll rounds" travels better.
- **Not asking for a change: the stamped-DONE-vs-old-coordinator constraint.** Your PR body already calls this out and accepts it, and the reasoning is sound. Mentioning it only because the failure mode if a deployment gets the order wrong is quiet and expensive — an unpatched coordinator using `strcmp` never matches `__DONE__1234567`, so every poll times out and the rejected payload echoes back as a parse error. One line in the docs saying "upgrade the coordinator first" would be cheap insurance.

</details>

<details>
<summary><b>Checked and dropped (so nobody re-walks them)</b></summary>

These came up and did not survive verification — recording them so the next reader doesn't repeat the work:

- **Divide-by-zero on `baud_rate`** in the airtime term (`:279`) — the division is unguarded, but a zero baud rate is not reachable through `create_serial` in practice.
- **`clock.locked` read outside the spinlock** in `step()` (`:73`) — on its own, a stale read costs at most one cycle of publication delay. Not worth a separate change; it falls out of whatever synchronization fix the use-after-free gets, along with `time_sync_enabled` and `sync_ready`.
- **`step()` rebuilding the property key string** and doing a map lookup every 10 ms cycle for a value that changes once per window — real, but not worth a change at this peer count.
- Style nits (`#include <climits>` redundant, C-style casts, `mutable` on `sync_mux`, member ordering, Properties-vs-Methods table order, unchecked `snprintf` return) — noted, not worth blocking on.

</details>

falkoschindler:
Great work — the continuous estimation via the poll cycle answers exactly the drift concern we had for multi-week missions (20 ppm = ~1.7 s/day, so one-shot sync was never going to fly). Two things we'd like to see before merge, one heads-up:

1. The `volatile bool sync_ready` handshake has a real TOCTOU on the dual-core ESP32: the comm task can pass the check while the main task reallocates `peer_clocks` (`push_back`) — use-after-free, even if only at (re)configuration time. An atomic with a proper handoff (or mutating under the existing spinlock) would close it.
2. DONE matching switched from exact to prefix match, so any bus payload *starting* with `__DONE__` is now swallowed as a DONE (and possibly parsed as a stamp). Worth rejecting trailing garbage explicitly or documenting the prefix as reserved.

Heads-up rather than blocker: while polling pauses (e.g. OTB flashing), the published offset freezes and silently drifts at crystal rate (~1.2 ms/min worst case at 20 ppm) — probably fine, but worth a sentence in the docs. And we're aware RS485 + multi-peer are untested; that's covered by the characterization task on our side (WA-51).

_Written and posted by Claude on behalf of Falko._

JensOgorek:
**Real-hardware validation update:** the time sync now has been validated on an actual plexus (ESP32-S3) connected to the F31 robot brain over half-duplex RS485 at 460800 baud — the earlier validation used a makeshift full-duplex setup. Headline results: the firmware offset estimate agrees with an independent host-side ground truth within **2.4 ms absolute** (typical ~1.2 ms, part of which is measurement-path asymmetry), 0 acausal stamps out of ~29k, and the wheels use case works end to end: a complete wm/ws/wd dataset arrives at the host at 100 Hz with ~10 ms total age and zero torn value pairs over 2 minutes.

<details>
<summary>Setup and detailed results</summary>

**Setup:** F31 robot brain (Jetson Orin), core ESP32 as coordinator (`Serial(13, 4, 460800, 2)` + `SerialBus(bus_serial, 1)` + `make_coordinator(2)` + `enable_time_sync()`), real plexus ESP32-S3 as peer (`Serial(39, 40, 460800, 1)` + `SerialBus(serial, 2)` + `enable_time_sync()`), auto-direction RS485 transceivers. Both nodes run this branch (ececaad); the S3 was flashed and configured over its CP2102N USB console, which also serves as an independent observation channel.

**Absolute ground truth (new):** the Jetson reads both consoles (core UART + plexus USB) against the same monotonic clock. Per 30 s bucket, a min-latency estimator per console yields each device's clock vs host; their difference is the true peer-vs-core offset, independent of the bus. Over 5 min at 460800 (29 999 samples per console, 0 bus warnings): firmware `offset_2` vs truth stayed within −0.9…−2.4 ms (systematic, drift tracked; the truth itself wandered ~1.4 ms and the estimate followed). Delivery latency after applying the offset: min 7.0 / median 7.4 / p95 9.0 ms — poll-cycle dominated on half-duplex, with 0 acausal stamps.

**Wheels end-to-end (2 min @ 460800):** peer pushes `wm`/`ws`/`wd` at 100 Hz via `bus.send(1, '!!wm=%d', core.millis)` etc.; the host sees 100.0 updates/s (11 999/12 000 ticks), freshness median 10.3 / max 11.4 ms, 0 skipped or stalled distance increments, 0 torn ws/wd pairs.

**Baud ladder on real RS485 (90 s per variant, dual-side error counting):** error-free at 115200/230400/460800 (push rates 59.9→99.9→84.5 updates/s for 1 value, 89.9→99.5→97.1 for 3 values; single-value pushes at 115200/460800 occasionally miss a poll, 3-value pushes keep the peer TX queue non-empty and reach ~100 Hz). Total breakdown at ≥921600 (coordinator: poll timeouts; peer: overflowing relay queues) — consistent with the ~460k limit of the RS485 transceivers, not a protocol property. A 460800 degradation seen earlier in the makeshift setup did not reproduce on real hardware. Round trip Jetson→core→peer→back: median 29.2 ms, p95 31.3 ms, 0/1200 lost, independent of baud rate.

**Soak (30 min):** combined load with the bus at 460800 and the Jetson↔core UART also at 460800: 179 925 checksum-valid output lines, 0 corrupt, 0 bus warnings on either side, freshness median 10.6 / max 15.3 ms, 0 acausal stamps.

Raw JSONs of all runs are archived internally (phaseB/C/D/E, F31 2026-08-07).

</details>

falkoschindler (changes requested):
Thanks @JensOgorek — the estimator design is sound and the hardware validation (both the original run and the RS485/plexus update) is convincing. The remaining issues all live in states the validation runs never entered: runtime reconfiguration, peer loss and reboot, and the pre-lock window. Requesting the following changes.

**1. Use-after-free when time sync is reconfigured at runtime.** `ensure_peer_clocks()` runs on the main task and can reallocate `peer_clocks` via `push_back` (main/modules/serial_bus.cpp (152)) while the comm task iterates the same vector in `update_peer_offset()` (163). The `sync_ready` flag is sampled once on entry (160), so it does not close the window — @evnchn's repro confirms this, and the atomic promotion suggested in the PR body would not help either, because the stale reference is acquired after the check. A shape that works: rebuild into a pre-allocated vector and swap it in under the existing `sync_mux`, take the same lock around the comm task's iteration (it is short and non-allocating), and drop `sync_ready` entirely. Rebuilding from `peer_ids` also fixes a related problem: a peer ID that is dropped and later re-used would currently inherit the previous board's clock state.

**2. `peer_ids` has the same reallocation race.** The main task assigns the vector in `make_coordinator()` (130) while the comm task indexes it (194, 200, 270). This predates the PR, but it is the same bug class and the same swap-under-lock pattern fixes it, so please handle it here rather than in a separate issue.

**3. `poll_index` can stay out of range after `peer_ids` shrinks.** The index is only wrapped modulo the vector size when the next poll is sent, so a `make_coordinator()` call with fewer peers leaves an out-of-range index that both the DONE match (270) and the timeout warning (200) dereference unchecked. Resetting `poll_index` and `is_polling` on reconfiguration and bounding the index at both use sites removes the whole class.

**4. The DONE prefix match swallows user payloads.** Any `bus.send()` payload starting with `__DONE__` is now consumed as a poll response and can even clear `is_polling` early — and the check runs on installations that never enable time sync. Treating a frame as DONE only when the suffix is empty or all digits keeps the compatibility win without reserving the whole prefix.

**5. Decide and implement the validity contract for `offset_<id>`.** The property reads `0.0` before the first window locks (indistinguishable from "clocks agree"), keeps a frozen value forever for peers dropped by a later `make_coordinator()`, and can be far off for a while after a peer reboots mid-window. A host that timestamps odometry with `t_peer - offset_<id>` silently mis-timestamps in all three states — exactly the failure the feature exists to remove. Our preference: publish NaN until locked and whenever the offset becomes invalid, which every consumer notices without polling a second property; a companion `offset_<id>_valid` would also work.

**6. Reset a peer's window on poll timeout.** Nothing resets the window on the timeout the comm task already detects; after a peer reboot the pre-reboot maximum wins the max filter and the halving convergence needs ~32 windows to walk it back. Resetting the window (and, with the NaN contract from point 5, dropping the lock) on timeout makes recovery a single window.

**7. Docs** (docs/module_reference.md (104-112)): state the units on both sides of `t_peer - offset_<id>` (the DONE stamp is in microseconds, the property in milliseconds), express convergence as 128 poll rounds rather than "about 2 seconds" (it scales with peer count), mention that the offset freezes and drifts while polling pauses (e.g. OTB updates), that the coordinator must be upgraded before enabling time sync on peers, and that `__DONE__<digits>` payloads are reserved for the protocol.

One thing already taken care of: I merged current main into the branch (a6e5d4e). #262 changed `Serial::read_line` to report oversized lines via negative return codes instead of exceptions, so the `try`/`catch` from the preparatory commit was resolved in favor of the new sentinel handling — the branch builds green against current main. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20analysis%20and%20wording%20by%20the%20model%2C%20building%20on%20evnchn%27s%20findings%3B%20scope%20and%20judgement%20calls%20from%20the%20author.&d=2026-08-14&a=claude-code "claude-fable-5: Review analysis and wording by the model, building on evnchn's findings; scope and judgement calls from the author.")

JensOgorek:
### All seven review points addressed in 1341fc8 — verified with ASan on the host and on F27's real RS485 bus

The rework follows the shapes you proposed: `peer_ids` and `peer_clocks` are now rebuilt off-lock and swapped in under `sync_mux` (with `sync_ready` gone and the comm task's iterations under the same short, non-allocating lock), `poll_index` is reset on reconfiguration and bounded at both use sites, `__DONE__` frames only count as DONE when the suffix is empty or all digits, `offset_<id>` publishes NaN until locked and whenever it turns invalid, a poll timeout resets the peer's window and lock, and the docs cover units, the 128-round convergence, the polling-pause drift, the upgrade order and the reserved prefix.

Verified today: @evnchn's ASan harness fires on the old control flow every forced run, the reworked shape survives the forced interleave plus 200,000 free-running reconfigure events clean; on F27 (classic ESP32 core + plexus ESP32-S3 over half-duplex RS485 @115200) all runtime states from the review ran through — details below. Ready for re-review.

<details><summary><b>Device test battery on F27 (core and plexus temporarily on this branch, robot restored afterwards)</b></summary>

| test | result |
|---|---|
| `__DONE__foo` sent peer→core | processed as a command (error echo visible in the round trip), sync untouched |
| ghost peers `make_coordinator(3, 9, 10)` | `offset_9`/`offset_10` stay NaN, polling to the real peer unaffected |
| shrink to `make_coordinator(3)` + 40 rapid reconfigures | no crash, offset stays locked and sane |
| peer reboot mid-run | offset drops to NaN after 1.8 s (timeout reset) and relocks 1.7 s later onto the new boot-time difference (+229.9 s → −47.8 s) — one window instead of ~32 windows of halving |
| soak: 1492 reconfigures under continuous ~50 Hz `bus.send` traffic, 60 s | core alive throughout, offset tracking |
| mixed versions | new coordinator with a legacy peer works (empty DONE suffix); a time-sync peer against an old coordinator spams command errors — confirming the documented "update the coordinator first" |

The boot-time NaN window closes in under 4 s at 115200 with one peer (128 rounds are quick), matching the reworded docs. Afterwards the robot was restored: plexus back on its rmd build with the arm startup (`c79e`), core back on its v0.12 fix build with the full config (`62c0`, 47.5 KB free heap), rosys reports matching checksums.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Rework%20and%20verification%20by%20the%20model%20directed%20by%20the%20author%3B%20the%20author%20reviewed%20the%20diff%20and%20supervised%20the%20robot%20tests.&d=2026-08-19&a=claude-code "claude-fable-5: Rework and verification by the model directed by the author; the author reviewed the diff and supervised the robot tests.")

falkoschindler:
Thanks @JensOgorek — re-review is done and all seven points check out against `1341fc8`. I verified the rework in code, not just the summary: the rebuild-and-swap keeps every allocation outside the critical sections, the comm task never holds a reference across a lock release, `poll_index` is bounded at both former dereference sites, the digit-suffix check restores `__DONE__foo` as a deliverable payload, and the NaN contract covers dropped and re-used peer IDs (safe because properties are resolved by name on every access). The F27 battery hits exactly the states the first validation missed — nice work.

Two small leftovers from the re-review are already fixed on the branch, so you don't need to act on them: `8327b9b` puts the remaining polling state (`is_polling`, `poll_start_millis`, `poll_index`) under `sync_mux` too — `make_coordinator()` took the lock on its side, but the comm task read and wrote them unlocked, so a reconfiguration racing a poll send could produce a spurious timeout that reset the wrong peer's clock — and caps the DONE stamp at 18 digits so a malformed but checksum-valid frame can't saturate `strtoll` and lock in a nonsense offset. `0ee3fd6` restructures the docs section into labelled bullets and states the offset direction in words (peer clock minus coordinator clock; a peer that booted later has a negative offset).

One heads-up on timing: we're briefly holding the merge. Falko and I sketched a two-way variant of the estimator that uses the existing POLL→DONE round trip (coordinator stamps T1/T4, peer reports T2/T3, so the peer's queue-draining time cancels out). It would remove the airtime model and the 128-round window, make convergence independent of peer count, and give a per-sample error bound instead of a max-filter hope. We want a quick falsification prototype on real hardware before deciding — measured with your harness, since that's the bar now. If it doesn't clearly win, this PR merges as is; either way your validation setup and results carry over. Expect a decision within days, not weeks.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Verification%20and%20wording%20by%20the%20model%3B%20the%20fixes%20on%20the%20branch%20were%20reviewed%20and%20the%20parking%20decision%20made%20by%20the%20author.&d=2026-08-20&a=claude-code "claude-fable-5: Verification and wording by the model; the fixes on the branch were reviewed and the parking decision made by the author.")

falkoschindler:
Closing as superseded by PR #277.

Issue #255, opened 2026-08-05
Expander flashing reliably fails on the first attempt with connect TIMEOUT


JensOgorek:
## Observed

```
I (…) replicator: Connecting..
E (…) replicator: Error while connecting: TIMEOUT
error processing uart0: could not flash expander "p0"
```

A second `p0.flash(true)` immediately afterwards succeeds and completes the replica (~48 s). The pattern (fail once, then succeed) occurred three times on Feldfreund F31 (rb67), Lizard 0439c76, 2026-08-05.

## Analysis

- `esp_loader_connect()` performs the enter-bootloader reset sequence **once**; its `trials = 4` only re-send the SYNC. If the single EN pulse does not reach the P0 cleanly, the application keeps running and SYNC goes unanswered.
- Before the loader runs, `Expander::flash()` calls `deinstall()`, which floats *all four* pins (RX/TX/EN/G0, pull-ups explicitly disabled). The loader then re-drives them — the first reset pulse is issued out of an undefined pin state. After a failed attempt the pins remain driven by the loader (defined levels), which plausibly is why the second attempt succeeds.
- `force` is *not* the differentiator — it only skips the remote strapping-pin checks; the pin sequencing is identical.
- Side finding: if the non-force strapping check throws, `flash()` aborts *before* `deinstall()` with `boot_pin` left LOW — the next natural P0 reset lands in the ROM bootloader unintentionally.

## Suggested fix

Include the bootloader-entry sequence in the retry loop (either in `ZZ::Replicator::connect()` or by looping `loader_port_enter_bootloader()` + sync), and restore `boot_pin` HIGH on every abort path of `Expander::flash()`.

## Open questions

- Clean-state reproduction: after a fresh `core.restart()` the first attempt failed once and succeeded once — not deterministic.
- The Robot Brain schematic should confirm the core→P0 NAND path polarity: espresso.py inverts levels for NAND boards on the Jetson side, the Expander C++ code does not. Empirically the pulse does reset the P0, but the polarity story deserves a schematic check.

_Full measurement context: https://github.com/zauberzeug/lizard/discussions/260_

JensOgorek:
### Not reproducible on a bench pair, and a TIMEOUT does not mean the reset pulse failed

[...]

PR #261, opened 2026-08-06
Stop hammering the UART driver after it is deinstalled

### LINKED ISSUE #257: Expander keeps stepping after disconnect(), spamming uart driver errors for 3 s
## Observed

After `p0.disconnect()`, the log fills with

```
E (…) uart: uart_pattern_get_pos(569): uart driver error
```

at ~100 Hz for about 3 seconds, plus occasional `uart_write_bytes(1455)` errors. Observed consistently on Feldfreund F31 (rb67), Lizard 0439c76, 2026-08-05.

## Analysis

`Expander::deinstall()` deletes the UART driver and floats the pins, but the module stays in `Global::modules` with `is_ready == true`. The main loop keeps calling `Expander::step()` → `ping()` (writes to the deleted driver) and `handle_messages()` → `Serial::has_buffered_lines()` → `uart_pattern_get_pos()` on the deinstalled driver. The spam stops only when the ping timeout (`ping_interval` 1 s + `ping_timeout` 2 s) flips `is_ready` to false.

## Suggested fix

- `Serial::has_buffered_lines()`: add the same `uart_is_driver_installed` guard that `available()` already has.
- `Expander::deinstall()`: set `is_ready = false` so `step()` stops pinging a dead link immediately.

The same missing guard bites much harder in the SerialBus module, where no timeout ends the loop — see #259.

_Sibling issue in the SerialBus module: #259. Full measurement context: https://github.com/zauberzeug/lizard/discussions/260_

### LINKED ISSUE #259: SerialBus communication task spins forever on a deleted serial driver
## Observed

If the underlying Serial driver is deleted while a SerialBus exists (e.g. another module deinstalls the UART, or an `Expander.disconnect()`-style teardown), the bus communication task spins on the dead driver and floods the log at task rate:

```
E (…) uart: uart_pattern_get_pos(569): uart driver error   (endless, ~1 kHz)
```

Unlike the Expander (which stops pinging after its 3 s ping timeout — see companion issue), the bus task has no such exit — the spam continues until reboot. Observed on Feldfreund F31 (rb67) during bus testing, 2026-08-06.

## Suggested fix

- `Serial::has_buffered_lines()`: add the `uart_is_driver_installed` guard that `available()` already has (fixes both this and the Expander variant).
- Longer term: SerialBus deserves an explicit `disconnect()`/teardown that stops the comm task before the driver goes away.

_Sibling issue in the Expander module: #257. Full measurement context: https://github.com/zauberzeug/lizard/discussions/260_

evnchn:
Fixes #257.

Partially addresses #259 and deliberately does **not** close it: the receive-side flood is gone, but a *coordinator* `SerialBus` still writes through an unguarded `write_checked_line()` (`serial_bus.cpp:131` → `:318`) at roughly the poll rate, and the explicit comm-task teardown that #259 itself calls "longer term" is not included.

**TL;DR** — Two lines. `Serial::has_buffered_lines()` gets the `uart_is_driver_installed()` guard that `Serial::available()` already has, and `Expander::deinstall()` marks the expander not-ready. This kills the log flood from both issues. Not verified on hardware — see Progress.

### Motivation

Issues #259 and #257 are the same root cause seen from two angles, and both were diagnosed down to the mechanism by Jens Ogorek. This PR implements his diagnosis.

`Serial::available()` guards its IDF call with `uart_is_driver_installed()` and returns 0 when the driver is gone. `Serial::has_buffered_lines()` did not — it was a bare `uart_pattern_get_pos()`. So after `p0.disconnect()`, `SerialBus::process_uart()` kept spinning that call at roughly 1 kHz against a deleted driver, and every call logged a UART driver error. That is #259.

Separately, `Expander::deinstall()` tore the driver down but left the module in `Global::modules` with `is_ready == true`, so `Expander::step()` kept calling `ping()` and `handle_messages()` at ~100 Hz until the ping timeout eventually flipped the flag. That is the write/read spam in #257.

### Implementation

`Serial::has_buffered_lines()` short-circuits on `uart_is_driver_installed()`, mirroring `Serial::available()` exactly. `Expander::deinstall()` clears `is_ready`, so `Expander::step()` stops touching the link immediately instead of waiting out the 1 s / 2 s ping timeout.

Nothing changes while a driver is installed: the added call is a bounds/null check, which is why the same pattern is already considered acceptable in `available()` on the same hot path.

<details>
<summary><b>Mechanism, with file:line</b></summary>

- `main/modules/serial.cpp:109-113` — `Serial::available()` already returns 0 when `!uart_is_driver_installed(this->uart_num)`.
- `main/modules/serial.cpp:118` — `has_buffered_lines()` was `return uart_pattern_get_pos(this->uart_num) != -1;`, no guard.
- `main/modules/serial_bus.cpp:165` — `SerialBus::process_uart()` loops on `has_buffered_lines()` from the pinned communication task, which has no timeout exit. This is the ~1 kHz spam path in #259.
- `main/modules/expander.cpp:72-74` — `Expander::step()` only pings and handles messages while `is_ready` is true.
- `main/modules/expander.cpp:231` — `Expander::deinstall()` now clears `is_ready` right after `serial->deinstall()`.

Bonus, and part of why the guard is worth having beyond the log noise: in IDF v5.3.1, `uart_pattern_get_pos()` (`uart.c:567`) checks only `p_uart_obj[uart_num]` and has **no** `uart_num < UART_NUM_MAX` bounds check, unlike its siblings. `uart_is_driver_installed()` (`uart.c:1806`) does bounds-check. Since `Serial(rx, tx, baud, uart_num)` never validates `uart_num`, a Lizard one-liner such as `Serial(26, 27, 115200, 99)` previously indexed a 3-element static array out of bounds. The short-circuit closes that path too.
</details>

<details>
<summary><b>Flash path: does clearing <code>is_ready</code> break it?</b> — no, checked every caller</summary>

`Expander::flash()` calls `deinstall()` (`expander.cpp:187`), flashes, saves startup, calls `reinitialize_after_flash()` (`:196`), and on success calls `restart()` (`:200`) — and `restart()` (`:128`) **already** sets `is_ready = false` and re-enters boot detection. So the post-flash state is unchanged by this PR. If flashing fails, the expander stays not-ready, which is the correct state.

Related question a reviewer will ask: **does `is_ready` ever come back after a `disconnect()`?** No. `check_boot_progress()` (`:80`) is the only writer of `is_ready = true` (`:93`), and it runs only from the constructor loop (`:60-66`). This is **not** introduced here — `Expander::restart()` (`:128`) and the ping timeout (`:113`) already reach the same terminal state on main today. Arguably worth its own issue; deliberately out of scope for this PR.
</details>

<details>
<summary><b>Build evidence — both targets</b></summary>

```
lizard.bin binary size 0x13ba40 bytes. Smallest app partition is 0x1f0000 bytes. 0xb45c0 bytes (36%) free.

lizard.bin binary size 0x139ca0 bytes. Smallest app partition is 0x1f0000 bytes. 0xb6360 bytes (37%) free.
```
Full clean builds (`rm -rf build sdkconfig sdkconfig.old` between targets — removing only `sdkconfig` is not enough, `build/CMakeCache.txt` retains the target and `idf.py` aborts with *"Target settings are not consistent"*).
</details>

<details>
<summary><b>Adversarial review — verbatim second-lineage verdict (OpenAI Codex, gpt-5.5)</b></summary>

```
Verdict: no blockers for the minimal diff as scoped, but it only partially fixes #259. #257 looks closed.

**BLOCKERS**

None found.

**SUGGESTIONS**

#257 is effectively closed. After `p0.disconnect()`, `Expander::deinstall()` now clears `is_ready` at expander.cpp:231, and `Expander::step()` only calls `ping()` / `handle_messages()` while `is_ready` is true at expander.cpp:71. That stops the 100 Hz write/read spam immediately after disconnect. The new `has_buffered_lines()` guard at serial.cpp:118 also prevents the `uart_pattern_get_pos()` error if a caller still checks after deinstall.

#259 is receive-side fixed, not task-teardown fixed. `SerialBus::process_uart()` now exits cleanly when the driver is already deleted because its loop condition goes through guarded `has_buffered_lines()` at serial_bus.cpp:163. However, the comm task still runs forever. If the bus is a coordinator, it can still call `send_message()` at serial_bus.cpp:131, which reaches unguarded `write_checked_line()` at serial_bus.cpp:318. Concrete residual: after driver deletion with `peer_ids` configured, polling attempts still produce `uart_write_bytes` errors roughly once per poll timeout, not at the former ~1 kHz receive-loop rate. That matches the issue's "longer term teardown" note, so I would not block this minimal fix on it.

The flash path is not broken by setting `is_ready = false`. `flash` calls `deinstall()` at expander.cpp:187, then flashes, saves startup, calls `reinitialize_after_flash()` at expander.cpp:196, and on success calls `restart()` at expander.cpp:200, which already sets `is_ready = false` and starts boot detection state. If flashing fails, the serial driver has still been reinitialized before throwing, but the expander remains not ready, which is consistent with failure.

Concurrency: the new `uart_is_driver_installed() && uart_pattern_get_pos()` has a TOCTOU window if another core deletes the UART driver between the two calls. That window already existed around all UART calls; this change narrows the common dead-driver path and does not make it worse. `read_line()` still has unguarded `uart_pattern_pop_pos()` at serial.cpp:132, but it is only reached after `has_buffered_lines()` returns true, so only the same pre-existing deletion race can hit it. The new `is_ready` write is read by the main loop's `Expander::step()`, not by the SerialBus task.

Exception/re-entrancy: `properties.at("is_ready")` should not throw on reachable `Expander::deinstall()` paths because `properties = Expander::get_defaults()` is populated in the constructor before `restart()` and before any `call()`. `deinstall()` is not a destructor path. Placement after `serial->deinstall()` is acceptable for current single-main-loop calls; moving it before would only reduce the theoretical interval where an already-deinstalled serial has `is_ready == true`.

Normal installed-driver behavior is unchanged. Cost-wise, the added `uart_is_driver_installed()` is just a bounds/null check and matches `available()`'s existing pattern, so ~1 kHz use in the bus task is not a concern.
```
</details>

<details>
<summary><b>Deliberately out of scope — the honest residual on #259</b></summary>

**This does not fully close issue 259, and I would rather say so than have you find it.** The receive-side flood is dead. But the communication task still runs forever, so a **coordinator** `SerialBus` still reaches unguarded `write_checked_line()` via `send_message()` (`serial_bus.cpp:131` → `:318`): after driver deletion with `peer_ids` configured, you should still see `uart_write_bytes` errors at roughly the poll rate (`POLL_TIMEOUT_MS = 250`, `serial_bus.cpp:19`) plus a "poll timed out" warning — instead of the previous ~1 kHz. A non-coordinator bus goes fully silent.

The proper fix is the explicit `SerialBus::disconnect()` / teardown that stops the comm task *before* the driver goes away — which #259 itself files under "longer term". That is a design change and belongs in its own discussion, not in this two-line guard.

Also unchanged, and correctly so: `Proxy::call` and `Proxy::write_property` (`proxy.cpp:24,28`) still write to the expander after `disconnect()`. They are user-command-triggered rather than loop-rate so they cannot flood, but they are the remaining source of the "occasional `uart_write_bytes(1455)` errors" in #257 if someone pokes a proxy post-disconnect.
</details>

### Progress

- [x] Tested on hardware. **Done — by @JensOgorek, not by me.** I have no Robot Brain + P0 expander here; my own ceiling was green builds for both targets plus static analysis and an adversarial second-lineage review. On F31/rb67, in an 8 s capture after `p0.disconnect()`: `uart_pattern_get_pos(569)` driver errors **223 → 0**, `uart_write_bytes(1455)` errors **2 → 0**, and the `expander p0 connection lost` warning gone — that last one being the direct read-out of the `is_ready` half of the diff rather than of traffic volume. [Full measurement and its one setup caveat](https://github.com/zauberzeug/lizard/pull/261#issuecomment-5203457638). I have deliberately not used a closing keyword so this does not auto-close on merge.

### DIFF
--- a/main/modules/expander.cpp
+++ b/main/modules/expander.cpp
@@ -230,6 +230,7 @@ void Expander::check_strapping_pins(const char *buffer) {
 
 void Expander::deinstall() {
     this->serial->deinstall();
+    this->properties.at("is_ready")->boolean_value = false;
     if (this->boot_pin != GPIO_NUM_NC && this->enable_pin != GPIO_NUM_NC) {
         gpio_reset_pin(this->boot_pin);
         gpio_reset_pin(this->enable_pin);
--- a/main/modules/serial.cpp
+++ b/main/modules/serial.cpp
@@ -116,7 +116,7 @@ int Serial::available() const {
 }
 
 bool Serial::has_buffered_lines() const {
-    return uart_pattern_get_pos(this->uart_num) != -1;
+    return uart_is_driver_installed(this->uart_num) && uart_pattern_get_pos(this->uart_num) != -1;
 }
 
 void Serial::flush() const {


JensOgorek:
### Hardware-verified on F31 (rb67)

The flood is gone. Measured on the real Robot Brain + P0, `p0.disconnect()` followed by an 8 s capture of UART0:

| | before | after |
|---|---:|---:|
| `uart_pattern_get_pos(569): uart driver error` | **223** | **0** |
| `uart_write_bytes(1455)` errors | 2 | 0 |
| `warning: expander p0 connection lost` | yes | no |

The 223 errors spanned 2982 ms at 74.8 Hz — matching `ping_interval` (1 s) + `ping_timeout` (2 s) exactly, which is the window the issue predicts.

The disappearance of `connection lost` is the interesting part: it is direct evidence for the second half of the diff. On main the flag only flips when the ping times out, which is why the warning fires; with `is_ready` cleared inside `deinstall()` the expander never gets pinged in the first place, so there is nothing to time out. Review-wise I also checked the flash path — `restart()` already sets `is_ready = false` at `expander.cpp:128`, so post-flash behaviour is unchanged, and on a *failed* flash the new line is a strict improvement.

Code review found nothing blocking.

<details>
<summary><b>Test setup, and one caveat on the comparison</b></summary>

- Robot: Feldfreund F31 / rb67, Robot Brain with NAND gates, core on `/dev/ttyTHS1`, rosys stopped for the duration, startup checksum `4a6c` verified before and after.
- "before" was measured on the firmware the robot was running (lizard-zz, Lizard `792a749`) with the production startup script, i.e. `serial = Serial(26, 27, 115200, 1)` + `p0 = Expander(serial, 25, 14)` plus four proxies.
- "after" was measured on `origin/main` + this PR + #262 + #263, with `serial`/`p0` created at runtime with identical arguments.
- **The two setups are therefore not byte-identical**, and I would rather name that than present a clean A/B I did not run: upstream Lizard cannot execute F31's production script (it contains private `InnotronicDriveMotor` modules), so the proxies were absent in the "after" run.
- That said, the mechanism is setup-independent: the spam comes from `Expander::step()` polling `has_buffered_lines()` on a deleted driver, which happens on every tick regardless of how many proxies exist. And the `connection lost` signal above is a direct read-out of the `is_ready` change, not of traffic volume.

</details>

<details>
<summary><b>Two side observations from the same session, neither caused by this PR</b></summary>

**`is_ready` never returns to true after `p0.flash()`.** Observed directly: after a flash attempt, `core.print(p0.is_ready)` returned `false` and stayed false; only a `core.restart()` brought it back to `true`. This is the terminal-state quirk you already describe in the PR body ("does `is_ready` ever come back after a `disconnect()`? No") — it reproduces on main today for the flash path too. It is pre-existing and out of scope here, but it is real and probably deserves its own issue.

**`core.get_pin_status()` cannot read back a driven output pin.** Querying the core's own boot pin returns `GPIO_Status[25]| Level: 0| InputEn: 0| OutputEn: 1| ...` — with the input buffer disabled, the reported `Level` does not reflect the driven output level. Noting it because it is the obvious way one would try to verify boot-pin state (e.g. for #255/#263), and it does not work.

</details>

evnchn:

**TL;DR — thank you, hardware box checked. The disappearance of `connection lost` is better evidence than the error count, and I hadn't thought to predict it.**

223 → 0 is the part I did predict. The missing `expander p0 connection lost` is the part I did not, and you're right that it's the stronger signal: the error count only exercises the `has_buffered_lines()` guard, whereas the absent warning is a direct read-out of the `is_ready` line — no ping, so nothing to time out. PR body updated to cite both, with your setup caveat named rather than smoothed over.

On that caveat: agreed it doesn't undermine the result. The flood path is `Expander::step()` → `has_buffered_lines()` once per tick, which is independent of proxy count; the absent proxies could only have changed the volume of the pre-existing `uart_write_bytes(1455)` noise, not whether the `uart_pattern_get_pos` flood exists.

<details>
<summary><b>Your two side observations — where I think each belongs</b></summary>

**`is_ready` never returns to true after `p0.flash()`.** Agreed it deserves its own issue, and I've filed it as #266 with your observation and reproduction attributed to you — correct me there if I've put words in your mouth.

My PR body claimed this only for `disconnect()`; you've shown it reproduces on the flash path, which is the more surprising case, because `flash()` calls `restart()` at the end of a *successful* flash specifically to re-enter boot detection. Tracing it: `is_ready = true` has exactly one writer (`expander.cpp:93`, inside `check_boot_progress()`), and `check_boot_progress()` has exactly one caller — the constructor's `while` loop at `:60-66`. `step()` never calls it. So `restart()` sends `core.restart()`, resets `boot_start_time`, and sets up for a boot detection that then never runs; the P0 comes back and sends `Ready.` with nobody listening. One-way flag, not a missed assignment.

**`core.get_pin_status()` can't read back a driven output.** Noted and already useful: it's why #263 can't be verified the obvious way, and I've said so in that PR's body rather than leaving the next person to rediscover it.

</details>

Nothing in your review changed the diff, so this is still the same two lines with a green `pre-commit`.

JensOgorek (approved):
Approved — code review clean, fix hardware-verified on two platforms (F31/rb67: 223→0 driver errors after disconnect; bench ESP32 pair: SerialBus spin 1556→0). Details in my comment above.

JensOgorek:
Status check: approved, CI green, hardware-verified on F31 — nothing left from my side. @falkoschindler since you usually do the merging: anything still open here, or can this go in?

falkoschindler (approved):

Issue #253, opened 2026-08-05
Expander handshake cannot survive cross-baud boot chatter


JensOgorek:
## Context

When the expander's UART1 rate differs from 115,200, every P0 boot has a cross-baud phase: the ROM/second-stage bootloader always outputs at 115,200, and the application only switches to its configured rate in `app_main`. The core, listening at the configured rate, receives that phase as garbage. This happens on **every** `Expander()` construction whenever the link runs above 115,200 — i.e. it makes any baud rate migration fragile by design.

Found and verified on Feldfreund F31 (rb67), Lizard 0439c76, 2026-08-05.

## Observed — three separate failure modes

1. **`Ready.` is silently dropped.** The garbage contains almost no `\n`, so pattern detection never fires and `check_boot_progress()` never consumes; the 2 KB RX buffer fills with garbage, and when the readable `Ready.` finally arrives it is discarded. The constructor times out: `warning: expander ex connection timed out.`
2. **An oversized garbage line kills the module.** Once a `\n` does appear after >1024 garbage bytes, `Serial::read_line` throws (`buffer too small. discarded line.`). Neither `check_boot_progress()` nor `handle_messages()` catches it, so the exception propagates out of the `Expander` constructor and the module is never created: `error processing uart0: buffer too small. discarded line.`
3. **The 5 s boot timeout is tight.** A P0 boot takes ~3 s (ROM + 1.5 s log-drain delay + script); anything extra pushes `Ready.` past the window.

## Fix (hardware-verified)

Three small changes, together sufficient — with them, runtime `Serial(26,27,460800,1)` + `Expander(...)` pairs connected reliably at 460,800 and 921,600 on real hardware:

- `serial.cpp`: raise `RX_BUF_SIZE` (the test used 16 KB; even 8 KB covers the ~2 KB boot chatter after cross-baud inflation),
- `expander.cpp`: wrap both `read_line` call sites in try/catch and skip the oversized line,
- `expander.cpp`: raise the `boot_timeout` default (the test used 15 s), or make the constructor non-fatal on timeout with a later retry.

## Discussion

A more fundamental option: a **self-healing baud handshake** — the Expander tries the configured rate, falls back to 115,200 on timeout, then issues `core.set_baudrate(<target>)` + restart itself. Changing `serial = Serial(26, 27, 460800, 1)` in the startup script would then be the *only* step of a migration, and mismatched core/P0 rates could never strand the P0.

[...]

Issue #254, opened 2026-08-05
p0.flash() silently replaces the expander's NVS with a copy of the core's


JensOgorek:
## Mechanism

`flashReplica` maps the running partition's *physical* flash pages starting at page 0 and writes them to the target beginning at **address 0** (`esp_loader_flash_start(0, usedSize, ...)` with `usedSize = 0x1F0000`). The copied range 0x0–0x1EFFFF includes the NVS partition at 0x9000. So every successful `p0.flash()` replaces the P0's entire NVS with a byte copy of the core's — including `uart/baudrate` and any `bus_backup` namespace.

`Expander::flash()` already works around one consequence: it calls `Storage::clear_nvs()` before and `Storage::save_startup()` after the copy, so the P0 does not inherit the core's startup script. That workaround covers *only* the startup key.

Observed on Feldfreund F31 (rb67), Lizard 0439c76, 2026-08-05.

## Consequences observed

- A rate set via `p0.set_baudrate(X)` is silently reverted by the next `p0.flash()` (to whatever the core's NVS says — typically unset → 115,200). This cost several very confusing hours until the mechanism was clear.
- Conversely it is exploitable as a *feature*: plant the target rate in the core's NVS (`core.set_baudrate(X)` **without** restarting), then `p0.flash(true)` — the P0 deterministically boots at X afterwards. This is how the UART1 measurements were done.
- Potential contamination channel: if the core ever stored a SerialBus backup, the P0 would inherit and replay it at every boot (`bus_backup::restore_if_needed` removes all Serial modules when it fires).

## Suggested fix / discussion

Either:
- start the replica write at 0x20000 (app only) so the P0's NVS survives, and drop the clear_nvs/save_startup dance; or
- keep the full copy but document it loudly and re-apply the P0's own `uart/baudrate` (and clear `bus_backup`) after flashing.

The first option seems cleaner but changes replica semantics (a factory-fresh P0 would keep whatever NVS it has) — maintainer decision.

## Open questions

The analysis is code-based and consistent with all observations; a direct verification (read back the P0's NVS region before/after a flash) would make it airtight.

_Full measurement context: https://github.com/zauberzeug/lizard/discussions/260_

PR #262, opened 2026-08-06
Guard expander read_line against oversized lines

evnchn:
**TL;DR** — Catch the oversized-line exception at both `Expander` `read_line` call sites so the module still gets created, and fix the discard path so it consumes the terminating newline. This covers **one of the three** failure modes in the issue; the two constant bumps are left as a recommendation for you, not a patch.

### Motivation

Issue #253 — the Expander handshake cannot survive cross-baud boot chatter. The P0 bootloader always talks at 115,200 while the core listens at the configured rate, so every boot above 115,200 delivers a burst of garbage. Jens Ogorek diagnosed this down to three distinct failure modes; this PR implements only the second, which is the one that is unambiguously a bug rather than a tuning decision.

`Serial::read_line` throws `std::runtime_error` when a line exceeds the 1 KB buffer. Neither `Expander` call site caught it, so on a garbage burst the exception propagated out of the **`Expander` constructor** and the module was never created at all.

### Implementation

Both `read_line` call sites — `check_boot_progress()` and `handle_messages()` — now catch `std::runtime_error`, report it via `echo()` in the same shape as the rest of the file, and skip that line. The skip is **not silent**; a discarded line is observable on the console.

The discard path in `Serial::read_line` also needed a one-character fix. It discarded `pos` bytes while the success path reads `pos + 1`, leaving the terminating `\n` in the RX ring — which then prepends itself to the next line. In this issue's exact scenario that next line is `Ready.` itself, so the fix would have converted "module never created" into "module created but never ready". `i < pos` → `i <= pos`.

<details>
<summary><b>The stream-desync trap, and why the diff is two files not one</b></summary>

Concrete failure without the one-character fix: cross-baud garbage of >1024 bytes containing no newline, terminated by the leading `\n` of the P0's `printf("\nReady.\n")` (`main/main.cpp:484`). `read_line` pops that pattern position, discards `pos` bytes, throws; the new catch continues. The next `read_line` then returns `"\nReady.\n"`. `check()` strips only trailing whitespace, so `buffer == "\nReady."` and `strcmp("Ready.", buffer)` at `main/modules/expander.cpp:98` fails. `Ready.` is sent exactly once, so the constructor spins to `boot_timeout` and returns not-ready.

This was verified on the host against the real `strip()` / `check()` from `main/utils/uart.cpp` (ASan/UBSan clean): a leading `\n` yields `checksum_ok == 0` on a checksummed line and `buffer == "\nReady."` on a bare one — `Ready.` is lost on both wire formats.

The alternative fix — a compensating `read()` inside the expander's catch block — was considered and rejected: it would be right for the "discarded line" throw but would eat a byte of the next good line on the "flushed serial" throw, where `uart_flush_input` has already cleared the ring.
</details>

<details>
<summary><b>Behaviour change worth naming, so "why touch both sites?" is pre-answered</b></summary>

Before this PR, `main.cpp:431 run_step()` already caught the escaping `runtime_error` from `Expander::step()`, so steady state was survivable and exactly one line was processed per 10 ms tick. After this PR the loop drains the whole RX backlog within a tick — which actively helps failure mode 1 (buffer filling with garbage).

So the two halves are not equal: the `check_boot_progress()` half closes the only strictly-fatal path (constructor throws, module never exists); the `handle_messages()` half is a refinement that also happens to drain faster.

One deliberate asymmetry: on the catch path, `handle_messages` does **not** update `last_message_millis` and does **not** clear `ping_pending`, whereas the checksum-mismatch path a few lines below does both. An oversized line therefore counts as "no traffic" and will eventually trip the ping timeout → connection lost. That reads to me as the correct interpretation of an untrustworthy link, but it is a deliberate-looking inconsistency, so: it is deliberate. Say the word if you want it the other way.
</details>

<details>
<summary><b>Build evidence — both targets, post-fix</b></summary>

```
lizard.bin binary size 0x13bb10 bytes. Smallest app partition is 0x1f0000 bytes. 0xb44f0 bytes (36%) free.

lizard.bin binary size 0x139d70 bytes. Smallest app partition is 0x1f0000 bytes. 0xb6290 bytes (37%) free.
```
Full clean builds of both targets (1161 / 1204 ninja targets), each recompiling `modules/serial.cpp.obj` and `modules/expander.cpp.obj`.
</details>

<details>
<summary><b>Adversarial review — verbatim second-lineage verdict (OpenAI Codex, gpt-5.5), which caught the newline bug</b></summary>

This review was run against the **first** version of the diff, which had the catch blocks but not the `i <= pos` fix. It found a real defect and rated it do-not-ship. The defect is fixed in what you are reading; the verdict is reproduced unedited because the reasoning is the best documentation of why the second file is in the diff.

```
BLOCKERS:
1. `main/modules/expander.cpp:85` / `main/modules/serial.cpp:142`: oversized-line discard does not consume the trailing `\n`, so the next real line can be swallowed with a leading newline. Concrete scenario: UART contains `1024+ bytes garbage + "\nReady.\n"`. First `read_line()` pops the garbage newline position, reads only `pos` bytes, leaves that newline in RX, throws, and the new catch continues. The next `read_line()` for `Ready.\n` reads `"\nReady.\n"` into the static buffer. `check_boot_progress()` then compares `strcmp("Ready.", buffer)`, which fails because `buffer` is `"\nReady."`. If `Ready.` is only sent once, the constructor waits until `boot_timeout` and returns with `is_ready=false`. That does not close the reported boot-handshake failure; it changes "module not created" into "module created but never ready" for the same cross-baud oversized garbage case.

SUGGESTIONS:
- In `Serial::read_line()`, discard `pos + 1` bytes on the oversized discard path if available, so the line terminator is actually skipped.
- The catch blocks themselves do not appear to livelock: both throw paths pop at least one pattern entry, and the flush-drain loop's `> 0` condition still pops a `0` entry before stopping.
- No stale-buffer use found on the catch path; `continue` prevents `len`/`buffer` from being consumed, and `len = 0` is not reachable-uninitialized.
- `SerialBus::process_uart()` has the same uncaught `read_line()` failure mode in its FreeRTOS task, but this diff leaves that unchanged; I would not block this minimal expander fix on it.

VERDICT: do-not-ship as-is. The diff is narrowly aimed and most of the exception-safety mechanics are sound, but the discard path does not actually skip the oversized line's newline, so it can corrupt and miss the very `Ready.` line this fix needs to preserve. Confidence: high.
```
</details>

<details>
<summary><b>Deliberately out of scope, including Jens's other two proposals</b></summary>

**This PR does not close issue 253.** Two of the three failure modes are untouched, on purpose:

- **`RX_BUF_SIZE` (`main/modules/serial.cpp:8`, currently 2048).** Jens tested 16 KB and reports 8 KB suffices. I did not bump it, because this constant is used for **every** `uart_driver_install`, i.e. every `Serial` instance on the device — a 2 KB → 8 KB bump is 4× RAM multiplied by the number of UARTs, on a chip where DRAM is the scarce resource. That is your call, not a drive-by constant edit in a bugfix PR. Recommending 8 KB per Jens's measurement.
- **`boot_timeout` (`main/modules/expander.cpp:28`, currently 5.0).** Jens tested 15 s. Also your call; it is a default that users can already override.
- **The self-healing baud handshake** (fall back to 115,200, `core.set_baudrate`, self-restart) from the issue's Discussion section is a feature proposal, not a bug fix.

Known adjacent work, not touched here: `SerialBus::process_uart()` (`main/modules/serial_bus.cpp:166`) has the identical uncaught `read_line` failure mode, and it runs inside the core-1 pinned task, where an escaping exception is worse than in the expander case. Both reviewers agreed it should not block this minimal fix.

Considered and declined: making the guard at `main/modules/serial.cpp:134` `available() <= pos` rather than `< pos`, now that the discard path consumes `pos + 1`. The pattern interrupt only fires once the `\n` is resident, so `available() >= pos + 1` holds in practice, and the extra `read()` returns -1 harmlessly otherwise. Not worth enlarging the diff.
</details>

### Progress

- [x] Tested on hardware, **against the oversized-line mechanism — not against the cross-baud sequence of #253.** Done by @JensOgorek, not by me; my own ceiling was green builds for both targets, a host-level ASan/UBSan reproduction of the newline desync against the real `strip()`/`check()`, and adversarial review. On F31/rb67 he triggered a >1024-byte reply from the P0 and confirmed both halves: the catch now reports locally (`p0: error while handling messages: …`) instead of escaping to `run_step()` (`error in module "p0": …`), and the `i <= pos` fix takes the follow-on `Checksum mismatch` from **1 per trigger to 0**, with the next line intact **3/3**. [Measurement and what he could not verify](https://github.com/zauberzeug/lizard/pull/262#issuecomment-5203457892). No closing keyword used, so this will not auto-close issue 253, which it should not.

### DIFF
--- a/main/modules/expander.cpp
+++ b/main/modules/expander.cpp
@@ -80,11 +80,9 @@ void Expander::step() {
 void Expander::check_boot_progress() {
     static char buffer[1024];
     while (this->serial->has_buffered_lines()) {
-        int len = 0;
-        try {
-            len = this->serial->read_line(buffer, sizeof(buffer));
-        } catch (const std::runtime_error &e) {
-            echo("%s: error while checking boot progress: %s", this->name.c_str(), e.what());
+        const int len = this->serial->read_line(buffer, sizeof(buffer));
+        if (len < 0) {
+            echo("%s: error while checking boot progress: %s", this->name.c_str(), Serial::read_line_error(len));
             continue;
         }
         bool checksum_ok = true;
@@ -137,11 +135,9 @@ void Expander::restart() {
 void Expander::handle_messages(bool check_for_strapping_pins) {
     static char buffer[1024];
     while (this->serial->has_buffered_lines()) {
-        int len = 0;
-        try {
-            len = this->serial->read_line(buffer, sizeof(buffer));
-        } catch (const std::runtime_error &e) {
-            echo("%s: error while handling messages: %s", this->name.c_str(), e.what());
+        int len = this->serial->read_line(buffer, sizeof(buffer));
+        if (len < 0) {
+            echo("%s: error while handling messages: %s", this->name.c_str(), Serial::read_line_error(len));
             continue;
         }
         bool checksum_ok = true;
--- a/main/modules/expander.cpp
+++ b/main/modules/expander.cpp
@@ -80,7 +80,13 @@ void Expander::step() {
 void Expander::check_boot_progress() {
     static char buffer[1024];
     while (this->serial->has_buffered_lines()) {
-        const int len = this->serial->read_line(buffer, sizeof(buffer));
+        int len = 0;
+        try {
+            len = this->serial->read_line(buffer, sizeof(buffer));
+        } catch (const std::runtime_error &e) {
+            echo("%s: error while checking boot progress: %s", this->name.c_str(), e.what());
+            continue;
+        }
         bool checksum_ok = true;
         check(buffer, len, &checksum_ok);
         if (!checksum_ok) {
@@ -131,7 +137,13 @@ void Expander::restart() {
 void Expander::handle_messages(bool check_for_strapping_pins) {
     static char buffer[1024];
     while (this->serial->has_buffered_lines()) {
-        int len = this->serial->read_line(buffer, sizeof(buffer));
+        int len = 0;
+        try {
+            len = this->serial->read_line(buffer, sizeof(buffer));
+        } catch (const std::runtime_error &e) {
+            echo("%s: error while handling messages: %s", this->name.c_str(), e.what());
+            continue;
+        }
         bool checksum_ok = true;
         len = check(buffer, len, &checksum_ok);
         if (!checksum_ok) {
--- a/main/modules/serial.cpp
+++ b/main/modules/serial.cpp
@@ -132,18 +132,23 @@ int Serial::read(uint32_t timeout) const {
 int Serial::read_line(char *buffer, size_t buffer_len) const {
     int pos = uart_pattern_pop_pos(this->uart_num);
     if (pos >= static_cast<int>(buffer_len)) {
-        if (this->available() < pos) {
+        if (this->available() <= pos) {
             uart_flush_input(this->uart_num);
             while (uart_pattern_pop_pos(this->uart_num) > 0)
                 ;
-            throw std::runtime_error("buffer too small, but cannot discard line. flushed serial.");
+            return LINE_FLUSHED;
         }
 
         for (int i = 0; i <= pos; i++)
             this->read();
-        throw std::runtime_error("buffer too small. discarded line.");
+        return LINE_DISCARDED;
     }
     return pos >= 0 ? uart_read_bytes(this->uart_num, (uint8_t *)buffer, pos + 1, 0) : 0;
+}
+
+const char *Serial::read_line_error(const int result) {
+    return result == LINE_FLUSHED ? "buffer too small, but cannot discard line. flushed serial."
+                                  : "buffer too small. discarded line.";
 }
 
 void Serial::clear() const {
--- a/main/modules/serial.cpp
+++ b/main/modules/serial.cpp
@@ -139,7 +139,7 @@ int Serial::read_line(char *buffer, size_t buffer_len) const {
             throw std::runtime_error("buffer too small, but cannot discard line. flushed serial.");
         }
 
-        for (int i = 0; i < pos; i++)
+        for (int i = 0; i <= pos; i++)
             this->read();
         throw std::runtime_error("buffer too small. discarded line.");
     }
--- a/main/modules/serial.h
+++ b/main/modules/serial.h
@@ -29,7 +29,10 @@ public:
     int available() const;
     bool has_buffered_lines() const;
     int read(const uint32_t timeout = 0) const;
+    static constexpr int LINE_DISCARDED = -1;
+    static constexpr int LINE_FLUSHED = -2;
     int read_line(char *buffer, size_t buffer_len) const;
+    static const char *read_line_error(const int result);
     size_t write(const uint8_t byte) const;
     void write_checked_line(const char *message) const;
     void write_checked_line(const char *message, const int length) const;
--- a/main/modules/serial_bus.cpp
+++ b/main/modules/serial_bus.cpp
@@ -164,6 +164,10 @@ void SerialBus::process_uart() {
     static char buffer[FRAME_BUFFER_SIZE];
     while (this->serial->has_buffered_lines()) {
         const int len = this->serial->read_line(buffer, sizeof(buffer));
+        if (len < 0) {
+            this->print_to_incoming_queue("warning: serial bus %s error while processing uart: %s", this->name.c_str(), Serial::read_line_error(len));
+            continue;
+        }
         if (bool ok; (check(buffer, len, &ok), !ok)) {
             this->print_to_incoming_queue("warning: serial bus %s checksum mismatch: %s", this->name.c_str(), buffer);
             continue;


JensOgorek:
### Hardware-verified on F31 (rb67) — including the newline fix

Both halves of the diff are confirmed on real hardware, with the same trigger before and after.

**Trigger.** I needed one line longer than 1024 bytes coming *from* the P0 while my own command stayed under the core's 1024-byte input buffer. The P0's own error echo is the amplifier: `error processing uart0: unknown variable "…"` adds 45 bytes of overhead, so `p0.run("<1000 a's>")` (1013 bytes on the wire) provokes a 1045-byte reply. Three triggers per run, each followed by a `p0.info()` so the *next* line is something I can check.

| | before (`792a749`) | after (this PR + #261 + #263) |
|---|---|---|
| reported as | `error in module "p0": buffer too small. discarded line.` | `p0: error while handling messages: buffer too small. discarded line.` |
| `Checksum mismatch` right after | **1 per trigger** | **0** |
| following `p0.info()` reply | corrupted | 3/3 received intact |

The middle row is the `i <= pos` fix, caught in the act. On main the discarded line's `\n` stays in the ring and prepends itself to the next line, which then fails its checksum — exactly the desync your Codex reviewer predicted, and it reproduces on every single trigger. With the fix the very next line arrives clean, three times out of three.

The left column also confirms the catch matters: before, the exception escapes all the way to `run_step()` in `main.cpp` (hence `error in module "p0"`); after, it is handled locally and names the expander.

Code review found nothing blocking. I specifically checked the guard you flagged as considered-and-declined at `serial.cpp:135` (`available() < pos` vs. the loop now reading `pos + 1`) and agree it is harmless — the pattern interrupt only fires once the `\n` is resident, so `available() >= pos + 1` holds, and `read()` is non-blocking and returns -1 otherwise. Same for the deliberate asymmetry in the catch block not refreshing `last_message_millis`; leaving it as "no traffic" reads correct to me.

<details>
<summary><b>What I could not verify: the cross-baud scenario of #253</b></summary>

I tried to reproduce the actual boot-chatter case by disconnecting `p0` and constructing a runtime `Serial(26, 27, <rate>, 1)` + `Expander(…)` pair at a rate the P0 does not speak. **Four attempts (one at 460,800, three at 921,600) all produced failure mode 1 only** — `warning: expander eN connection timed out.` — and never once mode 2. Zero occurrences of `buffer too small` in any of them.

That is expected in hindsight and worth recording: with the P0 *application* still at 115,200, everything the P0 says is garbage at the core's rate, permanently, so the RX buffer just fills and `Ready.` is never recognised. Mode 2 needs the original campaign's configuration — P0 flashed to the target rate, so that only the *bootloader* phase is garbage and a readable `Ready.` follows it. Reproducing that needs the `p0.flash()` + NVS-planting trick from #254, which I did not want to do on a production robot.

So: this PR's fix is verified against the oversized-line mechanism, but **not** against the specific cross-baud sequence in #253. Combined with your own note that `RX_BUF_SIZE` and `boot_timeout` are deliberately untouched, #253 should stay open — and my four attempts suggest mode 1 is the dominant one there, i.e. the two constants you left alone are doing most of the work.

</details>

<details>

Feldfreund F31 / rb67, Robot Brain with NAND gates, core on `/dev/ttyTHS1`, rosys stopped, startup checksum `4a6c` verified before and after. P0 untouched throughout (still `0439c76`, build of 2026-08-05). "after" firmware was `origin/main` + #261 + #262 + #263 merged; `serial`/`p0` created at runtime because upstream Lizard cannot run F31's production startup script (private modules). Robot fully restored afterwards, rosys reports `local checksum: 4a6c` / `core checksum: 4a6c`.

</details>

evnchn:

**TL;DR — thank you; the middle row of your table is the `i <= pos` fix caught in the act. Hardware box checked, scoped to the oversized-line mechanism only.**

`Checksum mismatch` at 1 per trigger → 0, with the next line intact 3/3, is the desync my reviewer predicted from a static read — reproducing on every trigger and then dying. I had that only as a host-level ASan reproduction against `strip()`/`check()`; you've now shown it on the wire. The left column is the other half: `error in module "p0"` → `p0: error while handling messages` is the exception no longer escaping to `run_step()`.

Your trigger is the part I'd have missed and the part worth keeping: using the P0's *own* 45-byte error echo as the amplifier, so a 1013-byte command that fits the core's input buffer provokes a 1045-byte reply that doesn't. That's a reusable recipe for any future oversized-line work on this link.

<details>
<summary><b>#253 stays open, and your four failed attempts are the useful finding</b></summary>

Four attempts, all landing on mode 1 (`connection timed out`) and never once on mode 2, is not a null result — it says the cross-baud case is dominated by the buffer filling with permanent garbage, i.e. by exactly the two constants I declined to touch (`RX_BUF_SIZE` 2048 → 8 KB per your measurement, `boot_timeout` 5 s → 15 s). This PR fixes the mode that is unambiguously a bug; those two remain a tuning decision for @falkoschindler, with the RAM cost I noted in the body (`RX_BUF_SIZE` applies to *every* `uart_driver_install`, so it is 4× per UART, not 4× once).

Your reasoning for why mode 2 needs the original campaign's configuration — P0 flashed to the target rate so only the *bootloader* phase is garbage and a readable `Ready.` follows — matches the mechanism as I understand it, and I agree that reproducing it via the #254 NVS-planting trick is not worth doing on a production robot.

</details>

Also noted your independent check of the considered-and-declined `available() < pos` guard at `serial.cpp:135`, and of the deliberate asymmetry in not refreshing `last_message_millis` on the catch path. Both were the calls I was least sure about, so a second opinion on them is worth more than one on the parts I was confident in.

evnchn:

**One ordering request for whoever merges: #263 before this one.**

Either order is functionally fine — both guards stay correct, and nothing about the shipped behaviour depends on it. But merging this PR first makes #263's description untrue the moment it lands, because the `Serial::read_line` throw that #263 lists as covered is swallowed by the catch added here. #263 first, and every claim in both bodies is true at every point in the history.

JensOgorek (approved):
Approved — both halves hardware-verified: newline-desync fix on rb67 (checksum mismatch 1-per-trigger → 0), and the catch observed firing in the real cross-baud scenario on the bench pair (3× oversized-line discard at 921600 with the module surviving construction). Scoping is right: bench matrix confirms the untouched constants are the load-bearing part of #253, which stays open. Note for the merger: #263 first, per the author's ordering request.

JensOgorek:
Status check: approved, CI green, hardware-verified on F31 — nothing left from my side. @falkoschindler anything still open, or can this be merged? Note the ordering request above: #263 should go in before this one.

falkoschindler (changes requested):
Full disclosure: this is a pure Fable review (multi-agent finder pass with adversarial verification of each finding) — I haven't checked the findings myself yet. But two of them look major enough that I'd rather block the merge until we've had a look.

The core fix is good: the `i <= pos` change correctly consumes the pattern byte (matching the `pos + 1` read in the success path), and the guards turn what used to be a startup-script abort into a graceful skip-and-continue. The findings below are about completing it.

### 1. The `available()` guard is now off by one — main/modules/serial.cpp (line 135) — major

The discard loop now drains `pos + 1` bytes, but the feasibility check above it still reads `available() < pos`. A full line including the newline is `pos + 1` bytes, so the guard should be `available() < pos + 1` (or `available() <= pos`). At the boundary `available() == pos`, the code takes the discard path believing the whole line is buffered, the final `read()` returns -1 (timeout 0), and the pattern queue is left undrained even though the state is exactly the desynced one the flush branch exists to repair. On main the guard matched the old `pos`-byte loop, so this mismatch is introduced by this PR — a one-character completion of the fix.

### 2. The third `read_line` call site still reboots the ESP — main/modules/serial_bus.cpp (line 166) — major

`SerialBus::process_uart` calls `read_line` with a 512-byte frame buffer and no try/catch, and it runs directly in the FreeRTOS task `communication_loop` — the task's only try/catch covers the poll-response block, not `process_uart`. Any single bus line of 512+ bytes (misbehaving or foreign peer, boot log spew from a rebooting peer, line noise delaying a newline — compliant peers cap payloads at 256 bytes) makes `read_line` throw, the exception escapes the task entry function, and the ESP aborts and reboots. The throw paths pre-exist on main, so this is not a regression — but it is exactly the crash class this PR sets out to fix, left open at the one call site where the consequence is a full device reboot rather than a caught error.

### 3. Suggestion: sentinel return instead of throw (fixes 2 for free)

Several review angles independently converged on this: the "line discarded" condition is routine peer misbehavior, yet it is signaled by a throwing exception that every caller must remember to catch or crash. If `Serial::read_line` returned -1 for "line discarded/flushed" instead of throwing, then (a) both new copy-pasted catch blocks in expander.cpp collapse to `if (len < 0) { echo(...); continue; }`, (b) the `int len = 0` dummy initializer disappears and `const int len` comes back, (c) serial_bus.cpp is fixed by one `len < 0` check instead of more try/catch boilerplate, and (d) the hot step loop stops paying a heap-allocating C++ throw per oversized line when a peer misbehaves persistently. If the throw contract is kept instead, the two identical catch blocks should at least move into a private helper so they cannot drift apart.

### 4. Minor: the rare flush branch now silently skips the strapping-pin gate — main/modules/expander.cpp (line 141)

In the non-force flash path, `handle_messages(true)` now swallows `read_line` errors. In the common "discarded line" case that is fine — only the oversized line is dropped and the queued `GPIO_Status` responses are still read and checked, which is actually better than main's behavior of aborting the flash after NVS was already cleared. But in the rare "cannot discard" branch, `uart_flush_input` wipes the entire RX buffer including the not-yet-read `GPIO_Status` lines; the loop then exits, `check_strapping_pins` never sees any pin status, and flashing proceeds unchecked, where main aborted with a propagated error. The gate was already fail-open by contract (no response also meant no check, and checksum mismatches skipped it too), so this only widens an existing gap — but if we want the gate strict, `handle_messages(true)` could rethrow, or track "saw all three GPIO_Status responses" and abort otherwise.

### 5. Minor: byte-by-byte discard — main/modules/serial.cpp (line 142)

The discard loop issues `pos + 1` separate one-byte `uart_read_bytes` calls — at least 1025 driver round-trips per discarded line, on a path that can run every 10 ms step while a peer keeps sending long lines. Draining in chunks (e.g. into the caller-provided buffer, `min(remaining, buffer_len)` at a time) would do it in two calls. No blocking risk either way, since `read()` uses timeout 0 — purely overhead.

### Pre-existing findings (out of scope here, but worth follow-up issues)

- main/main.cpp (line 420): the UART0 REPL path `process_uart()` pops a pattern position and reads `pos + 1` bytes into the static 1024-byte `input` buffer with no oversize guard, while the RX ring is 2048 bytes — a single host line over 1024 bytes writes up to ~1024 host-controlled bytes past the buffer. Same bug class this PR fixes, duplicated inline without the guard, and it needs no degraded driver state to trigger. This one deserves its own issue.
- main/modules/serial.cpp (lines 137-138): the manual pattern-queue drain after `uart_flush_input` is redundant on IDF v5.3.1 (verified against the IDF source: the flush reconciles the queue itself) and can briefly hide a line that arrives mid-flush. Cleanup would be deleting the loop.
- main/modules/serial.cpp (line 134): `pos >= buffer_len` accepts a line that fills the buffer completely with no room for the NUL that `strip()` writes later — a very narrow one-byte overflow, only reachable in a desynced state. `pos + 1 >= buffer_len` would close it.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20conducted%20and%20written%20entirely%20by%20the%20model%3A%20multi-agent%20finder%20pass%2C%20deduplication%2C%20and%20adversarial%20verification%20of%20each%20finding.&d=2026-08-13&a=claude-code "claude-fable-5: Review conducted and written entirely by the model: multi-agent finder pass, deduplication, and adversarial verification of each finding.")

evnchn:

**Took your suggestion 3 — and it made this better than what I'd have written.** Under the throw contract, fixing finding 2 meant a third copy of the same catch block, which is why I'd scoped it out. Sentinels make it four lines, so the scope argument just evaporates. Findings 1, 2 and 3 are closed in one change. Finding 1 was genuinely ours, and a good catch.

**It's a single commit on top of what you reviewed (`9e4b39a`), so it's easy to throw away.** If you'd rather keep the throw contract, drop it and I'll do the try/catch-plus-helper version instead — no attachment on my side, and you shouldn't feel you have to take a refactor you didn't ask to own.

**One thing to know before re-reviewing:** this is no longer the diff @JensOgorek hardware-verified. The console strings are byte-identical so his bench result should reproduce, but "should" isn't "did", and I didn't want his green check sitting under a diff it didn't cover.

@JensOgorek — if you get a chance, could you re-run the oversized-line trigger on rb67 against this commit? Same two observations as last time. No rush.

<details>
<summary><b>What changed</b></summary>

`Serial::read_line` now returns `LINE_DISCARDED` (-1) or `LINE_FLUSHED` (-2) instead of throwing. All three call sites — both expander loops and `SerialBus::process_uart` — check `len < 0`. The feasibility guard becomes `available() <= pos`.

```
 main/modules/expander.cpp   | 16 ++++++----------
 main/modules/serial.cpp     | 11 ++++++++---
 main/modules/serial.h       |  3 +++
 main/modules/serial_bus.cpp |  4 ++++
```
</details>

<details>
<summary><b>Findings 1 and 2 — confirmed independently before acting on them</b></summary>

**Finding 1 — you're right about the provenance.** On `main` the guard `available() < pos` matched the `i < pos` discard loop; this PR changed the loop to `i <= pos` and left the guard behind. Fixed.

**Finding 2 — mechanism confirmed:** `SerialBus::communication_loop` is `[[noreturn]]` (`main/modules/serial_bus.cpp:123`) and calls `process_uart()` at line 126. Its only `try`/`catch` is the poll-response block at line 143 — it does not cover the `read_line` at line 166. `FRAME_BUFFER_SIZE` is 512 and `PAYLOAD_CAPACITY` is 256, so a compliant peer never trips it and a non-compliant one reboots the device.
</details>

<details>
<summary><b>Why two sentinels rather than the single -1 — and how to undo it if you disagree</b></summary>

`"discarded line."` and `"flushed serial."` mean materially different things — one line lost, versus the entire RX ring wiped — and Jens read those two strings off the bench when diagnosing this, so collapsing them would have cost a diagnostic that was load-bearing in the field. `Serial::read_line_error()` maps the sentinel back to the original string so the three call sites carry an `if` rather than each carrying a ternary. Message text is unchanged from `main`.

If you'd prefer the single `-1` you actually asked for, it's deleting the helper and one constant.

**One wart, flagged rather than hidden:** `uart_read_bytes` can itself return -1 on a driver error, which now aliases `LINE_DISCARDED` and would print "discarded line." for something that isn't one. Previously that -1 flowed into `check()` as a *length*, so this is strictly an improvement — but the message would be wrong. A third `READ_ERROR = -3` fixes it; I left it out because reaching it needs an invalid-parameter or uninstalled-driver state. Easy either way.
</details>

<details>
<summary><b>Finding 4 — my instinct is a separate PR, but happy to be overruled</b></summary>

Rethrowing from `handle_messages(true)` would close the flush hole while leaving the two that were already open: no response at all also means no check, and a checksum mismatch skips the gate too (`main/modules/expander.cpp:146`). That would leave a gate that reads as strict at three of four entrances, which I find harder to reason about than one that's openly fail-open — but that's a judgement call and you know this code's history better than I do.

If you want it strict, I think the honest version counts three `GPIO_Status` responses and aborts otherwise, and I'd be glad to write it as its own PR. If you'd rather have the rethrow here and now, say so and it's in.

Either way the gate's own throws (GPIO0/GPIO2 HIGH, `main/modules/expander.cpp:229` and `:234`) sit outside the guarded region and still propagate, so the strict half is intact.
</details>

<details>
<summary><b>Finding 5 — confirmed, offered</b></summary>

Confirmed: `Serial::read()` is a one-byte `uart_read_bytes`, so the discard loop really is `pos + 1` driver calls. Draining in chunks through the caller's buffer is about five lines and safe, since we return a negative and every caller `continue`s without reading the buffer.

Left out only to keep the diff near what you blocked on. Say the word and it's in.
</details>

<details>
<summary><b>Pre-existing findings — two confirmed, one deliberately not filed yet, one I take no position on</b></summary>

**`main/main.cpp:420` — confirmed, and the sizing makes it worse than "same bug class".** `input` is `BUFFER_SIZE` = 1024 (line 40) while the ring is installed at `BUFFER_SIZE * 2` = 2048 (line 462), and `uart_read_bytes(..., input, pos + 1, 0)` has no guard at all — so the overflow is dimensioned in rather than incidental.

**Not filed yet, on purpose.** I can write the repro but not yet predict its symptom: the write lands in `.bss` and what it clobbers is link-order dependent, so "send a >1024-byte line and see what happens" would be a code-reading dressed as a repro. Running it on a board first, then filing with an observed symptom. Flagging it here so it isn't lost meanwhile.

**`main/modules/serial.cpp:134` NUL-room — confirmed, and here's why it's as narrow as your review says.** `strip()` decrements *before* writing its NUL (`main/utils/uart.cpp:43-52`), and on a well-formed line the last byte is the `\n` pattern byte, so it always decrements at least once and lands in bounds. It needs a desynced pattern position to bite at all.

**`main/modules/serial.cpp:137-138` redundant drain — I haven't verified this against the IDF v5.3.1 source myself**, so I'd rather take no position than repeat the claim.
</details>

<details>
<summary><b>Build evidence and second-lineage review</b></summary>

Full clean builds of both targets, post-change:

```
esp32     lizard.bin 0x13ba20 bytes, 0xb45e0 free (36%)   [1161 targets, exit 0]
esp32s3   lizard.bin 0x139c70 bytes, 0xb6390 free (37%)   [1204 targets, exit 0]
```

Both shrank by ~240 bytes versus the throw version — the exception machinery coming out.

Reviewed by OpenAI Codex (gpt-5.5) against the actual files, prompted to refute. Verdict: **ship, high confidence, no blockers.** Its one substantive point was the `uart_read_bytes` -1 aliasing above, which I chose to surface to you rather than patch silently. It also independently confirmed that no negative `len` reaches `check()` / `strip()` / `parse_message()` / `strcmp` at any of the three sites, that no negative-return path can livelock, and that `print_to_incoming_queue` is safe in that task context.

What this does **not** cover: the sentinel paths have not run on hardware. That's the ask at the top.
</details>

### Open questions, all of them low-stakes

1. **Findings 4 and 5** — leave as answered, or want either patched in?
2. **The `-1` aliasing** — add `READ_ERROR = -3`, or leave it?
3. **Ordering** — Jens's note said this should land after the other PR. Still the case?

evnchn:
**Filed** — [issue 270](https://github.com/zauberzeug/lizard/issues/270), with the observed symptom I said I wanted before filing.

<details>
<summary><b>It was worth waiting for the board</b></summary>

Reproduced on an ESP32 devkit: 1279 `A`s survive, 1280 panic with `LoadProhibited`. The boundary matches the link map exactly — `core_module` sits at `input + 1280`, and the crash starts on the first write that reaches it.

The reason I did not want to file it from code alone: at 1114 bytes it only reaches a scratch buffer and surfaces as `error processing uart0: buffer too small`, thrown downstream by `csprintf` and caught in the main loop. The device stays up. Filed from reading, it would have looked like a handled error rather than a device panic.

Your review found it; it just deserved a repro before it got a ticket.
</details>

falkoschindler (approved):
Verified 9e4b39a against the review (again as a Fable pass, diff-checked against 8a5a29c): findings 1–3 are cleanly closed — the guard is now `available() <= pos`, all three call sites handle the negative sentinels, and no negative `len` can reach `check()`. The two-sentinel design with `read_line_error()` is a good call; keeping the two bench-diagnostic messages distinct is worth the extra constant. Answers to the three open questions:

**1. Findings 4 and 5 — leave both as answered.** On the gate: agreed with your reasoning — a rethrow that closes one of three holes would make the gate read stricter than it is. The honest strict version (count three GPIO_Status responses, abort otherwise) as its own PR sounds right, and with #263 merged the boot pin now survives a gate abort, so a stricter gate got safer to add. On the chunked discard: also out of here; it touches the exact path the hardware re-run covers, and I'd rather not reset that.

**2. The `uart_read_bytes` -1 aliasing — leave it.** Reaching it takes an uninstalled driver or invalid arguments, and the current behavior is strictly better than what main did with that -1. If you write the strict-gate PR, fold `READ_ERROR = -3` and the chunked discard in there.

**3. Ordering — moot.** #263 merged today, and this PR reports mergeable/clean against main, so there's nothing left to sequence.

Thanks for filing #270 with an on-board repro instead of a code-reading — the 1280-byte boundary landing exactly on `core_module` is the kind of symptom description that makes that ticket actionable.

Approving from the software side. One caveat before merge: this approval covers the code, not the bench — please wait for @JensOgorek's re-run of the oversized-line trigger on this commit, since the sentinel paths haven't run on hardware yet.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Recommendations%20and%20wording%20from%20the%20model%2C%20based%20on%20its%20verification%20of%20the%20revised%20diff%3B%20the%20decisions%20were%20confirmed%20by%20the%20author.&d=2026-08-13&a=claude-code "claude-fable-5: Recommendations and wording from the model, based on its verification of the revised diff; the decisions were confirmed by the author.")

JensOgorek:
### Hardware re-test of `9e4b39a` passed — 3/3 clean, no checksum desync (bench S3 pair)

Result up front: I re-ran the oversized-line trigger against the sentinel commit and found **no errors** — the discard is reported locally with the byte-identical console string, and the line after each trigger arrives intact. Nothing left from my side: the "Tested on target hardware" box is now checked accordingly, and this is good to merge.

<details>
<summary><b>Setup — bench S3 pair instead of rb67</b></summary>

Two bench ESP32-S3s wired as a real core→expander pair (UART link plus boot/EN lines), core flashed with `9e4b39a`, expander untouched on its previous test build — so the robot stayed out of it entirely. Same amplifier as the rb67 run on 2026-08-06: `expander.run("<1000 a's>")` stays under the core's own 1024-byte input limit, and the expander's `unknown variable` error echo amplifies it to a ~1045-byte reply.

</details>

<details>
<summary><b>The three runs — same two observations as on rb67</b></summary>

| | `9e4b39a` (sentinels) |
| --- | --- |
| reported as | `e10: error while handling messages: buffer too small. discarded line.` — handled locally, names the expander |
| `Checksum mismatch` right after | **0**, in 3/3 triggers |
| following `core.info()` reply | 3/3 intact, valid checksums |

So the refactor preserves both halves of what I verified on rb67: the console string is byte-identical, and the `i <= pos` fix still holds — no leftover `\n`, no desync on the next line. Bonus datapoint: this run had the core on an ESP32-S3 (the rb67 run was a classic Robot Brain), so the sentinel path has now seen both targets on hardware.

</details>

Reviewed and tested with Claude Fable 5.

PR #263, opened 2026-08-06
Restore expander boot pin on a failed strapping-pin check

evnchn:
**TL;DR** — Six lines, one file. If the pre-flash strapping check throws, `boot_pin` stayed LOW and the next natural P0 reset dropped into the ROM bootloader. Restore it and rethrow. This is the issue's **side finding only** — the headline first-attempt TIMEOUT is untouched.

### Motivation

Issue #255 — expander flashing reliably fails on the first attempt with `Error while connecting: TIMEOUT`, and succeeds on the second. Jens Ogorek recorded a side finding while diagnosing it, and that side finding is a real, self-contained bug that this PR fixes.

In `Expander::flash()`, the code drives `boot_pin` LOW, then runs the remote strapping-pin check via `handle_messages(true)`. If that check throws, the function aborts **before** reaching `deinstall()` with `boot_pin` still LOW — so the next natural P0 reset enters the ROM bootloader unintentionally. Which is exactly the kind of state that makes a subsequent flash attempt behave differently from the first.

### Implementation

Wrap the `handle_messages(true)` call in `try`/`catch(...)`, restore `boot_pin` HIGH, and rethrow unchanged. `gpio_set_level(..., 1)` matches the boot-pin idle level the constructor establishes at `main/modules/expander.cpp:55`.

No new exception is introduced and nothing is swallowed — the original error text reaching `main.cpp` is byte-identical to before, e.g. `error processing uart0: GPIO0 current state is HIGH - must be LOW for boot mode`. Relevant because CONTRIBUTING.md:91 asks that exceptions be used sparingly.

<details>
<summary><b>Which throws this actually covers — more than the commit subject says</b></summary>

Between `gpio_set_level(this->boot_pin, 0)` (`main/modules/expander.cpp:179`) and `deinstall()` (`:192`), the guard wraps the whole `handle_messages(true)` call, so it covers:

- ~~`Serial::read_line` throws — `main/modules/serial.cpp:139` *"buffer too small, but cannot discard line. flushed serial."* and `:144` *"buffer too small. discarded line."*~~ **Struck: true only on `main` as it stands today.** #262 wraps that `read_line()` call inside `Expander::handle_messages()` in its own catch that echoes and `continue`s, so once #262 lands the throw never propagates out of `handle_messages(true)` and can no longer reach this guard. Caught by @JensOgorek — the merge order of the two PRs decides which paths this guard is protecting.
- `check_strapping_pins()` — `main/modules/expander.cpp:226` and `:231`
- any throw from the `message_handler` callback (`:149`) during normal Lizard processing
- theoretically, an `echo()` callback (`main/utils/uart.cpp:33`) or the `std::string` allocation inside `check()` (`:63`) — surfaced by a second-lineage re-enumeration of this list, not realistic in practice

The three `write_checked_line()` calls do not throw, `delay()` does not throw, and `Storage::clear_nvs()` runs *before* the pin is pulled low, so an NVS failure there leaves the pin in its pre-existing high state. The `catch(...)` breadth is deliberate, not accidental over-reach.
</details>

<details>
<summary><b>Where the bootloader-entry retry lives — investigated, and it is not ours</b></summary>

Jens's second suggestion was to include the bootloader-entry reset sequence in the retry loop. That code is in **`components/esp32-serial-flasher`**, a vendored submodule (upstream `espressif/esp-serial-flasher`), not in Lizard's own sources. `ZZ::Replicator::connect()` and `loader_port_enter_bootloader()` are not Lizard code, so a retry there is a submodule bump plus an upstream change — unverifiable without the hardware and out of scope for this PR.

The NAND-polarity question in the issue is left open deliberately: the Robot Brain schematic is not available here, and guessing at a polarity change in code would be exactly the wrong move.
</details>

<details>
<summary><b>Build evidence — both targets</b></summary>

```
lizard.bin binary size 0x13ba20 bytes. Smallest app partition is 0x1f0000 bytes. 0xb45e0 bytes (36%) free.

lizard.bin binary size 0x139c80 bytes. Smallest app partition is 0x1f0000 bytes. 0xb6380 bytes (37%) free.
```
The built artifact was independently checked to actually contain the new code rather than trusting the log text.
</details>

<details>
<summary><b>Adversarial review — verbatim second-lineage verdict (OpenAI Codex, gpt-5.5)</b></summary>

```
BLOCKERS: none.

SUGGESTIONS:
- none

VERDICT: ship. I could not refute this narrow fix. Between main/modules/expander.cpp:179 and `deinstall()` at main/modules/expander.cpp:192, the only realistic throwing path is inside `handle_messages(true)`: `Serial::read_line()` can throw at main/modules/serial.cpp:139, `check_strapping_pins()` throws at main/modules/expander.cpp:226, and the `message_handler` callback can throw via normal Lizard processing. The three `write_checked_line()` calls do not throw, `delay()` does not throw, and `Storage::clear_nvs()` is before the pin is pulled low, so an NVS failure there leaves the pin in its pre-existing high state. Later failures after `deinstall()` are unchanged by this diff and do not match the side-finding mechanism; `flashReplica()` reinitializes/drives GPIO0 high during bootloader entry, and the explicit `could not flash expander` throw is not made worse. `gpio_set_level(..., 1)` is consistent with the constructor's boot-pin idle level at main/modules/expander.cpp:55. Exceptions are enabled in both default configs, and the bare rethrow preserves the original runtime error message for the existing `main.cpp` catches. Confidence: high for this minimal side-finding fix.
```
</details>

<details>
<summary><b>Deliberately out of scope</b></summary>

- **The headline complaint of #255 is not fixed.** First-attempt `esp_loader_connect()` TIMEOUT is untouched.
- The abort paths **after** `deinstall()` — `Storage::save_startup()` at `expander.cpp:197` can throw on NVS failure, and there is the explicit *"could not flash expander"* throw at `:202` — leave `boot_pin` in whatever state esp-serial-flasher's deiniter left it. That is squarely the issue's main complaint territory (undefined pin state across the loader) and is unchanged here. Considered, not missed.
- NAND polarity: no schematic available, so no guess.
</details>

### Progress

- [ ] Tested on hardware (or is not necessary). **Not done — and @JensOgorek has since argued it may not be safely triggerable at all.** He reviewed the change on real hardware and found nothing blocking, but could not provoke the guard: the only remaining throw is a genuine strapping-pin failure, which on a live Feldfreund means deliberately driving the P0's GPIO2. He also established that the obvious verification route is closed — `core.get_pin_status(25)` reports `Level: 0 | InputEn: 0 | OutputEn: 1`, i.e. with the input buffer disabled the `Level` field does not reflect the driven output level, so the boot pin's state cannot be read back this way. Ceiling therefore stands at green builds for both targets, static analysis, adversarial review, and his review. No closing keyword: #255's main problem is still live and this must not auto-close it.

### DIFF
--- a/main/modules/expander.cpp
+++ b/main/modules/expander.cpp
@@ -182,7 +182,12 @@ void Expander::call(const std::string method_name, const std::vector<ConstExpres
             this->serial->write_checked_line("core.get_pin_status(2)");
             this->serial->write_checked_line("core.get_pin_status(12)");
             delay(100);
-            this->handle_messages(true);
+            try {
+                this->handle_messages(true);
+            } catch (...) {
+                gpio_set_level(this->boot_pin, 1);
+                throw;
+            }
         }
         deinstall();
         bool success = ZZ::Replicator::flashReplica(this->serial->uart_num,


JensOgorek:
### Code review clean, but I could not verify this one on hardware

Review found nothing blocking. The `boot_pin == GPIO_NUM_NC` question answers itself — `Expander::call()` throws at `expander.cpp:175-177` before the flash branch ever reaches your try/catch, and `docs/module_reference.md:1115` states the contract ("The `flash()` method requires the `boot` and `enable` pins to be defined"). HIGH is the right idle level per the constructor at `expander.cpp:55`, and the bare `throw;` preserves the original message.

The honest part: **I could not trigger the guard on hardware**, for two reasons.

**1. There is no safe trigger on a production robot.** The only remaining throw path is a genuine strapping-pin failure, which requires GPIO0 or GPIO2 to read HIGH on the P0. GPIO0 is held LOW by the core at that moment by construction, so it would have meant deliberately driving the P0's GPIO2 — a strapping pin on a live Feldfreund. I did not do that.

**2. #262 closes the other path.** With #262 merged, an oversized line during the pre-flash `handle_messages(true)` no longer propagates — it is caught and skipped inside `handle_messages()`, so it can never reach this PR's `catch (...)`. That is not an objection to either PR, but it is worth stating explicitly: **once #262 lands, this guard is reachable only via `check_strapping_pins()`**, not via `Serial::read_line`. Your PR body lists `read_line` as one of the throws it covers; after #262 that stops being true.

I also found that the obvious verification method does not work: `core.get_pin_status(25)` reports `GPIO_Status[25]| Level: 0| InputEn: 0| OutputEn: 1| …` — with the input buffer disabled the `Level` field does not reflect the driven output level, so the boot pin's state cannot be read back this way.

<details>
<summary><b>An accidental data point for #255: the first flash attempt failed</b></summary>

While trying to construct a trigger I did provoke a real `p0.flash()` on rb67 (the oversized lines were all consumed before `handle_messages(true)` ran, so the check never threw and the flash proceeded). Useful by-catch:

- `esp_loader_connect()` **succeeded on the first attempt** here — `Connecting..` at t+8240 ms, `Raising baudrate..` at t+8826 ms, no TIMEOUT. So #255's "reliably fails on the first attempt" did not reproduce in this instance, consistent with the issue's own note that it is not deterministic.
- The replication ran to ~950/2031 kb and then did **not** take: afterwards the P0 still reported its original `v0.12.0-7-g0439c76`. So the flash failed late, silently from the operator's point of view.
- `p0.is_ready` stayed `false` afterwards and only recovered on `core.restart()`.
- The core's startup script survived intact (checksum `4a6c` before and after), i.e. the `clear_nvs()` / `save_startup()` dance did its job.

</details>

<details>

Feldfreund F31 / rb67, Robot Brain with NAND gates, core on `/dev/ttyTHS1`, rosys stopped. Firmware under test was `origin/main` + #261 + #262 + #263 merged. Robot fully restored afterwards; rosys reports `local checksum: 4a6c` / `core checksum: 4a6c`.

</details>

So: happy with the change on review, but the "Tested on hardware" box should stay unchecked from my side.

evnchn:

**TL;DR — you're right and the PR body was wrong. Corrected: once #262 lands, `Serial::read_line` is no longer a path into this guard, and the merge order of the two PRs decides what this guard actually protects.**

Verified in the source rather than taking it on trust: #262 wraps the `read_line()` call inside `Expander::handle_messages()` in its own `catch (const std::runtime_error &)` that echoes and `continue`s, so the throw cannot propagate out of `handle_messages(true)` and can never reach the `catch (...)` here. The bullet is struck in the body with the reason, credited to you.

<details>
<summary><b>Re-enumerated what does still reach it</b></summary>

Ran the list past a second lineage adversarially rather than just deleting one bullet, since the body's claim to be exhaustive is what failed:

- `check_strapping_pins()` — `expander.cpp:226` (GPIO0) and `:231` (GPIO2). The realistic one, and after #262 the *only* realistic one.
- the `message_handler` callback (`:149`) — still escapes, so your "reachable only via `check_strapping_pins()`" is very slightly narrow.
- theoretically: an `echo()` callback (`utils/uart.cpp:33`) and the `std::string` allocation inside `check()` (`:63`). Listed for completeness, not because I think either fires.

`has_buffered_lines()` and `read()` do not throw. So the guard keeps a real job after #262 — just a smaller one than the body advertised.

</details>

**Hardware box stays unchecked, and I agree with both of your reasons.** Deliberately driving the P0's GPIO2 on a live Feldfreund to manufacture a trigger is not a trade I'd take either, and the `core.get_pin_status(25)` finding — `InputEn: 0`, so `Level` doesn't reflect the driven output — closes the one non-invasive route. I've put that in the PR body so the next person doesn't spend the same hour on it.

Your accidental flash is the most interesting thing in your comment, and it isn't #255's reported failure: `esp_loader_connect()` **succeeded** first try, then replication stopped taking at ~950/2031 kb with the P0 still reporting `v0.12.0-7-g0439c76` afterwards — a late, silent failure rather than a connect TIMEOUT. Worth posting on #255 as a distinct data point; it may well be the same root cause seen past the connect phase, but it isn't the same symptom.

JensOgorek:
### Hardware-verified after all — clean A/B on the bench pair

Follow-up to my earlier comment: with the bench pair now fully wired (EN + IO0 lines working), the guard turned out to be safely triggerable after all — my "no safe trigger" concern applied to production robots, not to a bench P0. Both legs use the identical trigger; only the core firmware differs.

| core firmware | strapping check | `p0.restart()` afterwards |
|---|---|---|
| `main` (020b32d) | throws `GPIO2 current state is HIGH - must be LOW or floating for flash mode` | **`DOWNLOAD_BOOT` — P0 stranded in the ROM bootloader** (the bug: `boot_pin` left LOW) |
| main + this PR | identical throw | **normal boot, `Ready.`** — the catch restored the pin |

That is exactly the side finding from #255, reproduced and fixed: same throw, same restart, and the P0's fate flips with the six lines of this PR. Verified with the P0's own console observed directly, so the boot mode is read from the ROM banner, not inferred.

<details>
<summary><b>Trigger recipe — and why the obvious one does not work</b></summary>

Working trigger, entirely from the core side, no hardware manipulation:

```
p0.run("g2 = Input(2)")
p0.run("g2.pullup()")
p0.flash()        → throws the GPIO2 error
p0.restart()      → observe the P0's boot mode on its console
```

The obvious variant — `p0.run("g2 = Output(2)")` + `g2.on()` — does **not** trigger the check, and the reason is worth recording: `core.get_pin_status()` reports driven outputs with `InputEn: 0`, so `Level` reads 0 regardless of the actual pin state. Two consequences beyond this test: the strapping check cannot detect a remote whose own application drives GPIO2 high, and (as noted before) `get_pin_status` cannot be used to read back the core's own `boot_pin` either. On my first attempt this silently skipped the check and started a real replica flash — which is also why the recipe above uses `Input` + pull-up.

Setup: 2× ESP32-D0WD devkit bench pair (core UART1 on 27/26 ↔ P0 UART0, boot wire core-GPIO14 → P0-IO0, enable wire core-GPIO25 → P0-EN), core on `main` resp. `main` + #261 + #262 + #263, three-line startup-free runtime setup, P0 boot mode read from its ROM banner (`DOWNLOAD_BOOT(UART0/UART1/SDIO_REI_REO_V2)` / `waiting for download` vs. `Ready.`).

</details>

The "Tested on hardware" box can be checked now. Verdict unchanged from my review: ship — and per your ordering request, this merges before #262.

JensOgorek (approved):
Approved — code review clean (see comments), and the fix is now hardware-verified with a falsifiable A/B: after a real strapping-pin throw, `p0.restart()` lands the P0 in `DOWNLOAD_BOOT` on main and boots normally with this PR. Details in the comment above. Merge before #262 per the author's ordering request.

JensOgorek:
Status check: approved, CI green, hardware-verified on the bench pair — nothing left from my side. @falkoschindler anything still open, or can this be merged? This one should go in before #262.

falkoschindler (approved):

PR #264, opened 2026-08-06
Pack requirements-macos.txt into the release zip-file

falkoschindler:
### Motivation

Since #204 split the Python requirements, `requirements.txt` starts with `-r requirements-macos.txt`. The release packaging list `.github/zip_content.txt` was not updated accordingly, so the `lizard_firmware_and_devtools_*.zip` assets ship a `requirements.txt` whose include target is missing. Installing from the zip fails with `failed to read from file 'requirements-macos.txt': No such file or directory` — this broke the RoSys v0.35.0 Docker build (https://github.com/zauberzeug/rosys/actions/runs/31081873200), which installs the requirements from the latest Lizard release zip. The v0.13.0 assets have been patched in place; this PR fixes the packaging for future releases.

### Implementation

Add `requirements-macos.txt` to `.github/zip_content.txt` so it is packed into the release zip alongside `requirements.txt`.

### Progress

- [x] Test on hardware is not necessary.
- [x] Documentation is not necessary.

### DIFF
--- a/.github/zip_content.txt
+++ b/.github/zip_content.txt
@@ -8,4 +8,5 @@ monitor.py
 upload_ssh.py
 README.md
 requirements.txt
+requirements-macos.txt
 LICENSE


evnchn:
should just merge?

JensOgorek:
### Code review clean, packaging path verified end-to-end

No issues found. I traced the change through to the produced artifact rather than just eyeballing the one-line diff:

- `requirements.txt:1` is indeed `-r requirements-macos.txt`, and `requirements-macos.txt` exists at the repo root.
- `.github/workflows/release.yml:41` consumes the list with `zip lizard_${target}.zip -@ < .github/zip_content.txt`, so `zip -@` packs each named path verbatim — adding the line does get the file into the asset.
- I checked the other entries for the same class of omission: every script listed in `zip_content.txt` (`configure.py`, `espresso.py`, `monitor.py`, `upload_ssh.py`) imports only stdlib and third-party packages, no local modules. `requirements-docs.txt` is correctly absent — it is only used by the separate `docs` job at `release.yml:115`, not by consumers of the firmware zip. No further missing files.

No hardware test applies to this one.

Reviewed with Claude Opus 5.

PR #265, opened 2026-08-06
Remove dead Kconfig symbols from sdkconfig defaults

Johannes-Thiel:
Both `sdkconfig.defaults` files carry five lines that have no effect on the build. Three of them name Kconfig symbols that do not exist in ESP-IDF v5.3.1 — kconfgen prints `warning: unknown kconfig symbol` for each on every configure run and then discards them. The other two name real symbols, but assign values that kconfgen ignores.

The one that matters is `CONFIG_ESP_SERIAL_FLASHER_MD5_ENABLED=y`. The real symbol is `SERIAL_FLASHER_MD5_ENABLED`, without the `ESP_` prefix. MD5 verification after flashing a peer is currently active only because the component's Kconfig defaults it to `y` — not because we ask for it. Anyone who later sets that line to `n` to disable the check would see nothing happen, and the three-line warning would be lost in a 1200-line ninja log. This PR corrects the prefix so the line does what it says, and removes the other four.

The two coredump lines are the more surprising ones: **the core dump partition has never been selected by config**. `esp_core_dump_flash` looks it up at runtime by partition type/subtype, so `CONFIG_ESP_COREDUMP_FLASH_PARTITION="coredump"` was pointing at nothing, and the `coredump` row in `partitions.csv` was doing all the work by itself.

**Verified byte-identical**: a clean `esp32s3` rebuild before and after the change produces an identical `sdkconfig` and `sdkconfig.h`. Unknown-symbol warnings go from three to zero. **Hardware-validated** on an ESP32-S3: the full core dump chain — panic, write to flash, checksum verification on the next boot, and symbolized decode of all 7 threads — still works with these lines gone.

<details>
<summary>Symbol-by-symbol analysis and validation</summary>

### What each line actually does

| Line (both files) | Status in IDF v5.3.1 | Effect |
|---|---|---|
| `CONFIG_ESP_SERIAL_FLASHER_MD5_ENABLED=y` | does not exist | warning; real symbol is `SERIAL_FLASHER_MD5_ENABLED` (`components/esp32-serial-flasher/Kconfig`), which defaults to `y` |
| `CONFIG_ESP_COREDUMP_FLASH_SAVE=y` | does not exist | warning; overlaps conceptually with `ESP_COREDUMP_ENABLE_TO_FLASH`, which is already set two lines up |
| `CONFIG_ESP_COREDUMP_FLASH_PARTITION="coredump"` | does not exist | warning; there is no config for the partition (see below) |
| `CONFIG_ESP_COREDUMP_ENABLE=y` | exists | no prompt, value comes only from `select` in `ESP_COREDUMP_ENABLE_TO_FLASH`; kconfgen ignores assignments to promptless symbols |
| `CONFIG_ESP_COREDUMP_DATA_FORMAT_BIN=n` | exists | choice member; `=n` is meaningless once `DATA_FORMAT_ELF=y` wins the choice |

These are not deprecated spellings. `components/espcoredump/sdkconfig.rename` maps all 14 renamed coredump options from the IDF 3.x/4.x era (`CONFIG_ESP32_ENABLE_COREDUMP_TO_FLASH` → `CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH` and friends); neither `FLASH_SAVE` nor `FLASH_PARTITION` appears there. A full-text grep over the IDF tree finds zero occurrences of either string.

### How the core dump partition is really located

`components/espcoredump/src/core_dump_flash.c:123`:

```c
core_part = esp_partition_find_first(ESP_PARTITION_TYPE_DATA, ESP_PARTITION_SUBTYPE_DATA_COREDUMP, NULL);
```

The API guide (`docs/en/api-guides/core_dump.rst`) states it explicitly: there are no requirements on the partition name, only that type is `data` and subtype is `coredump`. Our `partitions.csv` already satisfies that. The name matching the string in the config was a coincidence.

### Verification method

Baseline captured from a clean `python build.py esp32s3 --clean` on unmodified `main`, then the same clean build with the change applied:

```
diff sdkconfig            → IDENTISCH
diff build/config/sdkconfig.h → IDENTISCH
grep "unknown kconfig symbol" build/log/*  → 3 before, 0 after
```

The surviving defines confirm the reasoning — `ENABLE` is still set although its explicit line is gone, and `DATA_FORMAT_BIN` was never defined in the first place:

```
#define CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH 1
#define CONFIG_ESP_COREDUMP_DATA_FORMAT_ELF 1
#define CONFIG_ESP_COREDUMP_CHECKSUM_CRC32 1
#define CONFIG_ESP_COREDUMP_ENABLE 1          ← via select, no explicit line
#define CONFIG_ESP_COREDUMP_MAX_TASKS_NUM 16
grep -c DATA_FORMAT_BIN → 0
```

`esp32` was also rebuilt clean: build succeeds, zero warnings, same five coredump defines present.

### Hardware validation (ESP32-S3, CP2102N, IDF v5.3.1)

Boot on this branch's build, with the core dump partition erased beforehand:

```
I (470) esp_core_dump_flash: Init core dump to flash
I (475) esp_core_dump_flash: Found partition 'coredump' @ 400000 131072 bytes
Ready.
```

The full crash cycle was run on a build carrying these same edited defaults (older source base, identical generated coredump config). A deliberate panic produced:

```
Guru Meditation Error: Core 0 panic'ed (IntegerDivideByZero)
I (23741) esp_core_dump_flash: Save core dump to flash...
I (23793) esp_core_dump_flash: Core dump has been saved to flash.
[reboot]
I (491) esp_core_dump_flash: Core dump data checksum is correct
I (491) esp_core_dump_flash: Found core dump 14500 bytes in flash @ 0x400000
```

`idf.py -p … -b 115200 coredump-info` then decoded all 7 threads with `exccause 0x6 (IntegerDivideByZeroCause)` and a fully symbolized backtrace of the crashing task. ELF format, CRC32 checksum and `MAX_TASKS_NUM=16` all behaved as configured.

### Review notes

- Reading the dump from flash needs `-b 115200`; the default 460800 aborts with `Invalid head of packet` on this bridge.
- The `esp32` half of the change is compile-verified only — no ESP32 board was available.
- `core_dumper.py` hardcodes `chip='esp32'` and therefore cannot read an ESP32-S3 dump; `idf.py coredump-info` was used instead. Left alone here since it is out of scope for a config cleanup.

</details>


JensOgorek (approved):
Approved — verified end-to-end: clean rebuilds byte-identical (sdkconfig + sdkconfig.h), unknown-symbol warnings 6→0, and the falsifiable counter-test passes: setting the MD5 line to n now takes effect (`# CONFIG_SERIAL_FLASHER_MD5_ENABLED is not set`) while on main it stays y. Coredump partition found at runtime on flashed ESP32-S3 hardware. One nit: the body says "3 before" — the build log shows 6 occurrences of the 3 symbols.

PR #267, opened 2026-08-07
Read the ODrive's motor current over CAN

evnchn:
**Adds `0x014` (Get_Iq) to `ODriveMotor`, so the motor current the board already measures reaches Lizard.** Two properties, one subscription, one `case` — mirroring the existing encoder-estimate decode.

Untested on hardware — the flash needs someone at the bench. Reviewed first on the fork as evnchn/lizard PR 16.

<details>
<summary><b>Byte layout — verified against the firmware, not assumed</b></summary>

From `zauberzeug/ODrive`, branch `zz-0.5.6`, `CANSimple::get_iq_callback`:

```cpp
can_setSignal<float>(txmsg, Idq_setpoint->second, 0, 32, true);
can_setSignal<float>(txmsg, axis.motor_.current_control_.Iq_measured_, 32, 32, true);
```

Bytes 0–3 are the **setpoint**, bytes 4–7 the **measurement**, `isIntel = true` so little-endian — the same assumption the existing `0x009` decode already makes.

`MSG_GET_IQ` is 0x014, cross-checked rather than counted from a table: it is the 19th entry after `MSG_ODRIVE_ESTOP`, and the same arithmetic puts `MSG_GET_ENCODER_ESTIMATES` on 0x009 — which is exactly what this module already subscribes to.
</details>

<details>
<summary><b>Why both properties stay at 0 by default</b></summary>

The ODrive sends this message only when asked: `<axis>.config.can.iq_rate_ms` defaults to `0`. On the ts2 stand both axes read `0`, which is why the bench has always shown `0.00 A` — the board measures fine (`Iq_measured` over USB reads −0.02 A at rest and peaks of 1.9 A / 3.3 A during a move), it simply never broadcasts.

That is the same shape as the existing `motor_temperature`, which depends on the zauberzeug ODrive fork and is documented as remaining 0 otherwise, so the new properties are documented next to it.
</details>

<details>
<summary><b>Build</b></summary>

Green on ESP-IDF v5.3.1 (the version `release.yml` pins): `lizard.bin` generated, 36% of the app partition free.

Three things blocked a fresh-checkout build, none of them this change — recording them because they will bite the next person:

1. submodules not initialised → `Failed to resolve component 'esp32-serial-flasher'`
2. `compile.sh` runs a bare `idf.py build`, which does **not** apply `sdkconfig.defaults.esp32`. Without it `CONFIG_COMPILER_CXX_EXCEPTIONS` is off and every `throw` in the tree becomes `error: exception handling disabled`. A control build of **unmodified** upstream failed identically, which is how I know it was environmental.
3. `CONFIG_ZZ_BLE_DEV_PIN` needs `sdkconfig.defaults.secret`, copied from the template.

What worked: `idf.py -DSDKCONFIG_DEFAULTS="sdkconfig.defaults.esp32;sdkconfig.defaults.secret" set-target esp32 build`. Teaching `compile.sh` to do that would be a separate small PR.
</details>

Downstream, this is the last missing piece of three: the ODrive side is stock and merely switched off, and the host side already parses two currents behind a config flag (`zauberzeug/feldfreund` PR 584). After this, showing current in the panel is one config line plus persisting `iq_rate_ms`.

### DIFF
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -512,8 +512,8 @@ The `motor_temperature` will only update if the firmware generated from the
 Otherwise, the `motor_temperature` property will remain at 0.
 
 The ODrive sends its current only when asked to, so `current` and `current_setpoint` stay at 0
-until the periodic message is switched on for the axis — `<axis>.config.can.iq_rate_ms` defaults
-to 0 and is the interval in milliseconds.
+until the periodic message is switched on for the axis — `<axis>.config.can.iq_rate_ms` (requires version 0.5.6)
+defaults to 0 and is the interval in milliseconds.
 
 | Methods                        | Description                            | Arguments        |
 | ------------------------------ | -------------------------------------- | ---------------- |
--- a/docs/module_reference.md
+++ b/docs/module_reference.md
@@ -504,10 +504,16 @@ Version 0.5.6 allows to read the motor error flag.
 | `motor.motor_error`       | Motor error flat (requires version 0.5.6) | `int`     |
 | `motor.enabled`           | Whether the motor is enabled              | `bool`    |
 | `motor.motor_temperature` | Motor temperature (°C)                    | `float`   |
+| `motor.current`           | Measured motor current Iq (A)             | `float`   |
+| `motor.current_setpoint`  | Commanded motor current Iq (A)            | `float`   |
 
 The `motor_temperature` will only update if the firmware generated from the
 [zauberzeug/ODrive](https://github.com/zauberzeug/ODrive) fork is installed on the ODrive.
 Otherwise, the `motor_temperature` property will remain at 0.
+
+The ODrive sends its current only when asked to, so `current` and `current_setpoint` stay at 0
+until the periodic message is switched on for the axis — `<axis>.config.can.iq_rate_ms` defaults
+to 0 and is the interval in milliseconds.
 
 | Methods                        | Description                            | Arguments        |
 | ------------------------------ | -------------------------------------- | ---------------- |
--- a/main/modules/odrive_motor.cpp
+++ b/main/modules/odrive_motor.cpp
@@ -28,6 +28,8 @@ const std::map<std::string, Variable_ptr> ODriveMotor::get_defaults() {
         {"axis_error", std::make_shared<IntegerVariable>()},
         {"motor_error_flag", std::make_shared<IntegerVariable>()},
         {"motor_temperature", std::make_shared<NumberVariable>()},
+        {"current", std::make_shared<NumberVariable>()},
+        {"current_setpoint", std::make_shared<NumberVariable>()},
         {"enabled", std::make_shared<BooleanVariable>(true)},
     };
 }
@@ -40,6 +42,7 @@ ODriveMotor::ODriveMotor(const std::string name, const Can_ptr can, const uint32
 void ODriveMotor::subscribe_to_can() {
     this->can->subscribe(this->can_id + 0x001, std::static_pointer_cast<Module>(this->shared_from_this()));
     this->can->subscribe(this->can_id + 0x009, std::static_pointer_cast<Module>(this->shared_from_this()));
+    this->can->subscribe(this->can_id + 0x014, std::static_pointer_cast<Module>(this->shared_from_this()));
     this->can->subscribe(this->can_id + 0x01e, std::static_pointer_cast<Module>(this->shared_from_this()));
 }
 
@@ -130,6 +133,15 @@ void ODriveMotor::handle_can_msg(const uint32_t id, const int count, const uint8
             ticks_per_second *
             (this->properties.at("reversed")->boolean_value ? -1 : 1) *
             this->properties.at("m_per_tick")->number_value;
+        break;
+    }
+    case 0x014: {
+        float iq_setpoint;
+        std::memcpy(&iq_setpoint, data, 4);
+        this->properties.at("current_setpoint")->number_value = iq_setpoint;
+        float iq_measured;
+        std::memcpy(&iq_measured, data + 4, 4);
+        this->properties.at("current")->number_value = iq_measured;
         break;
     }
     case 0x01e: {
--- a/main/modules/odrive_motor.cpp
+++ b/main/modules/odrive_motor.cpp
@@ -136,12 +136,13 @@ void ODriveMotor::handle_can_msg(const uint32_t id, const int count, const uint8
         break;
     }
     case 0x014: {
+        const int sign = this->properties.at("reversed")->boolean_value ? -1 : 1;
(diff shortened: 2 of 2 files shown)


JensOgorek:
A review from my side is on the way. The one other open box is the hardware verification the PR body mentions — could you run this on ts2 (set `iq_rate_ms` > 0 on both axes and check that `motor_current` / `current_setpoint` move plausibly under load)? Once that is verified and the review is through, feel free to merge this one yourself.

JensOgorek:
### Review — two things to fix before merge, two optional

I went through the code and docs and verified the claims against the ODrive firmware sources (stock and the zz fork). The decode itself checks out — 0x014, the byte offsets and the little-endian floats are all correct. Two things should change:

**1. `reversed` is not applied to the new properties** (`odrive_motor.cpp`, case 0x014). `position` and `speed` are sign-corrected into the user frame, and `power()`/`speed()`/`position()` apply the sign on the command path — so `current`/`current_setpoint` become the only directional values in the module reported in the motor frame. On a reversed motor (the delta arm configures one, and host code stores these currents signed) `current_setpoint` reads the opposite sign of the commanded torque. Fix is two `* (this->reversed ? -1 : 1)` factors.

**2. Docs: `iq_rate_ms` needs a "(requires version 0.5.6)" note.** Stock 0.5.4 and 0.5.5 have neither the knob nor the cyclic Get_Iq send — checked `odrive-interface.yaml` and `can_simple.cpp` at tags fw-v0.5.4/0.5.5/0.5.6; both arrived in stock 0.5.6, and the zz fork inherits them (it is not fork-exclusive). Since the module's documented default is version 4, a reader on the default setup cannot follow the new paragraph. Same style as the existing `motor_error` note.

Optional, take or leave: put the new docs paragraph on one line and mirror the temperature paragraph's sentence shape (it is currently one sentence hard-wrapped mid-clause), mention "Iq" once instead of leaving it unexplained next to `limits(speed, current)` where "current" means the limit; and if you touch the case anyway, a file-local `demarshal_float` would collapse the six hand-rolled `memcpy` decodes in this file to one call site.

<details>
<summary><b>Checked and cleared — so nobody re-litigates these</b></summary>

- **RTR or short frames cannot inject garbage floats:** the IDF HAL zero-fills all data bytes beyond DLC and for RTR frames (`twai_ll_parse_frame_buffer`, esp32 and esp32s3 alike), so the worst case is transient zeros, and nothing on our buses sends RTR polls anyway. The missing `count` check is a pre-existing pattern shared with 0x001/0x009/0x01e, not a new exposure.
- **The new +0x014 subscription cannot collide for spec-conforming node ids:** the newly illegal base differences (0x0a/0x0b/0x13) are all non-multiples of 0x20, and no fleet config (wheels 0x000–0x300, tornado 0x400/0x500, CANopen 0x60/0x70, MKS 1/5/37) comes anywhere near a clash.
- **`is_boot_complete` widening changes nothing material:** encoder frames (10 ms default) already unlock the gate long before the first heartbeat, and 0x014 is off by default.
- **Broadcast cost is nil:** broadcast is opt-in, no ODriveMotor runs on an expander in the fleet, and the line stays ~600 bytes under the 1024 limit.

</details>

Hardware verification on ts2 remains the open box from my earlier comment — with that done and the two fixes above, merge away.

evnchn:
**Both fixes pushed.** <details><summary>Details</summary>

- `current`/`current_setpoint` now sign-corrected via `reversed` in `case 0x014`
- `iq_rate_ms` docs carry `(requires version 0.5.6)` note

Optional items (docs reflow, `demarshal_float`) deferred — can fold in if you want.

</details>

[※](https://zauberzeug.github.io/colophon/#m=qwen-code&t=Fixes%20implemented%20and%20pushed%20by%20AI%20agent%20from%20review%20feedback%3B%20reply%20drafted%20by%20AI. "qwen-code: Fixes implemented and pushed by AI agent from review feedback; reply drafted by AI.")

JensOgorek (approved):
### Approved — fixes verified, merge after the ts2 run

Both fixes in `dd99486` check out: the sign correction uses the same `reversed` idiom as the seven existing call sites in this file, and the docs note matches the existing `motor_error` style. CI is green, and the deferred optional items are fine to leave as-is.

One condition remains: the hardware verification on ts2 — next time you're using the stand, set `iq_rate_ms` > 0 on both axes and check that `current`/`current_setpoint` move plausibly under load. With that done, merge away.

falkoschindler:
@JensOgorek @evnchn If I understand correctly, this PR adds more load to the CAN bus for everyone using the ODriveMotor. That's probably fine in most cases, but could cause overload in others. An opt-in might be overkill, but we should make a conscious decision about it.

JensOgorek:
@falkoschindler exactly. Yet the feature is in development for a long time. In our Odrive fork it goes back for almost a year.
Reading the current is useful for Odrive as Delta-Arm and in normal driving situations.

The Odrives also use 1M bit and not the 500k as the alternative newer drives.

But optimizing the reporting might be useful, like expanding a CAN msg, we already send periodically.

falkoschindler:
@JensOgorek I have to admit I couldn't quite follow your reply, so let me untangle it — and partly answer myself.

On my own concern first: after a closer look at the diff I see it is already addressed. The Lizard side only subscribes and decodes — the ODrive broadcasts 0x014 only if `<axis>.config.can.iq_rate_ms` is set, and that defaults to 0. So this PR adds no bus load at all until someone consciously enables the message per axis, which is exactly the opt-in decision I was asking for. Fine by me, no changes needed.

On your other points, just to make sure I read them right: the feature's age and usefulness were never in question. And I don't see how the bitrate helps the argument — Lizard is a library, users run it on all kinds of buses, and if the newer drives run at 500 kbit/s instead of 1 Mbit/s, they would be affected more by extra traffic, not less. Or did you mean something else?

Lastly, by "expanding a CAN msg we already send periodically" — do you mean packing Iq into a frame the (forked) firmware already broadcasts periodically, instead of enabling a separate Get_Iq message? That would be a firmware-side optimization independent of this PR, right?

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Substance%20from%20the%20author%27s%20objections%20plus%20the%20model%27s%20code%20check%3B%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Substance from the author's objections plus the model's code check; wording by the model.")

JensOgorek:
@falkoschindler Yes — `iq_rate_ms` defaults to 0, so nothing is sent until someone enables it per axis. My "exactly" meant "agreed, it should be a conscious decision", not that the PR adds load.

The bitrate remark was narrower than it sounded: the buses that would carry these frames are our ODrive buses at 1 Mbit. And yes, packing Iq into an existing periodic frame would be a change in our fork, not in Lizard.

PR #268, opened 2026-08-12
Report the project name from core.version()

falkoschindler:
### Motivation

A robot cannot tell which firmware project it is running: `core.version()` reports only `esp_app_desc_t.version`, so integrator builds like [lizard-zz](https://github.com/zauberzeug/lizard-zz) look exactly like public Lizard. zauberzeug/lizard-zz#15 tried to encode the difference into the version string itself (`zz-v0.0.1`), which broke RoSys' version parsing and mixes two concepts into one field — the project's identity and its version. ESP-IDF already has a dedicated field for the identity, `esp_app_desc_t.project_name`; it is just neither settable by integrators nor reported anywhere.

### Implementation

- `CMakeLists.txt` lets integrators override the project name, e.g. `idf.py -DLIZARD_PROJECT_NAME=lizard-zz build` (default: `lizard`, nothing changes for this repo). ESP-IDF embeds the name in `esp_app_desc_t.project_name` and names the build artifacts after it (`lizard-zz.bin`, `lizard-zz.elf`).
- `core.version()` reports the project name in front of the version, e.g. `version: lizard v0.13.0`. The version stays the last token, so already-deployed RoSys installations (`response.split()[-1]`) parse the new format unchanged; `p0.version()` inherits the format via the expander proxy.
- `core.info()` gains a `project name: …` line.
- `espresso.py` reads `build/project_description.json` (`app_bin`/`app_elf`, rewritten by `idf.py` on every configure), so renamed integrator artifacts are found without `--firmware`/`--elf`; under `--host` the resolved paths are appended to the remote command, which has no build directory of its own.
- `compile.sh` runs the container as the calling user (`--user`, `HOME=/tmp`) so that git version detection works and the build artifacts are not root-owned. Linux users with a `build/` from the previous root-running script need a one-time `sudo rm -rf build`.

What consumers see:

| Firmware | `core.version()` | deployed RoSys parser yields |
| --- | --- | --- |
| public Lizard, before this PR | `version: v0.13.0` | `0.13.0` |
| public Lizard, after this PR | `version: lizard v0.14.0` | `0.14.0` |
| lizard-zz | `version: lizard-zz v0.0.2` | `0.0.2` |

Follow-ups: zauberzeug/rosys#452 parses the project name and shows it in the UI; lizard-zz switches from the version marker to `-DLIZARD_PROJECT_NAME` (zauberzeug/lizard-zz#15).

### Progress

- [ ] Tested on hardware (esp32 build with `-DLIZARD_PROJECT_NAME=lizard-zz` passes; `project_description.json` and the app descriptor in `lizard-zz.bin` carry the name and `espresso.py --host --dry-run` forwards the renamed artifact; not flashed to a board yet).
- [x] Documentation has been updated (module reference).

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Analysis%2C%20implementation%20and%20wording%20by%20the%20model%3B%20direction%20and%20decision%20from%20the%20maintainers%27%20discussion%20on%20lizard-zz%2315.&a=claude-code "claude-fable-5: Analysis, implementation and wording by the model; direction and decision from the maintainers' discussion on lizard-zz#15.")


evnchn:

**No blockers from my side.**

One note in case someone later reads `resolve_default_artifact` and files the fallback as a silent-failure bug — it isn't, and making it raise would be a regression.

<details><summary><b>Why handing back a non-existent default is the right behaviour</b></summary>

`resolve_default_artifact` returns `default` unchanged when the glob finds ≠ 1 candidate, i.e. it can return a path that does not exist. The tempting "fix" is to raise there instead. That breaks things:

- It runs for **every** local command (`espresso.py:607-608`), including the ones that never read the artifact — so a machine without a `build/` directory would start failing `enable` / `disable` / `reset`, which have nothing to do with the firmware or the ELF.
- The missing-file case is already handled where the file is actually needed: `espresso.py:341` raises `Missing build artifacts: …; run the ESP-IDF build first.` on the `--host` path, and esptool reports it on the local one.

Checked against this branch, in a scratch directory holding two stale `.elf` files and no `.bin` at all:

```console
$ python3 espresso.py enable --dry-run
Enabling the microcontroller...
Configuring EN and G0 pins...
  Setting G0 pin to 1
  Setting EN pin to 1
Finished. ☕️
# exit 0

$ python3 espresso_raise.py enable --dry-run   # same file, but ≠ 1 match raises
Espresso dry-running...
Error: cannot resolve build/lizard.bin: 0 candidates
# exit 1
```

Best-effort resolution plus lazy validation at the point of use is the correct split. Nothing to change.

</details>

ftilde:
Required some changes, but seems to work now. Maybe you can have another look @falkoschindler and merge if you don't have any complaints.

ftilde (approved):

falkoschindler:
A review pass found three real issues, all fixed in 617df45:

- `flash --host` / `coredump --host` always failed for a renamed build: the remote re-resolved the default itself, but only the `.bin` (or only the `.elf`) is rsynced, so the `.bin`/`.elf` pair heuristic could never match there.
- A stale `build/lizard.bin` silently shadowed a fresh `lizard-zz.bin` in the same tree.
- `core.info()` printed `lizard version:` next to a non-lizard project name.

`espresso.py` now reads `build/project_description.json` (`app_bin`/`app_elf`) instead of globbing, and under `--host` appends the resolved `--firmware`/`--elf` to the remote command. `LIZARD_PROJECT_NAME` became a cache variable with a 31-character check (ESP-IDF would truncate it silently), and `docs/tools.md` describes the override.

Verified with an esp32 Docker build using `-DLIZARD_PROJECT_NAME=lizard-zz` (descriptor in `lizard-zz.bin` carries the name, overlong name fails at configure time) and `espresso.py flash/coredump --host --dry-run`. The RoSys `split()[-1]` parser is unaffected.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%2C%20fixes%20and%20wording%20by%20the%20model%3B%20scope%20decisions%20and%20final%20check%20by%20the%20maintainer.&a=claude-code "claude-fable-5: Review, fixes and wording by the model; scope decisions and final check by the maintainer.")

PR #269, opened 2026-08-13
Re-enter boot detection in step() after restart/flash

### LINKED ISSUE #266: Expander is_ready never returns to true after flash(), restart() or disconnect()
*Posted by Claude Code on evnchn's behalf. Hardware observation and reproduction by @JensOgorek on Feldfreund F31 / rb67; mechanism analysis is mine.*

**TL;DR — `Expander::is_ready` is a one-way flag. Only the constructor can ever set it back to `true`, so after `flash()`, `restart()` or `disconnect()` the expander stays `is_ready == false` for the lifetime of the core, and only `core.restart()` recovers it.**

### Observed

While hardware-testing #261 on rb67, @JensOgorek ran a real `p0.flash()` and then checked the flag directly:

> `p0.is_ready` stayed `false` afterwards and only recovered on `core.restart()`.

Same on the disconnect path — which is expected there, but on the **flash** path it is not, because `Expander::flash()` deliberately calls `restart()` at the end of a successful flash specifically to re-enter boot detection.

### Mechanism

`is_ready = true` has exactly one writer, and it is unreachable after startup:

- `main/modules/expander.cpp:93` — `check_boot_progress()` sets `is_ready = true` on receiving `Ready.`. This is the **only** assignment of `true` in the file.
- `main/modules/expander.cpp:66` — `check_boot_progress()` has exactly one caller: the `while` loop inside the **constructor** (`:60-66`), bounded by `boot_timeout`.
- `main/modules/expander.cpp:72` — `Expander::step()` gates *all* of its work on `is_ready`, and never calls `check_boot_progress()`.

So every path that clears the flag is terminal:

- `restart()` — `:128` sets `is_ready = false`, reached from the `restart` method (`:165`) and from the tail of a successful `flash()` (`:200`).
- the ping timeout — `:113`.
- `deinstall()` — after #261, which is the intended behaviour there.

`restart()` sends `core.restart()` to the P0 and resets `boot_start_time`, i.e. it sets up for boot detection that then never runs. The P0 does come back up and does send `Ready.`; nothing is listening for it.

### Why it matters

`is_ready` is a user-visible property, so `p0.is_ready` reads `false` on a healthy, fully-flashed expander. Worse, `step()` gating on it means the module stays inert — no ping, no `handle_messages()` — until the whole core is restarted. After a successful `flash()` that is a surprising place to end up.

### Scope note

This is pre-existing on `main` and is **not** introduced by #261, #262 or #263 — it is called out as a known terminal state in #261's description, and this issue exists because the flash-path half of it was confirmed on hardware and is the case that looks like a genuine bug rather than a design choice. Filing separately so those PRs stay minimal.

Adjacent, from the same session and also worth knowing: `core.get_pin_status()` cannot read back a driven output pin (`InputEn: 0`, so the reported `Level` does not reflect the driven level), which closes the obvious verification route for boot-pin state in #255 / #263.

### Possible directions, not a proposal

- Let `step()` call `check_boot_progress()` while `!is_ready` and a boot is pending, so boot detection is a state rather than a constructor-only phase.
- Or have `restart()` re-enter a bounded boot-detection loop the way the constructor does.

Both change more than a bugfix should without hardware to test on, so I would rather surface the mechanism than push a patch.

evnchn:
### Motivation

`Expander::is_ready` is a one-way flag — only the constructor's boot loop can set it back to `true`. After `restart()` or `flash()`, the expander stays `is_ready == false` for the lifetime of the core, because `step()` gates all work on `is_ready` and never calls `check_boot_progress()`. The P0 does come back up and sends `Ready.`; nothing is listening.

Fixes #266.

### Implementation

Changes to `main/modules/expander.cpp`:

1. **`step()` calls `check_boot_progress()` while `!is_ready`** — bounded by the same `boot_timeout` that the constructor uses. This turns boot detection from a constructor-only phase into a state that `restart()` naturally re-enters by resetting `boot_start_time` and clearing `is_ready`.

2. **Ping timeout closes the boot window** — sets `boot_start_time = 0` alongside `is_ready = false`. Without this, a ping timeout within the boot window (e.g. fast boot + transient ping failure at ~3s with default 5s timeout) would cause `step()` to call `check_boot_progress()` instead of `handle_messages()`, consuming messages without processing them. With `boot_timeout = 0` — the constructor's "wait forever" — the window cannot be closed, so `step()` keeps checking boot progress after a loss and a `Ready.` from a self-rebooting P0 does bring `is_ready` back. That is the intended reading of `boot_timeout = 0`, not an exception to it.

3. **`restart()` flushes the serial input** — while the boot window is closed nothing reads the UART, so the P0's lines accumulate. Without the flush, the next `restart()` hands that backlog to `check_boot_progress()`, which accepts an old `Ready.` shortly after the enable pulse while the P0 is in fact still rebooting.

4. **Ping timeout clears `ping_pending`** — otherwise, once boot detection recovers `is_ready` from `step()`, the next `ping()` takes the pending branch with a stale flag and declares the connection lost again after `ping_interval + ping_timeout` without ever sending a ping.

### Progress

---

Fix authored by Qwen Code from issue analysis by evnchn and hardware reproduction by JensOgorek. [※](https://zauberzeug.github.io/colophon/#m=qwen-code&t=Fix%20authored%20by%20Qwen%20Code%20from%20issue%20analysis%20by%20evnchn%20and%20hardware%20reproduction%20by%20JensOgorek "qwen-code: Fix authored by Qwen Code from issue analysis by evnchn and hardware reproduction by JensOgorek")


evnchn:
@JensOgorek Only the hardware checkbox is left, and you have the P0 rig from the #266 repro — could you run `restart()` and confirm `is_ready` returns to true?

JensOgorek (changes requested):
### Hardware test on F31 passed for the `restart()` path, but two follow-on gaps need fixing before merge: a stale `Ready.` is accepted on the next `restart()`, and a pending ping is carried into the recovered state

Ran the branch with `origin/main` merged (no conflicts) on F31 (rb67: classic-ESP32 core, real P0 on `Serial(26, 27, 115200, 1)` + `Expander(serial, 25, 14)`, production startup script) as the lizard-zz customer build, against the robot's production firmware as the "before". The fix does what #266 asks: after `p0.restart()` the core echoes the P0's boot, sees `Ready.` and `is_ready` is back to `true` after ~2.5 s, where the production firmware stays `false` for good. `disconnect()` stays terminal, as intended. But once a connection loss has happened, the closed boot window leaves the P0's lines unread in the UART buffer, and the next `restart()` consumes that old `Ready.` 180 ms after the enable-pin pulse — `is_ready` goes `true` while the P0 is actually still rebooting. And when boot detection succeeds from `step()` (only reachable with `boot_timeout = 0` after a loss), the `ping_pending` left over from the lost connection makes `ping()` declare the connection lost again `ping_interval + ping_timeout` later without ever sending a ping. Both are small: flush the serial input in `restart()` (or when closing the window), and clear `ping_pending` when the connection is declared lost. With those, I would run the sequence once more and tick the hardware box.

<details>
<summary><b>Setup and sequence</b></summary>

- Firmware: this branch + `origin/main` (`db82c3e`-style local merge, 0 conflicts) built as lizard-zz for esp32, flashed onto the F31 core with its unchanged production script (`4fb4`); "before" = the robot's production build (Lizard `0439c76`). Robot restored afterwards, rosys reconnected (`local = core = 4fb4`).
- Console driven from the Jetson (`/dev/ttyTHS1`), `is_ready` and `last_message_age` polled once per second via `core.print`. The P0 "rebooting behind the core's back" is `p0.run("core.restart()")` — the core sends the restart but does not touch its own state, which stands in for a brown-out.
- Steps: A idle → B `p0.restart()` → C P0 reboots behind the core's back → D `p0.restart()` after that → E `boot_timeout = 0` + P0 reboots behind the core's back → F `p0.disconnect()`. Then, after a core restart, with `ping_interval = 0.3` / `ping_timeout = 0.5` to actually reach the ping timeout: G P0 reboots behind the core's back → H `p0.restart()` → I `boot_timeout = 0` + P0 reboots behind the core's back.

</details>

<details>
<summary><b>Results</b></summary>

| step | production build | this branch |
|---|---|---|
| A idle | `is_ready` true, age 4 ms | true, age 2 ms |
| B `p0.restart()` | false from the pulse on, never recovers (10 s observed) | false → `p0: Ready.` echoed → true after 2.5 s |
| C P0 reboots behind the core's back (default ping 1 s + 2 s) | false (still from B) | stays **true** throughout: the P0 is back in 2.5 s, its boot lines count as messages, no ping ever times out |
| D `p0.restart()` after C | false | false → true after 2.5 s |
| E `boot_timeout = 0`, P0 reboots behind the core's back | false | as C, stays true |
| F `p0.disconnect()` | false, terminal | false, terminal (a later `p0.restart()` hits the deinstalled driver: `uart_write_bytes: uart driver error`, as before) |
| G ping 0.3 s + 0.5 s, P0 reboots behind the core's back | — | `connection lost` 1.9 s after the reboot → false, window closed, the P0's `Ready.` is not read — as designed |
| H `p0.restart()` after G | — | **`p0: Ready.` + `Booting process completed successfully` 180 ms after the pulse** (the unread line from G), true at 0.7 s; the real boot follows (P0 boot warnings at +1.1 s), then `connection lost` at +2 s because the ping fell into the P0's boot → false; the real `Ready.` now sits in the buffer |
| I `boot_timeout = 0` | — | the moment the property is set, the leftover `Ready.` from H is consumed → true |
| I P0 reboots behind the core's back | — | `connection lost` → false → real `Ready.` read (window never closes) → true → **`connection lost` again 0.8 s later with no ping sent** (`ping_pending` still set from the first loss) → false for good |

</details>

<details>
<summary><b>Review notes on the diff</b></summary>

1. **Stale lines survive a closed window** (`step()`): once `boot_start_time = 0` closes the window, nothing reads the UART until the next `restart()`, whose `check_boot_progress()` then sees whatever the P0 sent in the meantime — including an old `Ready.` — before the pulse has even taken effect. `restart()` should flush the serial input (`Serial::flush()` is `uart_flush()`, i.e. input) right after the pulse / `core.restart()` line; flushing when the window is closed in `ping()` would do as well.
2. **`ping_pending` is not cleared on connection loss**, only by `handle_messages()` and `restart()`. Boot detection via `step()` → `check_boot_progress()` sets `is_ready` without touching it, so the next `ping()` call takes the `else` branch with a stale `pending` and declares the connection lost after `ping_interval + ping_timeout` although no ping went out. With default timings that is 3 s of `is_ready == true` followed by a permanent `false` in the `boot_timeout = 0` case. One line in the loss branch (`this->ping_pending = false;`) fixes it.
3. **`boot_timeout = 0` semantics**: the PR text says the ping timeout closes the boot window, but with `boot_timeout = 0` (constructor meaning "wait forever") `step()` keeps calling `check_boot_progress()` after a loss, so `Ready.` from a self-rebooting P0 does bring `is_ready` back — measured in step I. That is arguably the better behaviour; the description should say so rather than claim the window is closed.
4. **Observation, not this PR's problem**: a P0 that reboots faster than `ping_interval + ping_timeout` (2.5 s vs 3 s here) is invisible to the core — `is_ready` stays `true`, but the P0 has lost every proxy module (`battery_control_status`, `vdp_status` etc. stop broadcasting after the reboot in step C). Same before and after, and not a practical concern: the P0 only ever boots after the core, because the core's startup script triggers its restart.
5. `has_buffered_lines()` guards against a deinstalled driver since #261, so calling `check_boot_progress()` from `step()` after `disconnect()` is safe (measured in F).
6. Branch is 13 commits behind `main`, merges without conflicts; the docs need no change for this fix (the `is_ready` row already describes the intended behaviour).

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Test%20plan%2C%20harness%2C%20review%20findings%20and%20wording%20by%20the%20model%3B%20the%20robot%20work%20was%20supervised%20and%20the%20conclusions%20were%20checked%20by%20the%20author.&d=2026-09-01&a=claude-code "claude-fable-5: Test plan, harness, review findings and wording by the model; the robot work was supervised and the conclusions were checked by the author.")

evnchn:
**Both fixed in 3c4ab60, exactly where you pointed.** `restart()` now flushes the serial input after the pulse, and the connection-loss branch clears `ping_pending`. Two lines. Point 3 (the `boot_timeout = 0` semantics) is a wording bug in my description, not in the code — the PR body now says what you measured in step I, that with "wait forever" the window cannot close and a self-rebooting P0's `Ready.` legitimately brings `is_ready` back.

```diff
@@ void Expander::ping() {
              this->properties.at("is_ready")->boolean_value = false;
              this->boot_start_time = 0;
+             this->ping_pending = false;
@@ void Expander::restart() {
          this->serial->write_checked_line("core.restart()");
      }
+     this->serial->flush();
      this->boot_start_time = millis();
```

Thank you for the F31 run — steps H and I are both things I would not have found without hardware.

<details>
<summary><b>Why the flush lands in <code>restart()</code> and not when the window closes</b></summary>

You offered both placements. `restart()` is the one that holds: closing the window in `ping()` only discards what has arrived *by then*, and the P0 keeps talking afterwards — in your step G the reboot lines land after the loss is declared, so flushing at loss time would still leave a `Ready.` in the buffer for step H to consume. Flushing immediately before the new boot window opens covers the whole gap regardless of when the stale lines arrived.

The obvious worry with that placement is the discard window it opens: in the boot-pin path the enable pin is already high when the flush runs, so anything the P0 emits in between is thrown away. It cannot swallow a real `Ready.` — `main.cpp` prints it at line 484, after the `vTaskDelay(1500 / portTICK_PERIOD_MS)` at line 441 plus UART setup and startup restore. The flush is a few microseconds against a floor of 1.5 s. What it does discard is boot ROM noise, which `check_boot_progress()` would only have echoed.

</details>

<details>
<summary><b><code>Serial::flush()</code> semantics, since the name is misleading</b></summary>

`Serial::flush()` is `uart_flush()`, which in ESP-IDF is `__attribute__((alias("uart_flush_input")))` — RX only, so it cannot drop the `core.restart()` line we just wrote on the pin-less path.

It also leaves `has_buffered_lines()` consistent, which matters because that is what gates every read loop in this file. `uart_flush_input` drains the RX ring buffer and calls `uart_pattern_queue_update(size)` per item; that decrements each queued pattern position and advances `rd` past any that go negative, so `uart_pattern_get_pos()` returns `-1` afterwards rather than pointing into drained bytes.

With the driver deinstalled after `disconnect()`, `uart_flush_input` hits its `p_uart_obj[uart_num]` null check, returns `ESP_FAIL` and logs — the same shape as the `uart_write_bytes: uart driver error` that path already produces today, so `restart()`-after-`disconnect()` behaves as it did before.

Verified against `esp_driver_uart/src/uart.c` in ESP-IDF v5.3.1 (`uart_flush` alias at :1548, `uart_flush_input` at :1550, `uart_pattern_queue_update` at :535).

</details>

<details>
<summary><b>What I could and could not verify here</b></summary>

- Builds clean for esp32 with ESP-IDF v5.3.1 (`build.py esp32`, 1161/1161, `lizard.bin` 0x13ba80).
- The ESP-IDF call paths above were read in the installed v5.3.1 source, not inferred from the docs.
- **No hardware.** I have no P0, so steps G→I are yours to re-run — that sequence is exactly what these two lines target and I would not consider them confirmed until it passes.
- Your points 4 (a P0 rebooting faster than `ping_interval + ping_timeout` is invisible to the core) and 5 (`has_buffered_lines()` guarding the deinstalled driver) are left alone, per your own reading that neither belongs to this PR.
- Branch is still behind `main`; happy to merge or rebase if you would rather review it on top.

</details>

JensOgorek (approved):
### Hardware re-test of `3c4ab60` passed: `restart()` after a connection loss no longer accepts a stale `Ready.`, and a recovered expander stays ready (bench core/expander pair, A/B against the branch without the fix)

I re-ran the review sequence A–I on two classic ESP32s wired like a Robot Brain (`Serial(26, 27, 115200, 1)`, `Expander(serial, 25, 14)`), once with this branch minus `3c4ab60` and once with it, both with `origin/main` merged. Without the fix the bench reproduces both F31 failures exactly. With it, `p0.restart()` after a connection loss waits for the real boot (`Ready.` 2.47 s after the pulse, same as a plain restart). With `boot_timeout = 0`, an expander that recovers after a reboot stays ready. Steps A–F are unchanged. Approved from my side, hardware box ticked. @falkoschindler, over to you for the merge review.

<details>

- Firmware: PR head `3c4ab60` + `origin/main` (`28ad386`, merges cleanly) = `v0.14.1-11-g5ff424c`, built for esp32. The "without" variant is the same tree with `3c4ab60` reverted. Only the core firmware differs between the two runs, the expander board runs the fixed build in both.
- Hardware: two ESP32-D0WD-V3, core 27 → expander RX, 26 ← expander TX, 25 → GPIO0, 14 → EN. The expander has an empty startup script and prints `Ready.` about 2.4 s after reset (the F31 P0 took about 2.5 s).
- Modules are created on the core console, `is_ready` is polled every 250 ms via `core.print`. "P0 reboots behind the core's back" is `p0.run("core.restart()")`. G–I run after a core restart with `ping_interval = 0.3` and `ping_timeout = 0.5`.

</details>

<details>
<summary><b>Results (seconds after the step's command)</b></summary>

| step | without `3c4ab60` | with `3c4ab60` |
|---|---|---|
| B `p0.restart()` | `Ready.` at 2.48, true at 2.68 | same |
| C / E P0 reboots behind the core's back (default ping, E with `boot_timeout = 0`) | stays true, no loss | same |
| D `p0.restart()` after C | true at 2.68 | same |
| F `p0.disconnect()` | false, terminal | same |
| G ping 0.3 + 0.5, P0 reboots behind the core's back | `connection lost` at 2.1 → false | same (2.09) |
| H `p0.restart()` after G | **stale `Ready.` 0.14 after the pulse**, true at 0.35, `connection lost` at 1.78 → false | the first lines are the real boot's ROM output (`rst:0x1 (POWERON_RESET)`), `Ready.` at 2.47, true at 2.68, no loss |
| I `boot_timeout = 0` | the leftover `Ready.` is consumed at once → true | nothing to consume |
| I P0 reboots behind the core's back | loss at 0.05 → `Ready.` at 2.39 → true → **`connection lost` again at 3.20** (0.8 later, no ping sent) → false | loss at 1.69 → `Ready.` at 2.39 → true, stays true for the remaining 5.6 |

</details>

<details>
<summary><b><code>restart()</code> after <code>disconnect()</code></b></summary>

With the fixed build, `p0.restart()` after `p0.disconnect()` logs a single `E (…) uart: uart_flush_input(1553): uart driver error`, and nothing floods afterwards. The core keeps running, `core.print` still answers 7 s later with `is_ready` false. That matches the prediction in your comment.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5-5&t=Test%20plan%2C%20harness%2C%20A/B%20run%20and%20wording%20by%20the%20model%3B%20the%20bench%20run%20was%20supervised%20and%20the%20conclusions%20were%20checked%20by%20the%20author.&d=2026-09-23&a=claude-code "claude-opus-5-5: Test plan, harness, A/B run and wording by the model; the bench run was supervised and the conclusions were checked by the author.")

falkoschindler (approved):
Thanks @evnchn, this closes #266 the right way: boot detection is a state the expander can re-enter, not a constructor-only phase. And thanks @JensOgorek for the two hardware rounds, the stale `Ready.` in step H would not have been found otherwise.

I pushed one simplification on top of `3c4ab60` (`1ba3c9c`): `step()` now calls `check_boot_progress()` whenever the expander is not ready, without the `boot_timeout` window, and `ping()` no longer sets `boot_start_time = 0`. The window had no purpose: `handle_messages()` never runs while `is_ready` is false, so there was nothing `check_boot_progress()` could take messages away from. What it did do was make a connection loss terminal under the default `boot_timeout`, let the P0's lines pile up unread (the origin of the stale `Ready.`), and require a sentinel that is not actually closed (`millis_since(0)` is the uptime). With the change, the self-healing that step I showed with `boot_timeout = 0` is the normal behaviour, the `boot_timeout == 0` special case is gone, and `boot_timeout` only bounds the constructor's blocking wait, as the docs say. The flush in `restart()` and the `ping_pending` reset stay as they are. The net diff against `main` is four lines.

@JensOgorek, since this changes what happens after a connection loss, could you re-run G–I once before the merge? Expected: G declares the loss, the P0's `Ready.` brings `is_ready` back without a manual `restart()`, and H/I behave as in your second run.

One thing the PR makes visible but does not cause: after `restart()`/`flash()` the P0 starts with an empty module table, so all proxies are dead while `is_ready` reports `true`. That is pre-existing (your point 4, Jens) and goes to a follow-up issue: #301.

This is an AI-assisted review pass; I did not verify every point myself.

<details>

**1. Boot window in `step()` is unnecessary** main/modules/expander.cpp (75-80), (118-121) — the `boot_timeout` gate and the `boot_start_time = 0` sentinel made a connection loss terminal by default, let unread lines pile up, and did not actually close (uptime, 32-bit `millis()` wrap). Fixed in `1ba3c9c` by listening whenever the expander is not ready. This also removes the duplicated window predicate between constructor and `step()` and keeps `boot_timeout` at its documented meaning, so no doc change is needed.

**2. Proxies are not re-created after the P0 reboots** main/modules/proxy.cpp (16-21) — a proxy sends its definition once in its constructor; after `restart()`/`flash()` the P0 has no modules while `is_ready` is `true`. Pre-existing, follow-up issue #301.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Review%20findings%2C%20the%20simplification%20and%20the%20wording%20by%20the%20model%3B%20the%20decision%20to%20take%20it%20is%20the%20author%27s.&d=2026-09-24&a=claude-code "claude-fable-5-1: Review findings, the simplification and the wording by the model; the decision to take it is the author's.")

JensOgorek:
### Hardware re-test of `1ba3c9c` passed: after a connection loss the expander now comes back on its own under the default `boot_timeout`, and H/I behave as in the second run (bench core/expander pair, A/B against the branch without `1ba3c9c`)

I re-ran A–I on the same bench pair as last time, with this branch + current `main` (`642861f`, merges cleanly), once with and once without `1ba3c9c`. With it, G declares the loss at 1.65 s and the P0's `Ready.` brings `is_ready` back at 2.56 s without a manual `restart()`, where the branch without it stays `false`. H waits for the real boot (ROM banner first, `Ready.` at 2.43 s) and I stays `true` after recovering. A–F are unchanged. Good to merge from my side. @falkoschindler, over to you. The PR description still describes the boot window in points 1–3, which `1ba3c9c` removed.

One finding on the way belongs to #287, not to this PR, but it should be fixed before the next release: since #287 every UART driver is installed without an event queue, and ESP-IDF 5.3.1 then aborts in the UART ISR when a receive ring runs full while a line end arrives. Reproduced on the console (UART0) and, in step J below, on the expander UART of the branch without `1ba3c9c`. Fix follows in #302.

<details>

- Firmware: PR head `1ba3c9c` + `origin/main` (`642861f`) = `v0.14.1-14-g09f9465`, built for esp32. The "without" variant is the same tree with `1ba3c9c` reverted. Core and expander both run the full build, only the core differs between the runs.
- Hardware: the two ESP32-D0WD-V3 from the second run, wired like a Robot Brain (`Serial(26, 27, 115200, 1)`, `Expander(serial, 25, 14)`), empty expander startup, `Ready.` about 2.4 s after reset.
- Modules created on the core console, `is_ready` polled every 250 ms via `core.print`. "P0 reboots behind the core's back" is `p0.run("core.restart()")`. G–I run after a core restart with `ping_interval = 0.3` and `ping_timeout = 0.5`.

</details>

<details>
<summary><b>Results A–I (seconds after the step's command)</b></summary>

| step | without `1ba3c9c` | with `1ba3c9c` |
|---|---|---|
| B `p0.restart()` | `Ready.` at 2.41, true at 2.56 | same (2.40 / 2.55) |
| C / E P0 reboots behind the core's back (default ping, E with `boot_timeout = 0`) | stays true | same |
| D `p0.restart()` after C | true at 2.55 | same |
| F `p0.disconnect()` | false at once, terminal; `restart()` afterwards logs one `uart_flush_input(1553): uart driver error`, nothing floods | same |
| G ping 0.3 + 0.5, P0 reboots behind the core's back, default `boot_timeout` | `connection lost` at 1.60 → **false for good** | `connection lost` at 1.65 → `Ready.` at 2.30 → **true at 2.56**, stays true |
| H `p0.restart()` after G | ROM banner first, `Ready.` at 2.42, true at 2.52 | ROM banner first, `Ready.` at 2.43, true at 2.53, no loss |
| I `boot_timeout = 0`, P0 reboots behind the core's back | loss at 1.64 → `Ready.` at 2.29 → true at 2.54, stays true | same |

</details>

<details>
<summary><b>Extra step J: connection loss while the P0 keeps running</b></summary>

Forced with `ping_interval = 0.3` and `ping_timeout = -0.2`, so the step after a ping declares the loss before `__PONG__` can arrive. Then `p0.run("b = Input(4); b.broadcast()")` makes the still-running P0 send a line per loop.

- With `1ba3c9c`: `is_ready` stays `false` (the P0 never sends `Ready.` again, as before), and `check_boot_progress()` echoes every line, 314 `p0: !!b.active=true;…` lines in 3 s, until `p0.restart()` ends it.
- Without `1ba3c9c`: nothing reads the expander UART after the loss, and 0.75 s after the broadcasts start the core aborts with `assert failed: xQueueGenericSendFromISR queue.c:1180 (pxQueue)` (the #287 issue below).

So listening all the time is also the safer choice. The echo only appears for a P0 that stalls for `ping_interval + ping_timeout` and then carries on without rebooting. I would not change anything for that here.

</details>

<details>
<summary><b>The #287 crash: UART drivers without an event queue</b></summary>

- `install_uart_driver_on_core1()` from #287 calls `uart_driver_install(port, rx, tx, 0, nullptr, 0)`. Before #287, UART0 had an event queue of 20 and `Serial` one of 100 (`UART_PATTERN_QUEUE_SIZE` was passed as the queue size).
- In ESP-IDF 5.3.1 `uart_rx_intr_handler_default()`, the branch for "ring buffer full while a line end was detected" calls `xQueueSendFromISR(p_uart->event_queue, …)` without a NULL check (`uart.c:1160`, the check only follows for the log line), so a missing queue trips `configASSERT(pxQueue)`.
- UART0 on the bench core: block the main loop for 5 s (an `Expander` on unconnected pins) and stream lines into the console at 115200. 6 KB → 140/140 lines processed. 16 KB → assert, then the core hangs while saving the core dump, 0/373 processed.
- Same test with the event queue from #302: 16 KB and 32 KB → no crash, 187 lines (about the 8 KB ring) processed, the rest dropped with a single checksum mismatch. That is the overflow behaviour from before #287.
- Not released yet (#287 is newer than v0.14.1).

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5-5&t=Test%20plan%2C%20harness%2C%20A/B%20run%20and%20wording%20by%20the%20model%3B%20the%20conclusions%20were%20checked%20by%20the%20author.&d=2026-09-24&a=claude-code "claude-opus-5-5: Test plan, harness, A/B run and wording by the model; the conclusions were checked by the author.")

PR #271, opened 2026-08-14
Consolidate the duplicated enabled-sync pattern in Module

### LINKED ISSUE #232: Consolidate the duplicated enabled-sync pattern across motor modules
Follow-up from the #225 review (finding 5, explicitly out of scope for that behavior-preserving split).

#225 gave the enabled-sync pattern (edge-detecting direct writes to the `enabled` property in `step()` plus the dual bookkeeping of member and property in `enable()`/`disable()`) a shared home in the `Wheels` base class — but five motor modules still carry their own copies:

- `d1_motor.cpp` (~line 149)
- `dunker_motor.cpp` (~line 188)
- `linear_motor.cpp` (~line 40)
- `mks_servo_motor.cpp` (~line 186)
- `canopen_motor.cpp` (~line 562)

The copies differ in details, so hoisting into `Module` (or an intermediate base/mixin) needs some design work:

- `CanOpenMotor` guards the sync with its `initialized` flag and performs an extra property write.
- The motor modules gate their `speed`/motion commands on the *member* (one-cycle lag after a direct property write), whereas the wheels modules now deliberately gate on the immediate *property* — consolidation should either reconcile this or keep the gate out of the shared part.

Goal: the next enabled-capable module shouldn't have to add a seventh copy.

JensOgorek:
### Motivation

Closes #232.

The issue lists five motor modules that still carry their own copy of the enabled-sync after #225 gave `Wheels` a shared home for it. It turned out to be worse: **thirteen** modules carry the copy, and `Wheels` is a fourteenth variant under a different name (`last_applied_enabled`). Each copy is a `bool enabled` member, a seven-line edge-detection block in `step()`, and the dual bookkeeping of member and property in `enable()`/`disable()`.

### Implementation

The mechanism moves into `Module`: the `enabled` member, `sync_enabled()` for the edge detection, and virtual `enable()`/`disable()` that do the bookkeeping around two new hooks, `do_enable()`/`do_disable()`. A module now contributes only the hardware-specific action, and calls `this->sync_enabled()` at exactly the position in `step()` where its old block sat — the position matters, since several modules do work before or after it.

Both hooks run while `enabled` still holds the old value, and the flags are committed only after the hook returned. So `do_disable()` can still use methods that are gated on `enabled` — `RoboClawMotor::speed()` and `Output::deactivate()` do — while `do_enable()` cannot; none does. The payoff of committing late is that a hook which throws (an SDO write that exhausts its retries, a RoboClaw serial timeout) leaves the transition pending, and `sync_enabled()` retries it on the next cycle — the behaviour the per-module copies had on `main`.

The `enabled` property is resolved once and cached instead of being looked up twice per cycle in every enabled-capable module. A module that calls `enable()`/`disable()` without declaring the property gets a `runtime_error`, which the main loop reports as "error in module", instead of a `std::out_of_range` that escaped every catch and rebooted the Core.

The two design questions the issue raises resolve without special-casing. `CanOpenMotor` keeps its `initialized` guard (it simply calls `sync_enabled()` after the early return) and its extra property write (now in `do_enable()`). And the gate stays out of the shared part, so motor modules keep gating on the member while `Wheels` keeps gating on the property via `may_drive()` — the one-cycle difference between the two is preserved rather than reconciled.

`Motor` no longer declares `enable()`/`disable()`. Every `Motor` is also a `Module` with an `enabled` property, and `MotorAxis` was the only caller; it now keeps the motor as a `Module_ptr` alongside the `Motor_ptr` and calls through that. Without this, `StepperMotor`, `ODriveMotor` and `CanOpenMotor` each needed a forwarder to `Module::enable()` to satisfy both bases.

Net effect: 269 lines removed, 148 added.

### A crash the refactor uncovered

Tracing the ordering constraints turned up a pre-existing crash, present on `main` just as much as here: writing `enabled = false` on an `Output` module panics the Core. `Output::deactivate()` ended with `this->step()`, and `disable()` calls it while `enabled` is still set, so the nested `step()` saw the same pending transition and called `disable()` again, without bound. The method call `o.disable()` was safe — it writes the property only after `deactivate()` has returned.

The first version of this PR masked that with an in-flight flag in `Module::sync_enabled()`. The review pointed at the cause instead: `on()`, `off()`, `level()`, `activate()` and `deactivate()` all used `step()` as an apply-now entry point. The pin write and the `level`/`change` bookkeeping now live in `Output::apply_level()`, which the command paths and `step()` both call; nothing re-enters `step()` any more, and the guard is gone.

On the bench S3 the crash was deterministic while the exception flavour varied between runs (LoadProhibited, IntegerDivideByZero in ISR context, `task_wdt: task not found`) — what a smashed stack looks like. The same assignment from a rule in the startup script boot-looped the board (7 boots / 5 panics in 15 s; recovery via `esptool.py erase_region 0x9000 0x6000`), 0 / 0 with the guard. `PwmOutput` is the control that isolates the cause: identical enabled-sync, but no `step()` call in its disable path, and it never crashed.

### Verification on F31

With `origin/main` merged in (no conflicts, `db82c3e`), the branch was built as the customer firmware — lizard-zz `main` with this Lizard commit as the submodule — and run on F31 (rb67, classic ESP32 core) under its unchanged production startup script (checksum `4fb4`: `Output` ×2, `Bluetooth`, `Can`, two `InnotronicDriveMotor`, `InnotronicWheels`, `Expander`, the estop rules). Everything the script does through `enable()`/`disable()` still reaches the motors, and the `enabled = false` crash on an `Output` is reproduced on the robot's production firmware and gone with this branch. The robot was restored afterwards (production build, `4fb4`, rosys reconnected). No driving involved — the wheels were only enabled and disabled.

This run used the guard version (`db82c3e`). The three review follow-up commits on top (`56f7d7b`, `94e6b4b`, `eb3ba59`) are build-tested for esp32 only; the `Output` sequence below should be repeated once before merging.

<details>
<summary><b>Results on F31</b></summary>

`Output` on a spare pin (`t = Output(4)`), commands sent over the console, one print per step:

| step | production build (Lizard `0439c76`) | this branch + main |
|---|---|---|
| `t.on()` | level 1, enabled true | level 1, enabled true |
| `t.enabled = false` | **LoadProhibited panic, core dump, reboot** (`PRO CPU has been reset by WDT`) | level 0, enabled false, active false |
| `t.on()` while disabled | — | level stays 0 |
| `t.enabled = true` + `t.on()` | — | level 1 |
| `t.disable()` / `t.enable()` + `t.on()` | — | level 0 / level 1 |

`InnotronicWheels` (the lizard-zz drivetrain on the shared `Wheels` base), read back as `wheels.enabled left.enabled right.enabled`:

| step | result |
|---|---|
| `wheels.disable()` | false false false |
| `wheels.enable()` | true true true |
| `wheels.enabled = false` (property write, synced by `Module::sync_enabled()` → `Wheels` → `do_disable()`) | false false false |
| `wheels.enabled = true` | true true true |
| script `disable()` / `enable()` | re-enabled within one cycle by the script's own `when disabled == true and estop_active == false … then enable()` rule, as designed |
| `left.enabled = false` (motor's own sync, untouched by this PR) | true false true |

No warning or error from any module during the runs; heap unchanged across the sequence (98 752 → 97 720 B with the extra `Output` alive). lizard-zz esp32 binary: 1 382 544 B on lizard-zz `main` vs 1 378 832 B with this branch (−3 712 B), matching the −3 840 B measured for the upstream esp32s3 build.

</details>

lizard-zz (`main`, b162d93) builds cleanly against this branch, including the review follow-up commits. Its private modules carry the same copy three more times (`RmdMotorBase` with its own `bool enabled`, `InnotronicMotorBase` and `RmdDeltaArm` with `last_applied_enabled`); their `enable()`/`disable()` silently become overrides of the new virtuals and keep working as before. `InnotronicDriveMotor` is the one private `Motor`: its `enable()`/`disable()` forwarders still compile because `override` now resolves to `InnotronicMotorBase`'s virtuals, but their comment ("Satisfy Motor's pure-virtual enable()/disable()") is stale and the forwarders are redundant. No private `do_enable()` uses a method gated on `enabled`. Removing the forwarders and migrating the three copies to `sync_enabled()`/`do_enable()`/`do_disable()` is a lizard-zz follow-up, not part of this PR.

### Progress

- [x] Tested on hardware: guard version verified on F31, `eb3ba59` re-verified on a bench ESP32.


main/modules/module.cpp
```diff
@@ -71,6 +71,35 @@ void Module::step() {
...
+    }
+    if (this->properties.at("enabled")->boolean_value != this->enabled) {
+        this->syncing_enabled = true;
+        if (this->properties.at("enabled")->boolean_value) {
+            this->enable();
+        } else {
+            this->disable();
+        }
+        this->syncing_enabled = false;
+    }
```
evnchn, line 88: **`syncing_enabled` is not exception-safe.** Consider the following:

1. `wheels.enabled = false` → `Wheels::step()` → `sync_enabled()` sets `syncing_enabled = true`.
2. `RoboClawWheels::do_disable()` → `left_motor->disable()` → `RoboClawMotor::do_disable()` → [`speed(0)`](https://github.com/zauberzeug/lizard/blob/db82c3e958c621d522413c5675f823d5e8de0f75/main/modules/roboclaw_motor.cpp#L100), which throws after four failed retries.
3. The throw unwinds past `this->syncing_enabled = false`, and [`run_step()`](https://github.com/zauberzeug/lizard/blob/eb49587bcab85125ceced7908638f7c5145290fe/main/main.cpp#L431-L438) catches and echoes it.
4. The flag is still set, so every later `sync_enabled()` returns at the guard: `wheels.enabled` is no longer applied by the step sync for the lifetime of that module. Explicit `wheels.disable()` still works — only the property path dies.

On `main` step 3 costs one cycle and the next `step()` retries.

```suggestion
    if (this->properties.at("enabled")->boolean_value != this->enabled) {
        this->syncing_enabled = true;
        try {
            if (this->properties.at("enabled")->boolean_value) {
                this->enable();
            } else {
                this->disable();
            }
        } catch (...) {
            this->syncing_enabled = false;
            throw;
        }
        this->syncing_enabled = false;
    }
```

<details>
<summary>Runnable repro — host compiler, no ESP-IDF</summary>

`sync_enabled()`/`enable()`/`disable()` copied verbatim from db82c3e; everything else is scaffolding, so this exercises the function, not the real module.

```cpp
// Module::sync_enabled() from PR #271 (db82c3e), copied verbatim; everything
// else is the smallest scaffold that lets it run on a host compiler.
// c++ -std=c++17 mre.cpp -o mre && ./mre
#include <cassert>
#include <stdexcept>
#include <cstdio>

#ifndef FIXED
#define FIXED 0 // -DFIXED=1 for the proposed try/catch
#endif

struct Module {
    bool prop_enabled = true;   // properties.at("enabled")->boolean_value
    bool enabled = true;
    bool syncing_enabled = false;
    virtual void do_disable() {}

    void sync_enabled() {
        if (this->syncing_enabled) {
            return;
        }
        if (this->prop_enabled != this->enabled) {
            this->syncing_enabled = true;
#if FIXED
            try {
#endif
                if (this->prop_enabled) {
                    this->enable();
                } else {
                    this->disable();
                }
#if FIXED
            } catch (...) {
                this->syncing_enabled = false;
                throw;
            }
#endif
            this->syncing_enabled = false;
        }
    }
    void enable() {
        this->enabled = true;
        this->prop_enabled = true;
    }
    void disable() {
        this->do_disable();
        this->enabled = false;
        this->prop_enabled = false;
    }
};

// RoboClawMotor::do_disable() -> speed(0) -> throws after 4 failed retries.
struct FlakyMotor : Module {
    int fails_left;
    FlakyMotor(int n) : fails_left(n) {}
    void do_disable() override {
        if (this->fails_left-- > 0) {
            throw std::runtime_error("could not set speed after 4 retries");
        }
    }
};

// run_step(): catches, echoes, next cycle retries.
static void step(Module &m) {
    try {
        m.sync_enabled();
        printf("error in module \"m\": %s\n", e.what());
    }
}

int main() {
    FlakyMotor m(1); // one transient failure, then the hardware answers
    m.prop_enabled = false;
    for (int i = 0; i < 10; ++i) {
        step(m);
    }
    printf("prop_enabled=%d enabled=%d syncing_enabled=%d\n",
           m.prop_enabled, m.enabled, m.syncing_enabled);
    assert(!m.enabled && "disable never applied: sync is wedged");
    printf("ok\n");
}
```

```console
$ c++ -std=c++17 mre.cpp -o mre && ./mre
error in module "m": could not set speed after 4 retries
prop_enabled=0 enabled=1 syncing_enabled=1
Assertion failed: (!m.enabled && "disable never applied: sync is wedged")

$ c++ -std=c++17 -DFIXED=1 mre.cpp -o mre && ./mre
error in module "m": could not set speed after 4 retries
prop_enabled=0 enabled=0 syncing_enabled=0
ok
```
</details>

evnchn:
After this PR `enable()`/`disable()` are virtual on eleven modules where they were not (`Output`, `PwmOutput`, `RmdMotor`, `RoboClawMotor`, `LinearMotor`, `D1Motor`, `DunkerMotor`, `MksServoMotor`, `MotorAxis`, `RmdPair`, `Wheels`), so an out-of-tree subclass with same-signature methods now becomes the final overrider for calls made through the base — worth a check against lizard-zz, which compiles either way.

Super-TLDR: double-check if this is breaking change.

main/modules/module.cpp
evnchn: TL-DR for the diff: squint your eyes and it is mostly just adding try-catch...

(sadly cannot do whitespace-only diff for suggestion block)

falkoschindler:
@JensOgorek I ran a code review over this branch and fixed the findings in place instead of doing a review round-trip. Three commits on top of your merge commit:

- 56f7d7b: `Output` no longer re-enters `step()` from `on()`/`off()`/`level()`/`activate()`/`deactivate()`. The pin write and the `level`/`change` bookkeeping moved into `apply_level()`, which fixes the `enabled = false` crash at its cause, so the in-flight guard in `Module::sync_enabled()` is gone again.
- 94e6b4b: `Module::enable()` commits the `enabled` member and property only after `do_enable()` returned, mirroring `disable()`. A hook that throws (SDO retries exhausted, RoboClaw serial timeout) now leaves the transition pending and `sync_enabled()` retries it on the next cycle, which is what the per-module copies did on `main`. The `enabled` property is resolved once and cached, a module without it gets a `runtime_error` instead of an uncaught `std::out_of_range`, and `Wheels::disable()` clears `holding` before the motors go down.
- eb3ba59: `Motor` no longer declares `enable()`/`disable()`. `MotorAxis` keeps the motor as a `Module_ptr` next to the `Motor_ptr` and calls through that, so the three forwarders in `StepperMotor`, `ODriveMotor` and `CanOpenMotor` are gone.

I rewrote the PR body to match. lizard-zz `main` builds cleanly against this branch; the details are in the body.

What I could not do is the hardware check: your F31 run was on the guard version (db82c3e), and the three commits above are only build-tested. Could you repeat the `Output` sequence from the PR body on a bench board, in particular `t.enabled = false` must not panic, and tick the hardware checkbox if it passes? I will approve after that.

The lizard-zz side (drop the now-redundant `InnotronicDriveMotor` forwarders, migrate the three private enabled-sync copies to `sync_enabled()`/`do_enable()`/`do_disable()`) needs the submodule bumped past this merge first, so that is a follow-up for after this PR lands. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Written%20by%20the%20model%20from%20its%20own%20review%20of%20the%20PR%20and%20the%20fix%20commits%20it%20made%3B%20the%20request%20to%20re-test%20and%20the%20open%20follow-up%20decision%20are%20the%20reviewer%27s.&d=2026-09-21&a=claude-code "claude-fable-5-1: Written by the model from its own review of the PR and the fix commits it made; the request to re-test and the open follow-up decision are the reviewer's.")

main/modules/module.cpp
falkoschindler: Thanks @evnchn, this was real and your repro matches what I found independently. It is resolved by 56f7d7b in a different way than the suggestion: the guard is gone altogether. The re-entry came from `Output` using `step()` as an apply-now entry point in `on()`/`off()`/`level()`/`activate()`/`deactivate()`; those now call `apply_level()` instead, so nothing re-enters `step()` and `syncing_enabled` no longer exists. On top of that, 94e6b4b commits the `enabled` flags only after the hook returned, so a throwing `do_disable()` (your RoboClaw case) leaves the transition pending and the next cycle retries, as on `main`. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Wording%20from%20the%20model%3B%20the%20fix%20it%20refers%20to%20was%20made%20by%20the%20model%20in%20this%20PR%2C%20the%20decision%20to%20close%20the%20thread%20this%20way%20is%20the%20reviewer%27s.&d=2026-09-21&a=claude-code "claude-fable-5-1: Wording from the model; the fix it refers to was made by the model in this PR, the decision to close the thread this way is the reviewer's.")

falkoschindler:
@evnchn Checked against lizard-zz `main` (b162d93), which builds cleanly against this branch. Three private modules (`InnotronicMotorBase`, `RmdMotorBase`, `RmdDeltaArm`) have same-signature `enable()`/`disable()` and do become the final overriders, but they keep their own bookkeeping and never relied on the base, so behaviour is unchanged; the one private `Motor` (`InnotronicDriveMotor`) keeps compiling after eb3ba59 dropped `enable()`/`disable()` from `Motor`, its forwarders just became redundant. Details are in the PR description, and the migration is tracked in lizard-zz#32. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Wording%20from%20the%20model%2C%20based%20on%20its%20own%20build%20check%20of%20lizard-zz%20against%20this%20branch.&d=2026-09-21&a=claude-code "claude-fable-5-1: Wording from the model, based on its own build check of lizard-zz against this branch.")

JensOgorek:
### Hardware re-test of `eb3ba59` passed: the `Output` sequence runs clean on a bench ESP32, `t.enabled = false` no longer panics

Built the branch tip (`eb3ba59`, all three follow-up commits) for esp32 and ran it on a bench ESP32-D0WD with an empty startup script. The full `Output` sequence from the PR body passes over the console, plus the rule variant in the startup script that boot-looped the guard-less version: one boot, zero panics across a `core.restart()`. No warning or error from any module during the runs. Ticking the hardware checkbox; from my side this is ready for your approval.

<details>
<summary>Console sequence on <code>t = Output(4)</code>, read back as <code>level enabled active</code></summary>

| step | result |
|---|---|
| `t.on()` | 1 true false |
| `t.enabled = false` | 0 false false, no panic |
| `t.on()` while disabled | stays 0 false false |
| `t.enabled = true` + `t.on()` | 1 true false |
| `t.disable()` / `t.enable()` + `t.on()` | 0 false false / 1 true false |
| `t.pulse(0.2)` then `t.enabled = false` | pulsing stops, 0 false false |
| `t.activate()` then `t.enabled = false` / `= true` | 1 true true, then 0 false false, then 0 true false |
| `t.on(); t.enabled = false` in one line | 0 false false |

</details>

<details>
<summary>Rule in the startup script</summary>

Startup: `t = Output(4)`, `when t.level == 1 then t.enabled = false; end`, `t.on()`, then `core.restart()`. 15 s watched: 1 × `Ready.`, 0 panics, only the `SW_CPU_RESET` from the restart itself. Afterwards `t.on()` leaves level 0 (disabled), `t.enabled = true` re-enables, the next `t.on()` is disabled again by the rule within the same cycle, as designed. Board restored to its previous firmware and empty NVS afterwards.

</details>

Verified and tested with Claude Fable 5.1; bench ESP32-D0WD via CP2102N console, esp32 build of `eb3ba59`.

falkoschindler (approved):

PR #272, opened 2026-08-17
Fix the parse heap guard rejecting valid startup scripts

ftilde:
### Motivation

The pre-parse heap guard estimates owl's peak allocation as 4 KB plus 64 bytes per character, but owl's cost is dominated by fixed-size buffers rather than by line length.
A 3906-character startup script — an ordinary rosys robot configuration — is therefore rejected with `error: not enough free memory to parse (159072 free, 254080 required)` although the parse peaks at ~14 KB, so the machine never applies its startup script.
In the other direction the contiguous-block check is a flat 4096 bytes while the parse tree grows past that, so the `abort()` inside owl that the guard exists to prevent stays reachable for long lines.

### Implementation

- Add `owl_parse_heap_requirement()` and `owl_parse_block_requirement()` to `parser.c`, where `sizeof(struct owl_token_run)` and `OWL_TOKEN_RUN_LENGTH` are visible, declared in a new `parser_heap.h` (`main.cpp` only sees the public half of the generated `parser.h`)
- Derive the total from owl's allocation shape: one `owl_token_run` per `OWL_TOKEN_RUN_LENGTH` tokens, all kept alive until the tree is destroyed, plus the varint parse tree, whose `1.5x` growth in `grow_tree` keeps an old and a new buffer alive across each realloc
- Scale the contiguous-block requirement with the line length, since the parse tree overtakes a token run as the largest single allocation
- Drop `PARSE_HEAP_BASE` and `PARSE_HEAP_PER_CHAR` from `main.cpp`
- Update `docs/machine_safety.md`, which described the requirement as scaling with the line length only

Measured with the generated parser in a host harness with wrapped `malloc`, at `OWL_TOKEN_RUN_LENGTH=256`, parsing prefixes of a real 3906-character rosys startup script:

| chars | peak | largest block | required before | required after |
| ----- | ---- | ------------- | --------------- | -------------- |
| 129 | 2,933 | 2,568 | 12,352 | 3,600 |
| 936 | 6,259 | 3,594 | 64,000 | 17,760 |
| 3906 | 13,685 | 8,089 | 254,080 | 72,336 |

The measured intercept of 2,565 bytes matches `sizeof(struct owl_token_run)` = `10 * OWL_TOKEN_RUN_LENGTH + 8` = 2,568 bytes on the ESP32, and the largest-block progression 2,568 → 3,594 → 5,392 → 8,089 matches `grow_tree`'s `(n + 1) * 3 / 2`.
The new bounds stay above every measured peak, by 1.2x at 129 characters and 5.3x at 3906.

### Work In Progress

- Constants come from a host harness; not yet confirmed on hardware via `core.heap` around a parse
- Bounding token runs by one token per character over-estimates long lines by ~5x

### Progress

---
⬛ claude-opus-5

### DIFF
--- a/docs/machine_safety.md
+++ b/docs/machine_safety.md
@@ -37,7 +37,7 @@ when core.last_message_age > 500 then motor.stop(); core.clear_schedule() end
 
 ## Low memory
 
-Parsing a line requires a certain amount of free heap memory, which scales with the length of the line.
+Parsing a line requires free heap memory: a fixed amount for the parser's token buffers, plus an amount that grows with the length of the line.
 When there is not enough free or contiguous memory, Lizard drops the line and responds with `error: not enough free memory to parse (<free> free, <required> required)` instead of risking a crash and reboot.
 Like a line with an incorrect checksum, the command was not executed;
 the host system should react accordingly, for example by re-sending the command later or stopping the machine.
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -17,6 +17,7 @@
 #include "modules/expander.h"
 #include "modules/module.h"
 #include "nvs_flash.h"
+#include "parser_heap.h"
 #include "proxy.h"
 #include "rom/gpio.h"
 #include "rom/uart.h"
@@ -307,11 +308,6 @@ void process_tree(owl_tree *const tree, bool from_expander) {
     }
 }
 
-// owl's peak transient allocation scales with the token count of the line, so the
-// required-heap floor scales with its length (base covers a short line's fixed cost).
-static constexpr size_t PARSE_HEAP_BASE = 4096;
-static constexpr size_t PARSE_HEAP_PER_CHAR = 64;
-
 void process_lizard(const char *line, bool trigger_keep_alive, bool from_expander) {
     InterpreterLock lock;
     if (trigger_keep_alive) {
@@ -319,11 +315,13 @@ void process_lizard(const char *line, bool trigger_keep_alive, bool from_expande
     }
 
     // owl's generated parser can abort()/deref on a failed allocation instead of returning an error, so require both
-    // enough total heap (scaled by the line's length) and a large enough contiguous block before parsing. Other tasks
-    // may allocate between the check and the parse; the padded constants absorb that. Drop the line rather than reboot.
-    const size_t required_heap = PARSE_HEAP_BASE + PARSE_HEAP_PER_CHAR * strlen(line);
+    // enough total heap and a contiguous block for its largest buffer before parsing. Other tasks may allocate
+    // between the check and the parse; the padded requirements absorb that. Drop the line rather than reboot.
+    const size_t length = strlen(line);
+    const size_t required_heap = owl_parse_heap_requirement(length);
     const size_t free_heap = xPortGetFreeHeapSize();
-    if (free_heap < required_heap || heap_caps_get_largest_free_block(MALLOC_CAP_8BIT) < PARSE_HEAP_BASE) {
+    const size_t required_block = owl_parse_block_requirement(length);
+    if (free_heap < required_heap || heap_caps_get_largest_free_block(MALLOC_CAP_8BIT) < required_block) {
         echo("error: not enough free memory to parse (%u free, %u required)", (unsigned)free_heap, (unsigned)required_heap);
         return;
     }
--- a/main/parser.c
+++ b/main/parser.c
@@ -1,2 +1,20 @@
 #define OWL_PARSER_IMPLEMENTATION
 #include "parser.h"
+
+#include "parser_heap.h"
+
+// owl keeps one owl_token_run alive per OWL_TOKEN_RUN_LENGTH tokens until the tree is destroyed, and grows the
+// varint parse tree by 1.5x, so an old and a new buffer coexist during each realloc. Lizard's grammar measures
+// at ~2 bytes of tree per character and ~3 bytes of total peak; both bounds below leave room above that.
+#define PARSE_HEAP_PER_CHAR 8
+#define PARSE_TREE_PER_CHAR 4
+
+size_t owl_parse_heap_requirement(size_t line_length) {
+    const size_t runs = 1 + line_length / OWL_TOKEN_RUN_LENGTH;
+    return runs * sizeof(struct owl_token_run) + PARSE_HEAP_PER_CHAR * line_length;
+}
+
+size_t owl_parse_block_requirement(size_t line_length) {
+    const size_t tree = PARSE_TREE_PER_CHAR * line_length;
+    return tree > sizeof(struct owl_token_run) ? tree : sizeof(struct owl_token_run);
+}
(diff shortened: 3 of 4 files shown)


ftilde:
Note that I've not verified the calculations by hand (that's claude's work) but according to the measurements it's an improvement. And it makes lizard work on the robots again :grimacing:

evnchn:
@ftilde is it tested in hardware or you deem it unnecessary?

Since this would work with bare ESP32 I can help test (react 👍) or if you don't need (then react 👀).

ftilde:
@evnchn I tested in on a robot. Main doesn't run on the feldfreund due to lizard massively overestimating the heap requirements. This PR reduces the overestimation so that it doesn't refuse to apply the lizard configuration anymore.

JensOgorek:
### Hardware verification passed — the fix loads a real startup script the old guard rejects outright; two calibration findings for the constants

I reproduced both sides on a bench ESP32 (classic, `esp32` target) with the full F27 startup script — Bluetooth, serial bus coordinator, expander, ~3.2 KB total with a ~460-character `core.output` line. On current `main`, the boot-time parse of the stored startup is rejected as one block: `error: not enough free memory to parse (145272 free, 209088 required)` — the old guard demands 209 KB for a 3.2 KB script, more RAM than the ESP32 ever has, so the robot can never boot its config. On this PR's head (483b37b), the same script parses and boots cleanly, with BT down to 13.6 KB free heap. That confirms the false-rejection fix on real hardware.

Two findings from probing the new constants at low heap, details in the blocks below:

1. **Token-dense lines still hard-crash instead of being dropped.** At 52 KB free, `core.print(1+1+…+1)` (594 chars) passes the new guard but owl `abort()`s in `construct_node_alloc` — the parse tree for expression chains costs roughly 400–500 bytes per character, far above `PARSE_HEAP_PER_CHAR = 8`. The old guard rejected such lines by accident; the new one lets them through into the abort.
2. **The guard needs an absolute reserve for the BT controller.** Even when owl itself survives (a 103-char `1+…` chain), the transient allocation spike starves the Bluetooth controller: deterministic `LoadProhibited` crash in `hli_c_handler`/`r_rwbtdm_isr_wrapper` (BT high-level interrupt, null-deref after a failed alloc). A guard that only accounts for owl's own need can pass a line whose parse kills the radio. Suggestion: require `free_heap > required_heap + RESERVE` with a reserve in the 16–24 KB range when BT is active.

Neither finding blocks this PR for the fleet's real scripts (rosys lines are token-sparse; the F27 script parses fine even at 13.6 KB free), and heap relief for F27 lands separately (#273, #274). But the guard's promise in `machine_safety.md` — drop the line rather than crash — currently doesn't hold for expression-dense lines at low heap.

<details>
<summary>Test setup and measurements</summary>

- Bench: classic ESP32 (ESP32-D0WD-V3) on /dev/ttyUSB0, no expander/IMU/plexus attached; F27 startup generated from the robot's config (IMU line and its 5 output fields removed — no BNO055 on the bench).
- Firmware builds: lizard-zz core build (private Innotronic modules) with the `lizard` submodule at (a) v0.12.0-7-g0439c76 (no guard, F27's current firmware), (b) `main` 468d6c8 (old guard), (c) this PR's head 483b37b.
- (a) no guard: script loads; 72.2 KB free without BT, 13.7 KB with BT (BT costs 58.5 KB — matches the robot measurement from 2026-08-17).
- (b) old guard: whole stored startup rejected at boot, `209088 required` vs. `145272 free`; ESP stays up but unconfigured ("restoring serial bus from backup", `Ready.`).
- (c) this PR: script loads with BT, 13.6 KB free; re-sending the 460-char `core.output` line at runtime parses fine.

</details>

<details>
<summary>Finding 1: owl abort on token-dense lines</summary>

At 52.7 KB free (after the queue/BLE heap fixes), sending `core.print(` + `1+`*290 + `1)` (594 chars) aborts:

```
abort() was called at PC 0x40114b5b
0x40114b5b: construct_node_alloc  parser.h:2228   (calloc returned NULL -> abort)
0x4011507c: apply_actions         parser.h:7870
0x401151fe: build_parse_tree      parser.h:7905
0x400e7a91: process_lizard        main.cpp:334
```

Each token descends the expression-grammar cascade and allocates a `construct_node` with a per-rule slot array, so dense arithmetic amplifies to ~400–500 B/char — the same line that the new guard prices at `1 run + 8×594 ≈ 7 KB`. A length-based formula cannot bound this; either the per-char constant covers the worst (and re-rejects long sparse lines), or the parser needs to fail gracefully on OOM instead of aborting (owl's generated `construct_node_alloc`/`realloc` paths call `abort()`).

</details>

<details>
<summary>Finding 2: BT controller ISR crash on transient exhaustion</summary>

With BT active, `core.print(1+1+…)` with only 51 terms (103 chars) crashes deterministically — not in owl, but in the Bluetooth controller's high-level interrupt:

```
Guru Meditation Error: Core 0 panic'ed (LoadProhibited)
PC 0x4008ee44: r_rwbtdm_isr_wrapper (intc.c)
0x40083204: hli_c_handler          (bt/controller/esp32/hli_api.c:78)
EXCVADDR: 0x00000b6a   (null + struct offset)
```

The parse's transient spike leaves the BT controller unable to allocate its buffers; the controller ISR then dereferences the failed allocation. The reboot loop looks exactly like the F27 boot crash from 2026-08-17. This is why I'd add an absolute reserve on top of the parse estimate whenever more than owl competes for the heap.

</details>

evnchn:
**TL;DR for the above:** bench-verified on a classic ESP32: `main` rejects the 3.2 KB F27 startup outright (demands 209 KB), this PR boots it with BT. Non-blocking, ship it. 

Two constant gaps: 

- token-dense lines still `abort()` in owl instead of dropping, 
- guard needs a 16–24 KB absolute reserve for the BT controller.

falkoschindler:
Thanks @ftilde — the measurements here are what settled the question, just not in the direction of tuning the constants.

The review of this PR showed that the guard from #241 cannot be made correct with any length-based estimate: `PARSE_HEAP_PER_CHAR = 8` and `PARSE_TREE_PER_CHAR = 4` cover config-shaped lines but understate expression-dense ones (`x=1+1+…+1` at 999 chars peaks at ~120 KB vs. 18 KB estimated; `core.print(1,1,…)` needs a 12 KB tree block vs. 4 KB estimated), and raising the constants enough to cover those would reject your startup script again. There is no single curve that fits both.

So instead of tuning, we are reverting #241 on main (#278), which restores the behaviour you need without any guard, and reopening #240 for the real fix: getting owl's construction phase to return `ERROR_ALLOCATION_FAILURE` instead of aborting, preferably upstream (ianh has been responsive to our previous requests), otherwise via a fork or a patch step in `gen_parser.sh`. This PR then becomes unnecessary; Closing it in favour of #278. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Substance%20from%20the%20conversation%20with%20the%20maintainers%2C%20wording%20from%20the%20model.&a=claude-code "claude-fable-5: Substance from the conversation with the maintainers, wording from the model.")

PR #273, opened 2026-08-19
Report serial bus inbound drops and guard the echo relay

JensOgorek:
### Motivation

Three robustness issues in the serial bus, found while investigating the F27 heap exhaustion (fixed by #274):

- Inbound overflow was invisible: the warning about a full inbound queue was printed into that same full queue and silently discarded.
- When relaying echo output to the coordinator failed (`enqueue_outgoing_message` throws after blocking 50 ms on a full outbound queue), `handle_echo` echoed a warning — which re-entered `handle_echo` through the echo callback, tried to relay the warning too, failed again, and recursed on the main task for as long as the outbound queue stayed full.
- `__DONE__` handling indexed `peer_ids[poll_index]` without checking `is_coordinator()`, i.e. `operator[]` on an empty vector if a stray `__DONE__` reaches a peer.

### Implementation

Inbound drops are counted in the communication task (which must not `echo()` itself) and reported by `step()` on the main task, at most once per second ("serial bus ... dropped N inbound messages"), covering both real messages and the communication task's own warnings. `handle_echo` stops relaying before echoing its own failure, and `__DONE__` is only matched against `peer_ids` on a coordinator.

The queue lengths are unchanged at 32. An earlier revision of this PR trimmed them to save heap; that turned out to be unnecessary once #274 landed, and the unreachable-peer scenario that sizes the outbound queue has no upper bound anyway (each poll to a dead peer blocks the drain for the 250 ms poll timeout, and the queue is shared by all receivers) — that is a polling problem, tracked separately. What stays from that detour is `BUS_OTB_WINDOW` in `otb.h`, mirrored by `otb_update.py` like `CHUNK_SIZE`, with `static_assert`s that both queues exceed the OTB window: the peer's inbound queue legitimately holds all in-flight chunks while `step()` is stalled by a flash write, and the coordinator's outbound queue holds them across a stalled poll (spotted by @evnchn). The next heap squeeze will hit the assert instead of aborting uploads in the field.

Bench-verified on a classic ESP32 running the full F27 startup: polling, `bus.send` and the overflow error path keep working.

### Progress

- [x] Tested on hardware.
- [x] Documentation is not necessary.

### DIFF
--- a/main/modules/serial_bus.cpp
+++ b/main/modules/serial_bus.cpp
@@ -17,10 +17,16 @@ extern void process_line(const char *line, const int len);
 
 static constexpr size_t FRAME_BUFFER_SIZE = 512;
 static constexpr unsigned long POLL_TIMEOUT_MS = 250;
-// ~264 bytes per slot; outbound must stay above WINDOW = 8 in otb_update.py, or a stalled
-// poll (outbound drains on the comm task, gated by !is_polling) aborts an OTB update.
-static constexpr size_t OUTGOING_QUEUE_LENGTH = 16;
-static constexpr size_t INCOMING_QUEUE_LENGTH = 8;
+// NOTE: each slot costs sizeof(OutgoingMessage) / sizeof(IncomingMessage) bytes of heap.
+// Both queues must stay above otb::BUS_OTB_WINDOW: the peer's inbound queue legitimately holds the whole
+// window while step() is stalled by a flash write, and the coordinator's outbound queue only drains
+// between polls, so a single stalled poll (POLL_TIMEOUT_MS) would otherwise abort an OTB update.
+// The outbound queue is shared by all receivers and stops draining for POLL_TIMEOUT_MS whenever a peer
+// is unreachable, so it stays at 32 to keep serving the healthy peers at full rate in that case.
+static constexpr size_t OUTGOING_QUEUE_LENGTH = 32;
+static constexpr size_t INCOMING_QUEUE_LENGTH = 16;
+static_assert(OUTGOING_QUEUE_LENGTH > otb::BUS_OTB_WINDOW);
+static_assert(INCOMING_QUEUE_LENGTH > otb::BUS_OTB_WINDOW);
 static constexpr const char ECHO_CMD[] = "__ECHO__";
 static constexpr const char POLL_CMD[] = "__POLL__";
 static constexpr const char DONE_CMD[] = "__DONE__";
@@ -74,10 +80,12 @@ void SerialBus::step() {
         this->handle_incoming_message(message);
     }
 
-    // the communication task cannot echo() itself, and a full inbound queue would swallow its warnings
-    const uint32_t dropped = this->dropped_inbound.exchange(0);
-    if (dropped > 0) {
-        echo("warning: serial bus %s dropped %u inbound messages (queue full)", this->name.c_str(), (unsigned)dropped);
+    // the communication task must not echo() itself, and a full inbound queue would swallow its warnings,
+    // so drops are counted there and reported here, at most once per second
+    if (this->dropped_inbound > 0 && millis_since(this->last_drop_report_millis) > 1000) {
+        const unsigned dropped = this->dropped_inbound.exchange(0);
+        this->last_drop_report_millis = millis();
+        echo("warning: serial bus %s dropped %u inbound messages (queue full)", this->name.c_str(), dropped);
     }
 
     if (this->otb_session.handle != 0) {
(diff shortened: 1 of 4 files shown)


JensOgorek:
### Verified on F27 — the full Bluetooth + serial-bus + arm config boots with 47.5 KB free where it used to crash out-of-memory

Flashed on F27's core (rb66, classic ESP32, lizard-zz build on the robot's current v0.12 base) together with #274: the full target config — Bluetooth, serial bus coordinator, plexus-mirrored RMD arm state, IMU, expander — boots and streams with 47.5 KB free heap; the arm state arrives over the bus (live peer heartbeat, no poll timeouts) and the rosys checksum matches. The same firmware running the robot's current Bluetooth-only driving config idles at 51.9 KB, up from 12.9 KB before. Bench numbers with the same script: 13.7 KB → 52.7 KB.

Not covered: BLE pairing with an active PIN (F27 runs `deactivate_pin()`; mbedtls runs on the NimBLE host task — worth one pairing test on a robot with PIN before a release).

evnchn:
`OUTGOING_QUEUE_LENGTH = 8` now exactly matches `WINDOW = 8` in `otb_update.py`, which leaves the OTB uploader zero headroom.

The uploader keeps 8 unacked chunk commands in flight, all entering via `enqueue_outgoing_message`. Outbound is not drained on the main-loop step the way inbound is — `send_outgoing_queue()` runs on the communication task, gated by `!is_polling`. So while a poll to an unresponsive peer is outstanding (up to `POLL_TIMEOUT_MS`), nothing drains, the 8 slots fill, the next `bus.send` throws, the peer never ACKs that sequence, and `otb.cpp:98` rejects the following chunk as an unexpected sequence number. The update aborts on a remote board.

In steady state this is unlikely — wire time binds long before queue depth does — but the stalled-poll window makes it reachable, and OTB is not among the worst-case scenarios in the PR description.

Suggest keeping outbound strictly above `WINDOW` (12 or 16) and cross-referencing the two constants, the way `otb_update.py:10` already flags `CHUNK_SIZE` against `BUS_OTB_CHUNK_SIZE`. Inbound at 8 looks fine — it sits behind the 2048-byte RX ring and the 100-deep pattern queue.

One small thing on the new comment: it says the queues are drained every main-loop step, which holds for `incoming_queue` but not for the outgoing one.

main/modules/serial_bus.cpp
```diff
@@ -17,8 +17,10 @@ extern void process_line(const char *line, const int len);
 
 static constexpr size_t FRAME_BUFFER_SIZE = 512;
 static constexpr unsigned long POLL_TIMEOUT_MS = 250;
-static constexpr size_t OUTGOING_QUEUE_LENGTH = 32;
-static constexpr size_t INCOMING_QUEUE_LENGTH = 32;
+// 8 slots x ~264 bytes per direction; the queues are drained every main-loop step and
+// senders pace themselves (enqueue blocks up to 50 ms), so deeper queues only cost heap.
+static constexpr size_t OUTGOING_QUEUE_LENGTH = 8;
+static constexpr size_t INCOMING_QUEUE_LENGTH = 8;
```
evnchn: Suggestion for the OTB headroom point, keeping inbound at 8 and correcting the drain claim. Same line count as now:

```suggestion
// ~264 bytes per slot; outbound must stay above WINDOW = 8 in otb_update.py, or a stalled
// poll (outbound drains on the comm task, gated by !is_polling) aborts an OTB update.
static constexpr size_t OUTGOING_QUEUE_LENGTH = 16;
static constexpr size_t INCOMING_QUEUE_LENGTH = 8;
```

JensOgorek:
Adopted in 33b1a6e and a13c065 — outbound is 16 now, inbound stays at 8, and the constants comment takes your suggestion verbatim (OTB coupling, comm-task drain). Good catch: the original comment's "drained every main-loop step" claim was simply wrong for the outbound queue, and the WINDOW = 8 collision would have been nasty to debug on a remote board. The PR description now covers the OTB stall scenario and the updated numbers (~10.7 KB saved instead of ~12.8 KB).

JensOgorek:
### Queue sizing verified under real OTB load on F27 — the trimmed queues never overflowed

All eight OTB transfer runs for #276 ran against this PR's firmware on F27's core (outbound 16 / inbound 8): the ack stream stayed gap-free to the end of every transfer with the full window of 8 chunks in flight plus the arm's bus telemetry, so neither queue was ever the bottleneck. The frame losses that #276 now recovers from originate in the coordinator's console ingest, not in these queues — measurement details and the run table are in #276.

Tested with Claude (Fable 5).

falkoschindler:
Final state after the review round:

- The queue trimming is gone (32/32 as on main): #274 alone lifts F27 well clear of the abort, and the unreachable-peer case that sizes the outbound queue has no upper bound — a queue length only moves the threshold. That is now tracked as a polling problem in #279.
- What stays are the robustness fixes found along the way: inbound drops are counted in the communication task and reported from `step()` instead of being printed into the full queue; `handle_echo` no longer recurses through its own failure warning; `__DONE__` no longer indexes `peer_ids` on a non-coordinator; and `BUS_OTB_WINDOW` (mirrored by `otb_update.py`) is guarded by `static_assert`s so the next heap squeeze cannot break OTB updates (thanks @evnchn for the hint that started this).

Bench verification was done on the earlier 8/8 revision; the remaining changes build cleanly but the two guards have not been exercised on hardware — they are one-liners on paths that were previously UB or unbounded recursion, so I would not block on that.

falkoschindler (approved):

PR #274, opened 2026-08-19
Parse BLE lines on the main task, not the NimBLE host task

JensOgorek:
### Motivation

The Bluetooth module runs its message handler — and with it the whole owl parser — directly on the NimBLE host task. That is why the NimBLE task carries a 32 KB stack (440aae8): it must be big enough to parse arbitrary Lizard lines. Those 32 KB are the larger half of Bluetooth's ~58 KB heap cost, which on F27 (together with the ~30 KB serial bus footprint) exhausts the core's heap: the core aborts out-of-memory while parsing its own startup script (measured 2026-08-17 on the robot, reproduced on a bench ESP32).

### Implementation

The BLE receive callback now only hands the received line to a 32-slot queue without blocking; `Bluetooth::step()` drains the queue and runs the message handler on the main task, like lines from every other transport. If the main task is blocked long enough for the queue to overflow, further lines are dropped and one warning with the drop count is printed when the main loop continues. The NimBLE host task stack shrinks to 6144 bytes (`CONFIG_BT_NIMBLE_HOST_TASK_STACK_SIZE`), freeing 26 KB of heap. Latency changes by at most one main-loop step.

Bench-verified on a classic ESP32 running the full F27 startup (Bluetooth + serial bus coordinator + expander): together with #273, free heap after boot goes from 13.7 KB to 52.7 KB. BLE round-trip tested with a laptop BLE client: command written to the GATT characteristic is parsed on the main task and `bluetooth.send` notifies the response back. Bench-measured: the NimBLE host task uses ~3.2 KB of stack up to the passkey stage of LE Secure Connections pairing (2976 bytes free of 6144; identical usage with a 32 KB stack).

### Progress


JensOgorek:
### Verified on F27 — the full Bluetooth + serial-bus + arm config boots with 47.5 KB free where it used to crash out-of-memory

Flashed on F27's core (rb66, classic ESP32, lizard-zz build on the robot's current v0.12 base) together with #273: the full target config — Bluetooth, serial bus coordinator, plexus-mirrored RMD arm state, IMU, expander — boots and streams with 47.5 KB free heap; the arm state arrives over the bus (live peer heartbeat, no poll timeouts) and the rosys checksum matches. The same firmware running the robot's current Bluetooth-only driving config idles at 51.9 KB, up from 12.9 KB before. Bench numbers with the same script: 13.7 KB → 52.7 KB.

Not covered: BLE pairing with an active PIN (F27 runs `deactivate_pin()`; mbedtls runs on the NimBLE host task — worth one pairing test on a robot with PIN before a release).

evnchn:
The lambda now captures `this` where it previously captured `message_handler` by value, which adds a lifetime dependency the old code did not have.

`BleCommand::init()` runs in the constructor, so the callback goes live before `Global::add_module()` has a chance to reject the module. `main.cpp:209` pre-checks `has_module` but not `variables`, and `global.cpp:35` throws on a collision with a variable name:

```
int bluetooth = 0
bluetooth = Bluetooth("robot")
```

The constructor registers the callback and brings BLE up, `add_module` throws, the `shared_ptr` is destroyed — but BLE keeps advertising with a dangling `this`, and `ble_command.cpp:367` (`if (running) return;`) means the callback can never be replaced. The next GATT write from any paired client runs `xQueueSend` on freed memory.

Capturing the queue handle by value avoids this; the lambda needs nothing else from the object:

```cpp
QueueHandle_t queue = this->line_queue;
ZZ::BleCommand::init(device_name, [queue](const std::string_view &message) { ... });
```

Separately, worth confirming the drop behaviour is intended: on queue overflow `on_chr_access` still returns 0, so a BLE-only client sees success for a line that was discarded. With `WRITE_NO_RSP` there is no throttle, so a bulk config upload can lose statements mid-script.

main/modules/bluetooth.cpp
```diff
@@ -14,17 +17,41 @@ const std::map<std::string, Variable_ptr> Bluetooth::get_defaults() {
...
+    // Only copy the line here: this callback runs on the NimBLE host task, whose stack is far
+    // too small for the parser. The line is parsed in step() on the main task instead.
+    ZZ::BleCommand::init(device_name, [this](const std::string_view &message) {
+        char *line = static_cast<char *>(malloc(message.length() + 1));
+        if (!line) {
+            return;
+        }
+        memcpy(line, message.data(), message.length());
+        line[message.length()] = '\0';
+        if (xQueueSend(this->line_queue, &line, 0) != pdTRUE) {
```
evnchn: Suggestion for the `this`-capture lifetime issue — an init-capture keeps the queue handle without the lambda holding the object, so nothing has to outlive the constructor. Two lines change, none are added:

```suggestion
    ZZ::BleCommand::init(device_name, [queue = this->line_queue](const std::string_view &message) {
        char *line = static_cast<char *>(malloc(message.length() + 1));
        if (!line) {
            return;
        }
        memcpy(line, message.data(), message.length());
        line[message.length()] = '\0';
        if (xQueueSend(queue, &line, 0) != pdTRUE) {
```

JensOgorek:
Fixed in de070a6 and 4a44fd7 — the callback now uses your init-capture and holds only the queue handle, so a module rejected by `add_module` can no longer leave BLE advertising with a dangling `this`. On the drop behaviour: silent-but-successful wasn't intended. The send now blocks up to 100 ms before dropping, which gives bulk uploads backpressure via the main-loop drain — still strictly less blocking for the NimBLE host task than before this PR, where the entire parse ran in the callback. If you'd rather keep the callback fully non-blocking, happy to go back to 0 ticks.

falkoschindler:
I reviewed this and pushed two follow-up commits on top of the branch instead of round-tripping comments, since @JensOgorek is about to be away and we agreed to get this merged.

- The queue-full path called `echo()` on the NimBLE host task. `echo()` writes an unlocked static buffer and fans out to `SerialBus::handle_echo`, which can throw when the outbound queue is full — an exception unwinding through NimBLE's C frames would abort. The callback now only counts drops; `step()` prints one warning with the count on the main task.
- `xQueueSend` no longer blocks the host task for 100 ms: the characteristic has `WRITE_NO_RSP`, so the wait gave no backpressure for write-without-response but still stalled HCI handling and `bluetooth.send` notifications. The queue is 32 slots with a 0-tick send instead, like SerialBus.
- `CommandCallback` now hands over ownership of the flat buffer that `on_chr_access` already allocates (`std::unique_ptr<char[]>`), so a line is allocated and copied once instead of twice; `step()` owns it via `unique_ptr`, so the free no longer depends on the catch clause.
- `CONFIG_NIMBLE_TASK_STACK_SIZE` is only a legacy alias in IDF 5.3.1 (`components/bt/sdkconfig.rename`); renamed to `CONFIG_BT_NIMBLE_HOST_TASK_STACK_SIZE` in both defaults files. Same effect, no dependency on the rename table.
- Documented in the module reference that BLE lines are queued, parsed one main-loop step later at most, and dropped with a warning if the queue overflows.

Bench-tested on an ESP32: BLE round-trip through the new queue path works, and the `nimble_host` stack high-water mark with 6144 bytes is 2976 bytes free after the passkey stage of LE Secure Connections pairing (identical ~3.2 KB usage with a 32 KB stack), so the smaller stack has roughly 2x headroom for the measured path. Pairing itself currently fails before PIN entry on my bench setup with both stack sizes — that is independent of this PR and tracked in #282.

A second commit fixes a pre-existing gap this PR made visible: main.cpp ran a module constructor before checking whether a variable of that name exists, so a constructor with side effects (Bluetooth starting NimBLE) could complete and then be discarded, leaving the BLE callback bound to a queue nobody drains.

Not changed, for the record: lines can now be dropped under a long main-task stall, which the old design never did (the host task blocked on the interpreter lock). With 32 slots and no blocking that only happens during things like `expander.flash()` or `reset_bonds()`; if we ever want hard flow control, the clean way is dropping `WRITE_NO_RSP` from the characteristic so the ATT response throttles the client — but that changes the protocol for the apps, so I left it.

falkoschindler (approved):

PR #275, opened 2026-08-19
Queue every BNO08x sensor report instead of keeping the last one per packet

falkoschindler:
### Motivation

With `ImuBno085` in its default `ndof` mode the quaternion (`quat_w/x/y/z`) reaches the core message at only ~1 Hz although every report is enabled with a 100 ms interval. On the AIR² test platform that meant a heading that lagged every rotation burst by up to a second (the 2026-08-06 rosys-side fusion fix worked around exactly that), and it looked like a slow sensor until the serial logs were counted: ~100 Hz core messages, quaternion changes at 1.1–1.3 Hz even while turning, gyroscope changes at ~10 Hz.

The cause is in the driver, not the sensor: `sh2.c` calls the sensor callback once per report of an SHTP packet, but `Bno08x::getSensorEvent` (derived from the Adafruit driver) decodes every callback into the same single `pending_value`, so only the last report of a packet survives. The hub bundles the reports due at the same tick into one packet with a fixed order — the gyroscope sits late and survives, the game rotation vector rarely does. Enabling fewer reports (`imu.set_mode("imu")`: accelerometer, gyroscope, rotation vector) lifts the quaternion to ~9 Hz with the unpatched firmware, which is how the cause was narrowed down.

### Implementation

A fixed ring buffer of decoded `sh2_SensorValue_t` in `Bno08x` (32 entries): `sensor_handler` decodes each report and pushes it (oldest dropped when full), `getSensorEvent` pops one event per call and only calls `sh2_service()` when the queue is empty. `pending_value` goes away; the `SH2_GYRO_INTEGRATED_RV` special case is no longer needed because every decoded event is queued regardless of its timestamp. `ImuBno085::step()` already drains up to 16 events per step at ~100 Hz, which covers 7 × 10 Hz with room to spare. Method names follow the file's existing Adafruit-style camelCase.

Docs: one line in the BNO085 section — the gyroscope is reported in rad/s (SH2 calibrated gyroscope report), not degrees/s as on the BNO055; verified against a commanded 0.2 rad/s turn (gyro integral 92.1° vs. quaternion yaw change 92.7°).

Measured on a Robot Brain (ESP32, Jetson Orin, BNO085, 0.12.0 + this patch, spot turns at 0.2 rad/s, `ndof`): quaternion changes at 10.0 Hz while rotating (1.0 Hz before), gyroscope 10.0 Hz (unchanged). Compiles for esp32 and esp32s3 on main and on v0.12.0 (IDF 5.3.1 Docker image). Only the esp32 build was flashed.

### Progress

- [x] Tested on hardware.

🤖 Generated with [Claude Code](https://claude.com/claude-code) [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Diagnosis%20from%20the%20serial%20logs%20and%20the%20driver%20code%2C%20patch%20and%20text%20by%20the%20model%3B%20the%20author%20directed%2C%20flashed%20and%20measured%20the%20result%20on%20the%20robot.&d=2026-08-19&a=claude-code "claude-fable-5: Diagnosis from the serial logs and the driver code, patch and text by the model; the author directed, flashed and measured the result on the robot.")

### DIFF
--- a/components/bno08x_driver/bno08x.cpp
+++ b/components/bno08x_driver/bno08x.cpp
@@ -69,7 +69,7 @@ void sensor_handler(void *cookie, sh2_SensorEvent_t *event);
 
 Bno08x::Bno08x(gpio_num_t reset_pin)
     : port(I2C_NUM_0), address(I2C_ADDR_DEFAULT), int_pin(GPIO_NUM_NC),
-      reset_pin(reset_pin), pending_value(nullptr), reset_occurred(false) {
+      reset_pin(reset_pin), reset_occurred(false) {
     std::memset(&prodIds, 0, sizeof(prodIds));
     std::memset(&hal, 0, sizeof(hal));
 }
@@ -172,14 +172,29 @@ bool Bno08x::enableReport(sh2_SensorId_t sensorId, uint32_t interval_us) {
 }
 
 bool Bno08x::getSensorEvent(sh2_SensorValue_t *value) {
-    pending_value = value;
-    value->timestamp = 0;
+    if (queue_count == 0) {
+        sh2_service(); // decodes every report of the next packet into the queue via sensor_handler
+    }
+    return popSensorEvent(value);
+}
 
-    sh2_service();
+void Bno08x::pushSensorEvent(const sh2_SensorValue_t &value) {
+    if (queue_count == EVENT_QUEUE_SIZE) {
+        // consumer too slow: drop the oldest event, the sensor keeps reporting anyway
+        queue_head = (queue_head + 1) % EVENT_QUEUE_SIZE;
+        --queue_count;
+    }
+    event_queue[(queue_head + queue_count) % EVENT_QUEUE_SIZE] = value;
+    ++queue_count;
+}
 
-    if (value->timestamp == 0 && value->sensorId != SH2_GYRO_INTEGRATED_RV) {
+bool Bno08x::popSensorEvent(sh2_SensorValue_t *value) {
+    if (queue_count == 0) {
         return false;
     }
+    *value = event_queue[queue_head];
+    queue_head = (queue_head + 1) % EVENT_QUEUE_SIZE;
+    --queue_count;
     return true;
 }
 
@@ -314,15 +329,17 @@ void hal_callback(void *cookie, sh2_AsyncEvent_t *event) {
 
 void sensor_handler(void *cookie, sh2_SensorEvent_t *event) {
     auto *imu = static_cast<Bno08x *>(cookie);
-    if (!imu || !imu->pending_value) {
+    if (!imu) {
         return;
     }
 
-    int rc = sh2_decodeSensorEvent(imu->pending_value, event);
+    sh2_SensorValue_t value{};
+    int rc = sh2_decodeSensorEvent(&value, event);
     if (rc != SH2_OK) {
         ESP_LOGE(TAG, "sh2_decodeSensorEvent failed: %d", rc);
-        imu->pending_value->timestamp = 0;
+        return;
     }
+    imu->pushSensorEvent(value);
 }
 
 } // namespace
--- a/components/bno08x_driver/bno08x.cpp
+++ b/components/bno08x_driver/bno08x.cpp
@@ -180,12 +180,22 @@ bool Bno08x::getSensorEvent(sh2_SensorValue_t *value) {
 
 void Bno08x::pushSensorEvent(const sh2_SensorValue_t &value) {
     if (queue_count == EVENT_QUEUE_SIZE) {
-        // consumer too slow: drop the oldest event, the sensor keeps reporting anyway
+        // a single cargo pushed more reports than fit (hub batched several ticks after a host stall):
+        // drop the oldest event, the sensor keeps reporting anyway
+        if (!overflow_logged) {
+            ESP_LOGW(TAG, "sensor event queue overflow, dropping oldest event");
+            overflow_logged = true;
+        }
         queue_head = (queue_head + 1) % EVENT_QUEUE_SIZE;
         --queue_count;
     }
     event_queue[(queue_head + queue_count) % EVENT_QUEUE_SIZE] = value;
     ++queue_count;
+}
+
+void Bno08x::clearSensorEvents() {
+    queue_head = 0;
+    queue_count = 0;
 }
 
 bool Bno08x::popSensorEvent(sh2_SensorValue_t *value) {
--- a/components/bno08x_driver/bno08x.h
+++ b/components/bno08x_driver/bno08x.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <array>
+#include <cstddef>
 #include <cstdint>
 #include <memory>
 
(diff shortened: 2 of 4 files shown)


evnchn:
*Drafted by Claude Code on evnchn's behalf.*

**The queue is the right fix. One follow-up worth doing before merge.**

Dropping the `!pending_value` guard also makes `sensor_handler` queue reports that arrive outside a `getSensorEvent` window. `sh2_setSensorConfig` spins `shtp_service()` internally (`sh2.c`, `opProcess`), so reports landing during `apply_mode` are now decoded and queued instead of discarded. After `imu.set_mode("acconly")` the next one or two `step()` calls still write `gyr_*` / `quat_*` for the sensors just disabled; same after a hub reset, with pre-reset samples. Bounded at 2 steps (32 queued, 16 drained per step), so about 20 ms at 100 Hz. Cosmetic, not a correctness break.

Worth a `clearSensorEvents()` next to the existing push/pop, called from `hardwareReset()` and after the reset-triggered `apply_mode`?

<details><summary><b>Two nits</b></summary>

The overflow comment names a case that is essentially unreachable. "Consumer too slow" is not the path that gets there: one `sh2_service()` pushes every report of a cargo before `getSensorEvent` returns even once, so an overrun happens inside a single push burst. With `SH2_HAL_MAX_TRANSFER_IN` at 384 B and an ndof tick around 75 B, a host stall lets the hub batch several ticks into one cargo and overrun 32 entries in that burst. Harmless under drop-oldest, but the invariant would invert if the policy changed. (The batching estimate is inference from the transfer size, not something I measured.)

Depth 32 costs 1536 B, at `sizeof(sh2_SensorValue_t)` of 48 B, for a queue that ends virtually every step empty: 70 events/s in against up to 1600/s drained. Either 8–16 covers the bound the PR body reasons about, or 32 stays and the comment names the burst bound instead of consumer speed.

</details>

<details><summary><b>Checked at source, no hardware here</b></summary>

Against the Adafruit submodule fetched from upstream, since it is not checked out in a fresh clone:

- The diagnosis holds. `sensorhubInputHdlr` in `sh2.c` calls `sensorCallback` once per report while walking the cargo, so a single `pending_value` did keep only the last one.
- The non-atomic head/count is safe: `sh2_service()` is reachable only from `getSensorEvent` and `opProcess`, both on the Lizard main task, and `int_pin` is bound to no ISR with no extra task or `esp_timer` callback in play.
- The docs line is right. `sh2_SensorValue.c` decodes `SH2_GYROSCOPE_CALIBRATED` with `SCALE_Q(9)` and documents `[rad/s]`, and `imu_bno085.cpp` copies `un.gyroscope` unscaled.

</details>

JensOgorek (commented):
### The queue is the right fix and the diagnosis is convincing — one correctness gap remains (no flush on mode change/reset), plus two cheap wins

The ring buffer solves exactly the last-report-wins bug your serial-log counting pinned down, the removal of the `timestamp == 0` special case checks out (the old check even rejected valid reports), and nothing in the hot path got more expensive. What I would still fix before merge: events decoded *during* reconfiguration survive it — details below. Queue size and the silent drop are cheap follow-ons; the rest is optional.

<details><summary><b>Stale-event replay: the queue is never flushed on a mode change or reset</b></summary>

`sh2_setSensorConfig` runs through `opProcess`, which busy-loops `shtp_service()` until the response arrives (sh2.c:493-501) — and that dispatches the sensor callback for every in-flight report (sh2.c:625). So each of `apply_mode()`'s up to 7 blocking config calls pushes reports of the *old* configuration into the queue, and no flush exists anywhere in the driver. The next `step()`s then publish gyro/quat values the user just disabled — a Lizard rule on `imu.gyr_z` can fire after `set_mode("acconly")` — and after a sensor brown-out, pre-reset samples are published as current. The old code discarded these via the `!pending_value` guard. A small `clearEvents()` called from `apply_mode()` (and on `wasReset()`) closes it.

</details>

<details><summary><b>Queue size: 32 slots are ~1.5 KB of DRAM for a physical maximum of 7</b></summary>

`sizeof(sh2_SensorValue_t)` is 48 bytes (verified with a `_Static_assert`), so `event_queue` costs 1536 bytes per instance, value-initialized on construction. One SHTP packet can carry at most one report per enabled sensor and only 7 types exist, so 8 slots cover the worst case and save 1.15 KB — welcome in a project that has just been hunting heap. While there: the `{}` on `event_queue` buys nothing, the indices are the invariant.

</details>

<details><summary><b>Silent drop-oldest, recovery latency, and an optional cleaner cut</b></summary>

Overflow drops the oldest event with no log — the only lossy path without one (the decode error next to it logs). Reachable via the config window above or a long main-loop stall; a rate-limited `ESP_LOGW` on the first drop makes it diagnosable. Related but latency-only: `sh2_service()` now runs only on an empty queue, so a post-stall backlog drains at ~2-3 packets per step instead of 16, and `wasReset()` is seen only once the queue empties. If you ever want the cleaner shape: split `service()` (pump the transport once) from `popSensorEvent()` (pure queue) — one bus interaction per step, full drain, and both magic numbers (32 here, 16 in the module) disappear.

</details>

<details><summary><b>Question on the gyro unit, one style nit, one pre-existing issue</b></summary>

The rad/s difference to the BNO055 is documented, but the module otherwise promises drop-in compatibility with the same property names — was normalizing to degrees/s in the module (3 lines, next to the existing `accuracy_from_status()` normalization) considered? If AIR² already consumes rad/s, keeping and documenting is the right call, then this is settled. Nit: the new private block declares the variables before `popSensorEvent`, CONTRIBUTING wants methods first (the camelCase is consistent with the file's Adafruit heritage, fine). And noticed while tracing, pre-existing and not for this PR: `~Bno08x()` never calls `sh2_close()` or deregisters the callback, so a failed `ImuBno085` construction leaves a dangling cookie and the shtp instance occupied — re-creating the IMU fails until reboot. Worth its own issue.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Review%20by%20the%20model%20%288%20finder%20passes%2C%20key%20findings%20re-verified%20against%20sh2.c%29%2C%20directed%20and%20curated%20by%20the%20author.&d=2026-08-19&a=claude-code "claude-fable-5: Review by the model (8 finder passes, key findings re-verified against sh2.c), directed and curated by the author.")

falkoschindler:
Thanks @evnchn and @JensOgorek, addressed in the latest commit:

- **Stale events after a mode change / reset:** new `Bno08x::clearSensorEvents()`, called at the end of `ImuBno085::apply_mode()`. That covers both `set_mode(...)` and the `wasReset()` path (which re-runs `apply_mode`), so reports decoded while `sh2_setSensorConfig` waited for its response are no longer published.
- **Queue size:** 32 → 16 (768 B instead of 1.5 KB). I did not go down to 8 because Evan's point stands: one cargo carries one report per sensor only while the host keeps up; after a stall the hub can batch a few ticks into one cargo, and that burst is pushed before a single pop happens. 16 keeps one batched tick of headroom; drop-oldest stays harmless either way. The comment now names that burst bound instead of "consumer too slow". The `{}` on `event_queue` is gone.
- **Silent drop:** `ESP_LOGW` on the first overflow (once per instance, so a stalled loop cannot spam the log).
- **Style:** methods now precede the data members in the private block.

Left as is: the `service()` / `popSensorEvent()` split — valid cleaner shape, but it changes the module's drain loop and I would rather keep this PR at the bug fix. Gyro unit stays rad/s and documented: rosys consumes the raw SH2 value and converting in the module would change what the running AIR² setup sees. The missing `sh2_close()` in `~Bno08x()` is pre-existing and gets its own issue.

PR #276, opened 2026-08-20
Recover from lost frames during OTB transfers instead of aborting

JensOgorek:
### Motivation

OTB (update over the serial bus, #182) aborts the whole transfer on the first lost or corrupted frame: the target answers `__OTB_ERROR__:unexpected sequence number` and the session dies. On a real robot bus that makes large transfers practically impossible — on F27's core↔plexus link (RS485 @115200), pushing a 1.36 MB image failed reproducibly after 54–423 of 7842 chunks, in three runs with window sizes 8, 4 and 1. The link itself is not the problem (200/200 idle ping roundtrips without a single loss); the losses only appear under sustained transfer load, and the dominant source is the coordinator's own console ingest: at saturation, incoming chunk lines get torn apart (fragments of base64 show up as `error processing uart0: unknown module "AAs"`), and roughly one in 256 corrupted frames slips through the 8-bit XOR frame checksum and gets flashed silently — one run delivered all 7842 chunks and then failed the image SHA check at `esp_ota_end`.

### Implementation

Coordinator firmware is untouched: acks are reused as the recovery signal and relayed as before, so existing coordinators forward everything needed.

- **Go-back-N retransmission.** On a sequence gap, duplicate, corrupted header or corrupted payload the target no longer kills the session; it re-acknowledges the last chunk it has written (or `__OTB_ACK_BEGIN__` when nothing arrived yet). The host treats a duplicate acknowledgement — or 2 s of ack silence — as the signal to rewind and resend from there. After a rewind the host ignores the stale re-acks of the resent range, so a spurious rewind (late acks, a second begin ack) costs one window and cannot turn into a permanent double-send. Chunks are only ever written in order, so duplicates are discarded safely and `esp_ota_write` stays strictly sequential. A begin retry is idempotent while nothing has been written, and the host retries `__OTB_BEGIN__` against a still-booting target.
- **Per-chunk CRC32, negotiated.** Because the frame checksum is only 8 bits, each chunk carries a zlib CRC32 over its decimal sequence number and its decoded bytes (`__OTB_CHUNK_<seq>__:<8-hex-crc32>:<base64>`), so neither a corrupted payload nor a chunk landing at the wrong offset passes unnoticed. The target advertises support with `__OTB_ACK_BEGIN__:crc32`; the host only sends the CRC field when advertised, so the new script still updates targets on released firmware, and a new target still accepts CRC-less chunks from an older script. Once a session has seen a CRC, a chunk without one is treated as corrupt (the field cannot silently drop out). The firmware uses the ROM's `esp_rom_crc32_le`, which is bit-identical to `zlib.crc32`.
- **Bounded by progress, not traffic.** The target's 10 s session timeout only moves when a chunk is written (or on begin), so a sender that never delivers a valid chunk cannot keep the OTA handle claimed forever; the host gives up after 15 s without progress. `BUS_OTB_CHUNK_SIZE` is lowered to 165 so the chunk line fits the 256-byte bus payload with the CRC field (a `static_assert` in `serial_bus.cpp` derives the bound).

### Verification on hardware (F27, core fix-firmware ↔ plexus ESP32-S3, RS485 @115200)

<details>
<summary>Measurement history — from reproducible failure to a clean flash over the bus</summary>

| Run | Sender | Receiver | Result |
| --- | ------ | -------- | ------ |
| 1 | stock, window 8 | stock | abort at chunk 423/7842 (`unexpected sequence number`) |
| 2 | window 4, paced | stock | abort at chunk 203 (same) |
| 3 | window 1 (ping-pong) | stock | ack lost → sender stalled → target `session timed out` at chunk 54 |
| 4 | go-back-N | retransmission | **all 7842 chunks transferred** (192 resends), but commit failed the image SHA — a corrupted frame had slipped through the 8-bit frame checksum |
| 5 | + CRC32 | + CRC32 | begin frame hit a still-booting target (fixed by begin retry) |
| 6 | + CRC32 | + CRC32 | died at chunk ~2251: a fully lost window was never resent while the console kept talking (review finding 1) |
| 7 | + instrumentation, `--expander p0` paused | + CRC32 | same at ~2951 — p0 broadcasts are not the bottleneck |
| 8 | + 10 ms pacing | + CRC32 | complete: 8270 chunks in 289 s, 504 resends, commit passed |
| 9 | review rework | **stock (released receiver)** | negotiation falls back to CRC-less chunks, 50+ chunks accepted, then the stock receiver aborts on the first lost frame as before |
| 10 | review rework, 10 ms pacing | review rework | **complete: 8271 chunks in 306.6 s, 592 resends absorbed, commit passed, target rebooted with its NVS config intact** |
| 11 | review rework, no pacing | review rework | **complete: 307.4 s, 560 resends** — pacing makes no difference, removed |
| 12 | review rework, window 6, no pacing | review rework | complete: 328.4 s, 516 resends — smaller window means more rewinds and a slower transfer, window stays 8 |

Runs 9–12 ran with the robot's E-stop engaged, i.e. the core was additionally firing its 1 Hz `delta.stop();delta.off()` rule at the target; the target's console showed 14 frame-checksum errors on the wire during run 10 (4 of them chunks) against 74 window rewinds, so the bulk of the losses still originates on the coordinator's console path, not on the RS485 link.

</details>

<details>
<summary>Host state machine simulated against a model of the receiver</summary>

A fake coordinator/target (FIFO relay with injected loss, corruption and latency, and a receiver mirroring `otb.cpp`) drives `send_chunks()` from the script: 5 % uplink loss, 5 % downlink loss, 5 % corruption and a 3 %/3 %/3 % mix all complete with the image intact; two completely lost windows while the console keeps printing cost exactly 16 resends; two >2 s ack stalls cost exactly 16 resends and no double-send afterwards; a late second begin ack costs one window; a CRC-less target completes with 0 resends; a dead target raises after 15 s.

</details>

### Progress


falkoschindler (changes requested):
Nice work on the measurement series — the table makes it very clear where the losses come from, and go-back-N with cumulative acks is the right shape for this link. I traced the two state machines against each other and found a few cases where the recovery itself gets stuck or loops; the first three are the ones I'd fix before merging.

## Correctness

**1. The ACK_TIMEOUT rewind can only fire on a silent console** — otb_update.py (93-98)

The `time.time() - last_ack_at > ACK_TIMEOUT` check lives inside `if not raw:`, i.e. it only runs when `readline()` returned nothing for 0.5 s. On a coordinator that keeps printing anything (`core.output(...)`, `bus[..]` relays, warnings, expander output before the pause takes effect) `raw` is never empty, the deadline is never compared, and a window that is lost entirely (all 8 chunk frames dropped, so the target has nothing to dup-ack) is never resent. The host idles until the target's 10 s session timeout produces `__OTB_ERROR__:session timed out` and the transfer aborts — exactly the failure the PR removes. Fix: evaluate the deadline on every iteration after the `readline()`, regardless of whether a line arrived. (Related: `last_ack_at` is also refreshed by stale dup-acks that are deliberately ignored while `recovering`, which pushes the deadline out by up to a window's worth of acks; it should only move when `acked` advances or a rewind happens.)

**2. A spurious rewind turns into a self-sustaining double-send for the rest of the transfer** — otb_update.py (112-118)

`recovering` is cleared on the first ack that advances `acked`, not when the rewound window has been fully acknowledged. Trace: window k..k+7 was actually delivered but the acks arrive late (or were lost on the way back, or the target stalled >2 s), so the host hits ACK_TIMEOUT and resends k..k+7. The target acks k..k+7 for the originals, then answers each duplicate with a re-ack of k+7 (eight of them). On the host, ack k clears `recovering`, the window refills with k+8..k+15, then the first stale `ACK_CHUNK_{k+7}` arrives with `n == acked` and `sent > acked` → rewind to k+7 and resend k+8..k+15, which are duplicates again → eight dup-acks of k+15 → next rewind. From then on every window is sent twice; throughput halves and `resends` grows by WINDOW per window without any error surfacing. The fix is to remember the `sent` value at the time of the rewind and keep ignoring dup-acks until `acked` has caught up with it (a fresh loss inside that range is then covered by the timeout). The begin retry makes the same thing reachable from the start: if the first `__OTB_ACK_BEGIN__` was merely delayed past the 10 s `wait_ack`, the target answers the second BEGIN as well, the second ack lands in the chunk loop with `acked < 0 and not recovering` and rewinds a window that was delivered.

**3. The new host cannot flash targets that run the released firmware** — otb_update.py (90), main/utils/otb.cpp (136-148)

OTB shipped in v0.11.0, so there are targets in the field without the CRC field. The compatibility is implemented in the direction we need least: a new target tolerates an old sender, but the new `otb_update.py` sends `<crc>:<base64>` and an old target base64-decodes the whole thing, fails, and aborts with `__OTB_ERROR__:base64 decode failed` on chunk 0. The script and firmware ship together, but the *target* of an OTB update is by definition the old version. Two options: negotiate (new targets answer e.g. `__OTB_ACK_BEGIN__:crc32`, the host only sends the CRC field when advertised — then the `b64[8] == ':'` sniffing can go), or at least let the host fall back to CRC-less chunks when chunk 0 is rejected with `base64 decode failed`.

**4. Header corruption still kills the session** — main/utils/otb.cpp (111-119)

A corrupted `__:` separator or a non-digit in the sequence field still goes through `fail()` → `esp_ota_abort` + session reset, while the very same bit flip one byte later (in the CRC or base64 field) is now recovered. With the rewind machinery in place these two branches are the only single-frame events left that abort a transfer; they should answer via `resend_from_last_written()` as well.

**5. Neither side bounds the retry loop any more** — main/utils/otb.cpp (124), otb_update.py (86-99)

`resend_from_last_written` refreshes `last_activity` on every out-of-order, duplicate or corrupt frame, so `bus_tick()`'s timeout now measures traffic, not progress: a chunk that never verifies (systematic corruption, or the size mismatch from point 3 in the other direction) keeps the target dup-acking forever and the OTA handle claimed until reboot; another sender gets `session already active` indefinitely. The host has the mirror problem: the old `wait_ack` raised after 10 s, the new main loop resets `last_ack_at` on every timeout rewind and has no cap on `resends`, so an unreachable target makes the script spin until Ctrl-C. A progress-based timeout on the target (only bump `last_activity` when `next_seq` advances or on BEGIN, or fail after N re-acks of the same position) plus a resend cap or overall deadline on the host would close both.

**6. A corrupted CRC field silently degrades to the unchecked path** — main/utils/otb.cpp (141-148)

If the `:` at `b64[8]` is hit by a flip that passes the 8-bit frame checksum, `has_crc` stays false and the decoder sees `<8hex><x><base64>`. mbedtls does not require the input length to be a multiple of 4, so if the corrupted byte is a valid base64 character the 229 chars decode to 171 bytes (≤ BUS_OTB_CHUNK_SIZE) and are flashed unchecked. Rare, but it's the one residual hole in the end-to-end check. A sticky per-session flag (`session.uses_crc` set by the first chunk that carries one; later chunks without it are treated as corrupt) closes it without affecting old senders. While here: the CRC covers only the payload, not the sequence number, so a seq digit flip that happens to hit `next_seq` while the real chunk was lost would write the wrong bytes at the right offset with a passing CRC — including `seq` in the CRC (`zlib.crc32(chunk, zlib.crc32(str(seq).encode()))`) is a one-liner on both sides.

## Docs and constants

**7. `CHUNK_SIZE` and `BUS_OTB_CHUNK_SIZE` have drifted apart, and the format docs are stale** — otb_update.py (12), main/utils/otb.h (23), main/utils/otb.cpp (102), docs/tools.md (protocol table)

The real constraint is the bus payload: `__OTB_CHUNK_` (12) + seq digits + `__:` (3) + 9 (CRC) + base64(n) < PAYLOAD_CAPACITY (256). 165 bytes → 220 chars → 248 + digits, fine; 174 bytes → 260, rejected by `enqueue_outgoing_message` as `payload is too large` — and that error is a plain console `error:` line, not `__OTB_ERROR__`, so the host would never see it and loop through timeouts. Right now the header still says 174, the Python comment points at it as the cap, and someone will "fix" the mismatch. Suggest lowering `BUS_OTB_CHUNK_SIZE` to 165 with a comment (or `static_assert`) deriving it from the payload budget, and updating the comment at otb.cpp (102) and the protocol table in docs/tools.md, which still show `__OTB_CHUNK_<seq>__:data` without the CRC field.

## Reuse and simplification

- main/utils/otb.cpp (12-21): `esp_rom_crc32_le(0, buf, len)` (`esp_rom_crc.h`, already linked via ROM, zero-init/final-invert inside) is bit-identical to `zlib.crc32` and replaces the hand-rolled bit loop.
- otb_update.py (95-98, 104-107, 116-118): the rewind is written out three times with slightly different arithmetic; a single `rewind()` helper (`resends += sent - acked; sent = acked; recovering = True; last_ack_at = now`) covers all three — the ACK_BEGIN case is just `acked == -1`.
- otb_update.py (69-75): the begin retry distinguishes timeouts by `'Timeout' in str(e)`; a typed `OtbTimeout(OtbError)` raised from `wait_ack` would be robust to rewording the message.
- otb_update.py (92-102): the readline/decode/`__OTB_ERROR__` check duplicates `wait_ack`; a shared `read_line()` would keep the error handling in one place.
- main/utils/otb.cpp (139-148): `has_crc` + `expected_crc` could be a `std::optional<uint32_t>`.
- main/utils/otb.cpp (126): new line is 124 columns (CONTRIBUTING.md says 120; clang-format doesn't enforce it).

## Questions

- PACE: at 115200 baud one chunk line is ~265 bytes ≈ 23 ms on the wire, so a 10 ms sleep after a non-blocking `write()` shouldn't throttle anything at the default baud — yet run 8 clearly improved over runs 6/7. Was the host→coordinator link running at a higher baud in those runs? If so the comment should say the pacing is for high baud rates; if not, something else changed between runs 7 and 8 and it would be good to know what. The underlying budget is the coordinator's 2048-byte UART0 RX ring vs. a window of 8 × ~265 B ≈ 2.1 KB — sizing WINDOW × line length below the ring would make the tearing structurally impossible instead of paced around. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Findings%20from%20a%20multi-agent%20code%20review%2C%20verified%20and%20written%20up%20by%20the%20model%3B%20the%20reviewer%20decided%20the%20ranking%20and%20what%20to%20post.&d=2026-08-21&a=claude-code "claude-fable-5: Findings from a multi-agent code review, verified and written up by the model; the reviewer decided the ranking and what to post.")

JensOgorek:
### All seven findings fixed and re-verified on F27 — three full transfers over the bus passed, including one against the released receiver

Thanks, the state-machine trace was spot on. The rework is pushed as one commit on top; the PR body is updated accordingly. Summary of what changed and what the new measurements say:

<details>
<summary>Correctness findings 1–6</summary>

1. **Timeout only on a silent console** — the deadline is now checked on every loop iteration after `readline()`, and `last_ack_at` only moves on progress or on a rewind. Verified with console noise on every third read in the simulation (two fully lost windows cost exactly 16 resends).
2. **Spurious rewind → permanent double-send** — `rewind()` remembers `sent` as `stale_until`; duplicate acks with `n <= stale_until` are ignored until acked overtakes it, fresh losses inside that range fall back to the 2 s timeout. A late second `__OTB_ACK_BEGIN__` is handled by the same rule (one wasted window, then stable). Simulated with two >2 s ack stalls: exactly 16 resends, frame count stays at N+16.
3. **New host vs released targets** — negotiated: a new target answers `__OTB_ACK_BEGIN__:crc32`, the host sends the CRC field only when advertised. Verified on F27 against the stock zz4 image (run 9): the fallback kicks in, 50+ chunks are accepted, then the stock receiver aborts on the first lost frame as it always did. The `:` sniffing stays on the target so older scripts keep working against new firmware.
4. **Header corruption** — `__:` missing or a non-digit sequence now goes through `resend_from_last_written()` too. While tracing this in the simulation I found one more frame-level abort: a flip inside the `__OTB_CHUNK_` prefix reached the `unknown command` branch and the host treated that as fatal; the target now re-acks unknown `__OTB_*` frames from its active sender instead.
5. **Unbounded retry** — the target's `last_activity` only moves on begin and when `next_seq` advances (no bump in the re-ack path); the host gives up after 15 s without progress (the target's own timeout is 10 s, so its error message normally arrives first).
6. **CRC field degrading silently** — sticky `session.uses_crc`; the CRC now covers the decimal sequence number followed by the payload on both sides (`zlib.crc32(chunk, zlib.crc32(str(seq).encode()))` ↔ `esp_rom_crc32_le` chained), verified end-to-end by the passing `esp_ota_end` image check in runs 10–12.

</details>

<details>
<summary>Constants, docs, simplifications</summary>

- `BUS_OTB_CHUNK_SIZE = 165` with the line budget spelled out and a `static_assert(otb::BUS_OTB_CHUNK_LINE_SIZE < SerialBus::PAYLOAD_CAPACITY)` in `serial_bus.cpp`; the comment in otb.cpp and the protocol table/flow/paragraph in docs/tools.md show the CRC field and the `:crc32` advertisement.
- `esp_rom_crc32_le` replaces the bit loop; `std::optional<uint32_t>` for the CRC; the three rewinds are one `rewind()`; `OtbTimeout(OtbError)` instead of string matching; shared `read_line()`; the 124-column line is gone.

</details>

<details>
<summary>Your pacing question — you were right, and it changed the diagnosis of runs 6/7</summary>

The host↔coordinator link was at 115200 in all runs, so the 10 ms could not have throttled anything. With the reworked host, run 10 (10 ms pacing) and run 11 (no pacing) are indistinguishable — 592 vs. 560 resends, 306.6 vs. 307.4 s — so pacing is removed. In hindsight runs 6/7 did not die of console collapse but of finding 1: a fully lost window was never resent while the console kept printing, and the target's 10 s session timeout ended the transfer; run 8 simply got lucky. I also tried your structural idea of keeping WINDOW × line below the 2048-byte UART0 ring (window 6, run 12): it completes as well but with 86 rewinds instead of 74 and 328 s instead of 307 s, so the tearing is not a pure ring-size effect and the window stays at 8. Runs 9–12 also ran with the robot's E-stop engaged, i.e. with the core's 1 Hz `delta.stop();delta.off()` rule on the same bus; the target saw 14 frame-checksum errors on the wire during run 10 (4 of them chunks) against 74 window rewinds, which keeps pointing at the coordinator's console path as the main loss source.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Rework%20and%20hardware%20verification%20drafted%20by%20the%20model%3B%20the%20author%20reviewed%20and%20decided%20what%20to%20post.&d=2026-08-21&a=claude-code "claude-fable-5: Rework and hardware verification drafted by the model; the author reviewed and decided what to post.")

falkoschindler (approved):
Thanks @JensOgorek, that's a thorough rework. I traced the new host loop and the target handler against the cases from the review again (genuine loss, spurious rewind after late acks, a late second `__OTB_ACK_BEGIN__`): with `stale_until` all three converge, the timeout now fires regardless of console noise, and the retries are bounded by progress on both sides. The chained `esp_rom_crc32_le` over `seq ‖ payload` is bit-identical to `zlib.crc32(chunk, zlib.crc32(str(seq).encode()))`, and the passing `esp_ota_end` check in runs 10–12 confirms that end-to-end. Nice that the pacing question turned into a better diagnosis of runs 6/7, and good to see run 9 against the stock receiver.

I pushed two small docs commits: bd43376 adds the missing `:crc32` suffix to the begin ack in the flow diagram and makes the placeholders consistently `<seq>`, `<crc>`, `<data>` so the literal `:crc32` token is distinguishable from the CRC value field; bc848c8 wraps the new retransmission paragraph with semantic line breaks like the rest of the newer docs.

One optional follow-up, not for this PR: the host-side simulation you used for the timeout and rewind cases would be worth keeping in the repo as a test for `send_chunks`. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Re-verification%20of%20the%20rework%20traced%20by%20the%20model%3B%20the%20reviewer%20decided%20the%20verdict.&d=2026-08-21&a=claude-code "claude-fable-5: Re-verification of the rework traced by the model; the reviewer decided the verdict.")

PR #277, opened 2026-08-20
Add two-way clock offset estimation to SerialBus

falkoschindler:
**Depends on #291** (`nan` and `inf` literals): merge that first, then main into this branch.

### Motivation

#252 adds one-way time sync to the SerialBus: the peer stamps its DONE, the coordinator subtracts the modelled wire airtime and takes a windowed maximum over 128 poll rounds, because the unknown delays (TX queueing on the peer, RX-to-read on the coordinator) are always ≥ 0 and make every sample low. That works and is hardware-validated, but it has no per-sample error bound, locks only after a full window (~1.6 s per peer, scaling with the number of peers), and needs a full window again after every poll timeout.

This is a sibling branch from `main`, not stacked on #252, with the two-way variant: four timestamps around the POLL/DONE round trip that runs anyway, NTP style. With `T1` coordinator sends POLL, `T2` peer receives it, `T3` peer sends DONE, `T4` coordinator receives it, the offset is `((T2 - T1) + (T3 - T4)) / 2`. The peer's processing time between `T2` and `T3` (the `Ready.` line, queued frames) cancels out, so a single sample locks the estimate, and the remaining transport time `(T4 - T1) - (T3 - T2)` bounds the error of each sample. The hardware A/B (see Verification) did not separate the two estimators in accuracy; this PR goes ahead because of what the measurement cannot see: relock in one sample instead of a full window, convergence independent of the peer count, and no locks between the tasks. #252 stays parked and is closed when this merges.

### Implementation

- **Queue ownership instead of locks.** `make_coordinator()` and `enable_time_sync()` send a `Config` through a length-1 overwrite queue; the communication task adopts it at the top of its loop and exclusively owns `peer_ids`, the polling state and all per-peer clock state. Estimates travel back to the main task as full-state `OffsetUpdate` snapshots through a second queue (published on accept and reset only, retried when the queue is full), and only the main task touches the properties. A generation counter in the config lets the main task drop snapshots from a superseded config. No `portMUX`, no rebuild-and-swap.
- **The estimator.** The peer reports `T3`, `T3 - T2` and the sequence number of the POLL it answers in the DONE frame (`__DONE__<t3>,<t3-t2>,<seq>`, 18-digit cap per field, anything else with that prefix stays a user payload); the coordinator stamps `T1` right before the POLL write and `T4` after `read_line`, and takes a sample only for the POLL with the matching sequence number. Published as `offset_<id>` (peer clock minus coordinator clock, ms, NaN while invalid; same name, sign and NaN contract as #252) and `offset_<id>_accuracy`. The known airtime asymmetry (short POLL vs. long DONE) is subtracted, so the bound covers only the unknown delays. Filter: an unlocked estimate takes the first sample, whatever its bound; from then on a sample with a bound ≤ 0.5 ms replaces the estimate immediately, otherwise the best sample of every 2 s does, so the estimate keeps following crystal drift on a busy bus. The estimate snaps to each accepted sample (no blending) so that the bound is testable per sample. A poll timeout or a renewed `make_coordinator()` / `enable_time_sync()` invalidates the estimate.
- **Compatibility.** The coordinator sends `__POLL__<seq>` only to a peer that has answered with a stamped DONE, and falls back to the bare `__POLL__` after a timeout, so peers on older firmware keep working (without an estimate) until they are updated. Without `enable_time_sync()` the wire protocol is unchanged. Update the coordinator first; an old coordinator would treat a stamped DONE as a user message.
- **Side changes.** `Serial::CHECKSUM_TRAILER_LENGTH` names the trailer that the airtime model needs. A NaN offset is broadcast as `nan`, which needs the literal from #291 on the host; the earlier workaround of skipping NaN properties in `Module::step()` (`554f6e7`) was dropped in `941bb34` because it froze proxied values on the last estimate.

### Verification

Jens ran `7e891f6` against #252 on F31 (three runs, 60 400 samples, both consoles against one Jetson clock; his review below has the details). The per-sample bound was never contradicted in 319 two-second windows; it sits at 0.43–0.50 ms when samples are accepted directly and 0.8–0.93 ms from the best-of-window, set by the 1 ms tick of the two communication tasks. The rig cannot resolve absolute error below about 2 ms, so neither estimator is measurably better than the other. The review also found two real pairing bugs, fixed in `554f6e7`:

- a runtime `enable_time_sync()` / `make_coordinator()` cleared `is_polling` while a POLL was outstanding and sent a second POLL into the in-flight DONE (measured: two poll timeouts and a garbled `__POLL__` on the peer console) — a reconfiguration now lets the outstanding POLL finish;
- a late DONE after a poll timeout was paired with the next POLL's `T1` — the sequence number makes such a DONE fall through.

Both tunables (`SYNC_ACCEPT_ACCURACY_US = 500`, `SYNC_WINDOW_MS = 2000`) are unchanged: on Jens's data they decide how soon a sample is taken, not the bound. Builds for esp32 and esp32s3 (IDF 5.3.1 Docker image), pre-commit clean. Still to do on hardware: repeat the run-2 injections (`enable_time_sync()` and `make_coordinator()` at runtime, peer reset) on `941bb34` and confirm that the relock takes one to two poll rounds without a timeout, plus one mixed-version round (new coordinator, old peer) to see the bare-POLL fallback.

Open after the merge: whether `POLL_TIMEOUT_MS` should keep doing double duty (peer lost vs. estimate invalid), and whether the tunables become `enable_time_sync()` parameters. The Bluetooth/RS485 frame-loss finding from the review gets its own issue.

### Progress

- [x] Tested on hardware.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Fixes%2C%20build%20verification%20and%20wording%20by%20the%20model%3B%20the%20decision%20to%20continue%20with%20this%20PR%20by%20the%20author.&d=2026-09-09&a=claude-code "claude-fable-5-1: Fixes, build verification and wording by the model; the decision to continue with this PR by the author.")


JensOgorek (changes requested):
### Hardware A/B on F31 done: the two-way bound is self-consistent but not measurably better than the one-way estimator, and two POLL/DONE pairing bugs need fixing before this could merge

Ran `7e891f6` on F31 (classic ESP32 core as coordinator, plexus ESP32-S3 as peer, half-duplex RS485 at 460800) for three runs with both consoles read against one Jetson clock, plus a code review. Result: the per-sample bound `offset_<id>_accuracy` was never contradicted (319 two-second windows, zero inconsistent brackets), but it sits at 0.43–0.50 ms in the best case and 0.8–0.93 ms whenever no sample clears the 500 µs accept threshold for two seconds, and the one-way estimate from #252 sits on the lower edge of every one of those brackets. With this rig the absolute error cannot be resolved below about 2 ms (details below), so the question the PR asks — is the error within ±accuracy against independent ground truth, and clearly better than one-way — cannot be answered in favour of either estimator; the hardware does not force a decision. The review found two real bugs in the POLL/DONE pairing that #252 does not have: a runtime `enable_time_sync()`/`make_coordinator()` sends a second POLL into an in-flight DONE (measured: two poll timeouts and a garbled `__POLL__` on the peer console), and on full-duplex links that can leave every later DONE paired with the wrong T1 while the published accuracy reads 0.

If this PR is to go ahead, it needs the pairing fixed (a sequence or originate stamp in the POLL, echoed in the DONE), the `_oneway` prototype removed, and current main merged (`serial_bus.cpp`/`.h` conflict with #273; note that #252 conflicts with main as well by now). Otherwise #252 merges as parked, and the environment finding at the end applies to both.

<details>
<summary><b>Setup and method</b></summary>

- Firmware: `7e891f6` built for esp32 (core) and esp32s3 (plexus) with one test-only addition, a `core.micros` property (`esp_timer_get_time()` mirrored in `Core::step()`), not part of the PR. Core startup replaced by a minimal script (Output 15/12, `Serial(13, 4, 460800, 2)`, `SerialBus(bus_serial, 1)`, `make_coordinator(2)`, `enable_time_sync()`, output of `core.micros bus.offset_2:3 bus.offset_2_accuracy:3 bus.offset_2_oneway:3`), plexus on the Phase-D script (`Serial(39, 40, 460800, 1)`, `SerialBus(serial, 2)`, `enable_time_sync()`, `wm`/`ws`/`wd` pushed at 100 Hz).
- Runs: (1) 300 s with the 3 × 100 Hz push, (2) 180 s with `bus.enable_time_sync()` injected at 30 s, `bus.make_coordinator(2)` at 75 s and a plexus reset via RTS at 120 s, (3) 150 s with the push silenced. 60 400 core samples in total, 0 bus warnings on either side outside the provoked events.
- Host reference: the Jetson reads the core (`/dev/ttyTHS1`) and the plexus (`/dev/ttyUSB0`, CP2102N) against `monotonic_ns`. Two references were tried: NTP-style pings (`core.print('PONG', core.micros)`), which are useless here because the round trip has a 10 ms floor through the Jetson UART path and my `core.micros` is only refreshed once per main loop; and the passive minimum-latency estimate from the Phase C campaign (best `t_device − t_host` per 30 s bucket per console, difference = peer minus core). The passive reference carries the unknown asymmetry between the THS1 and the USB delivery path, which is of the order of 1–3 ms, so it can confirm drift tracking and rule out gross errors but cannot arbitrate a 0.4 ms difference.
- Internal consistency check: for every 2 s window, the intersection of all accepted `offset ± accuracy` brackets. If the airtime model and the delays-are-non-negative assumption hold, the true offset lies in that intersection; an empty intersection would falsify the bound.
- Raw logs (both consoles, every line stamped), the harness and the analysis are archived internally (F31 2026-09-01). Robot restored afterwards: core back on its production build with startup `4fb4` (rosys: local = core checksum), plexus back on the `ececaad` build it had before.

</details>

<details>
<summary><b>Results</b></summary>

| | run 1 (push 3 × 100 Hz) | run 2 (injections) | run 3 (idle bus) |
|---|---|---|---|
| samples valid / NaN | 27 949 / 203 | 16 780 / 214 | 15 041 / 214 |
| accuracy, median (min–max) | 0.77 ms (0.37–0.87) | 0.48 ms (0.36–0.85) | 0.46 ms (0.43–0.93) |
| 2 s bracket windows / contradictions | 152 / 0 | 91 / 0 | 76 / 0 |
| bracket width, median | 1.42 ms | 0.81 ms | 0.85 ms |
| one-way above bracket / below lower edge (median) | 0 / 8 µs | 0 / 5 µs | 0 / 1 µs |
| two-way − passive reference | −1.6 … −2.3 ms | −1.8 … −3.0 ms | −1.4 … −2.2 ms |
| one-way − passive reference | −2.0 … −2.6 ms | −2.2 … −3.5 ms | −2.2 … −2.9 ms |

- Two accuracy regimes, independent of bus load: 0.43–0.50 ms when samples are accepted directly, 0.8–0.93 ms when the estimate comes from the best-of-2 s window. The residual is set by the 1 ms tick of the two communication tasks, not by traffic, so the desk value `SYNC_ACCEPT_ACCURACY_US = 500` sits exactly on the edge of what this setup produces and decides which regime a run lands in.
- The one-way value equals the lower edge of the two-way bracket (it is the same quantity, `T3 − T4 + airtime`, max-filtered), so the two estimates differ by one accuracy value. Which of the two is closer to the truth depends on how the transport delay splits between the POLL and the DONE direction, which nothing in this rig observes.
- Both track the same crystal drift as the passive reference (about 5.5 ppm, −1.5 ms over 270 s in run 1).
- Events (run 2): `enable_time_sync()` at runtime → NaN, two poll timeouts, relock after 0.60 s; `make_coordinator(2)` → NaN, relock after 0.05 s (one sample); peer reset → NaN after 0.25 s, relock 2.15 s after the peer's `Ready.` (4.0 s after NaN). Initial locks took 2.2–4.0 s in all three runs because the first samples were above 500 µs, so "a single accepted sample relocks" is true but the accepted sample took up to a full 2 s window to arrive.

</details>

<details>
<summary><b>Review findings (code, verified against 7e891f6)</b></summary>

1. **CONFIRMED on hardware — `adopt_config()` clears `is_polling` while a POLL is outstanding** (`serial_bus.cpp:280`). The next loop iteration sends a second POLL while the peer's DONE is on the wire. On RS485 that is a collision: at 30.355 s the plexus logged `checksum mismatch: <<…>|$$1:2$$__POLL__`, the core two poll timeouts, and the estimate was NaN for 0.6 s. On a full-duplex link the DONE survives and is paired with the *second* POLL's T1; the residual clamps to 0, so the sample is accepted immediately with `offset_<id>_accuracy = 0` and an offset error of half a poll period, and with a single peer every later DONE stays one POLL behind. #252 resets `is_polling` the same way but its estimator does not use T1, so there it is only the collision.
2. **CONFIRMED in code — a late DONE after a poll timeout is paired with the new T1** (`serial_bus.cpp:351`). The `transport_us < 0` guard only rejects the stale stamp when the peer's processing time exceeded the new round trip. Fix for 1 and 2: a sequence number or the originate stamp T1 in the POLL payload, echoed in the DONE, and samples taken only for the matching POLL.
3. **CONFIRMED in code — the first sample after a polling pause longer than 2 s is accepted unconditionally** because `window_start_ms` is only reset on accept (`serial_bus.cpp:417`); its (large) accuracy is published honestly and the next direct sample replaces it, so this is minor.
4. **PLAUSIBLE, static only — NaN in the expander broadcast.** With `bus.broadcast()` on an expander, `Module::step` prints `!!bus.offset_3=nan;…` and the host parses `nan` as an unknown identifier, so the host logs an error every step while any estimate is invalid. Shared with #252's NaN contract; a guard in the broadcast path (skip NaN properties) would fix both.
5. **CONFIRMED — branch is 9 commits behind main and conflicts in `serial_bus.cpp`/`.h`** with #273 (`push_incoming()`, `dropped_inbound`, non-const `print_to_incoming_queue`); a PR-side resolution would drop the inbound-drop accounting.
6. Prototype residue marked `remove before merge` at eight sites (`ONEWAY_SYNC_WINDOW`, `PeerClock::oneway_*`, `OffsetUpdate::oneway_*`, the `_oneway` property, `<climits>`); `apply_offset_update` refuses to publish if that property is missing, so the removal must be complete.
7. Cost: the two new queues (~1.7 KB internal heap per `SerialBus`, also on peers and on coordinators without time sync); a 260 byte `Config` copy living in the communication task's frame on an unchanged 4 KB stack (headroom not measured); `apply_offset_update` building three `std::string` keys per update on the main task, published on every DONE even when nothing changed.
8. Style: `rfind("offset_", 0) == 0` where `starts_with()` exists, `std::numeric_limits<double>::quiet_NaN()` spelled out five times, the `+ 4` trailer length duplicating `write_checked_line`, lines over 120 columns, the docs paragraph hard-wrapped and the Properties table placed after Methods.

</details>

<details>
<summary><b>What #277 does differently from #252 (structural comparison, not re-evaluated)</b></summary>

| | #252 (`0ee3fd6`) | #277 (`7e891f6`) |
|---|---|---|
| estimator | one-way: `T3 − (T4 − airtime)`, windowed maximum over 128 poll rounds, halving convergence | two-way: four stamps, `((T2 − T1) + (T3 − T4)) / 2` with airtime asymmetry corrected, per-sample bound |
| sample filter | max filter per window | direct accept if bound ≤ 500 µs, otherwise best sample of a 2 s window; estimate snaps to the accepted sample |
| lock / relock | one full window (128 rounds, scales with peer count) | one accepted sample (0.05–2.2 s measured) |
| DONE payload | `__DONE__<t3>` | `__DONE__<t3>,<t3 − t2>` (same 18-digit cap and digits-only rule) |
| published properties | `offset_<id>` | `offset_<id>`, `offset_<id>_accuracy` (plus the `_oneway` prototype) |
| main ↔ comm task | rebuild-and-swap under `sync_mux`, polling state under the same lock | no locks: `Config` through a length-1 overwrite queue, `OffsetUpdate` snapshots back, generation counter drops stale updates |
| per-instance cost | per-peer vector only | two extra FreeRTOS queues (~1.7 KB) on every `SerialBus` |
| runtime reconfiguration | resets `poll_index`/`is_polling`, estimator insensitive to the resulting double POLL | same reset, estimator sensitive to it (findings 1 and 2) |
| base | conflicts with current main (10 commits behind) | conflicts with current main (9 commits behind, #273) |

</details>

<details>
<summary><b>Environment finding, independent of the PR: Bluetooth on the core costs bus frames at 460800</b></summary>

With the F31 production script partially running on the upstream build (it aborts at the first private module, but `Bluetooth("feldfreund-f31")` is created before that), the coordinator saw about 15 checksum mismatches per 10 s plus poll timeouts at 460800 with the 3 × 100 Hz push; the frames themselves carried valid checksums, they arrived merged or with bytes missing, i.e. RX loss on the core. With the minimal startup (no Bluetooth) the same traffic ran clean for the whole campaign. That points at interrupt latency from the BLE controller on the classic ESP32 rather than at either PR, but it matters for any fleet configuration that combines `SerialBus` at high baud rates with Bluetooth on the core; worth a separate issue.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Measurement%20plan%2C%20harness%2C%20review%20findings%20and%20wording%20by%20the%20model%3B%20the%20robot%20work%20was%20supervised%20and%20the%20conclusions%20were%20checked%20by%20the%20author.&d=2026-09-01&a=claude-code "claude-fable-5: Measurement plan, harness, review findings and wording by the model; the robot work was supervised and the conclusions were checked by the author.")

falkoschindler:
@JensOgorek thanks for the F31 campaign and the review. The branch is updated and pushed (`554f6e7`, `7eaaa23`, `941bb34`; net +408/−36 against main). We go ahead with this PR for the structural reasons in the description (relock in one sample, no locks, convergence independent of the peer count); #252 stays parked and gets closed when this merges.

Per finding:

1. and 2. POLL/DONE pairing: the coordinator sends `__POLL__<seq>` and the peer echoes the sequence number as the third DONE field, so a sample is only taken for the POLL it answers and a late DONE after a timeout falls through. `adopt_config()` no longer clears `is_polling`; a reconfiguration lets the outstanding POLL finish. The sequence number goes only to peers that have answered with a stamped DONE, so a legacy peer keeps getting the bare POLL and keeps working.
3. A window now starts with its first sample, so the first sample after a polling pause is no longer accepted unseen. An unlocked estimate also takes the first sample whatever its bound, which should remove the 2–4 s initial lock you measured.
4. NaN in the broadcast: we first did what you suggested and skipped NaN properties in `Module::step()` (`554f6e7`), then took it out again (`941bb34`). The skip hides the value instead of transporting it: a proxied offset kept its last estimate and accuracy across a peer reboot with no way for the host to notice, and the same applied to every other module (`temperature_sensor.temperature_c` on an unplugged NTC). The root cause is that the grammar has no `nan`, so #291 adds `nan` and `inf` as literals. **This PR depends on #291**: merge #291 first, then main into this branch. Until then a NaN offset produces the `unknown variable "nan"` error on the host that NaN properties have always produced. Out of scope and documented: a host Proxy only learns dynamic properties with the first broadcast, so a startup script cannot reference `bus.offset_<id>` behind an expander.
5. main merged (`cbb9815`), the #273 conflict resolved keeping the inbound-drop accounting.
6. The `_oneway` prototype is removed completely.
7. Cost: the offset queue went from 32 to 8 entries with a retry when full (snapshots are published on accept and reset only), the `Config` is received into a member instead of the task's stack, and the main task caches its property handles.
8. Style: `starts_with()`, `NAN`, `Serial::CHECKSUM_TRAILER_LENGTH`, lines under 120 columns, docs one sentence per line with Properties before Methods.

A self-review pass (`7eaaa23`) added small things: one DONE buffer sized from the stamp constants, the round robin starts with the first listed peer after a reconfiguration, the constructor cleans up in one place. Deliberately left alone: a sync-enabled peer stamps its DONE even for a bare POLL, so a coordinator rolled back to older firmware stalls 250 ms per poll (soft, the documented update order covers it, and the alternative would leave legacy peers unreachable from a new coordinator); a timeout clears `sync_capable`, which costs one extra bare round per lost frame but is what the legacy fallback needs.

What we would ask you to re-verify on hardware with `941bb34`: the run-2 injections (`enable_time_sync()` and `make_coordinator()` at runtime, peer reset) to confirm that the relock takes one to two poll rounds without a timeout and that no second POLL goes out, and one mixed-version round (new coordinator, old peer) for the bare-POLL fallback. If the setup involves a proxied bus, the core needs #291 as well. The Bluetooth/RS485 frame loss deserves its own issue and is not touched here. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Branch%20summary%20and%20wording%20by%20the%20model%3B%20the%20decisions%20on%20scope%2C%20the%20NaN%20handling%20and%20the%20split%20into%20%23291%20from%20the%20author%27s%20review%20discussion.&d=2026-09-09&a=claude-code "claude-fable-5-1: Branch summary and wording by the model; the decisions on scope, the NaN handling and the split into #291 from the author's review discussion.")

JensOgorek (approved):
### Hardware re-test of `941bb34` on F30 passed: relock in one bare plus one sequenced round, 0 double POLLs in 29 777, legacy fallback clean — approving, merge #291 first

Ran the requested re-checks on the four-arm robot F30 (classic-ESP32 core as coordinator, two plexus ESP32-S3 as peers, RS485 at 460800): runtime `enable_time_sync()` and `make_coordinator()` in both peer orders, an RTS reset of a peer, and mixed firmware in both directions. Every claim from the rework holds: after each injection the outstanding POLL finishes first, then one bare POLL proves the peer capable and the next sequenced POLL relocks the estimate (`offset_<id>` is NaN for 5 to 20 ms, no timeout), a rebooted peer relocks 10 ms after its `Ready.`, and a wire sniffer saw no second POLL to a peer while its DONE was outstanding. The two pairing findings from the F31 review are closed. Good to merge once #291 is in and main is merged into this branch; the code-level points below are non-blocking.

<details>
<summary>Setup</summary>

- F30 (rb68): core = classic ESP32, `bus_serial = Serial(13, 4, 460800, 2)`, `SerialBus(bus_serial, 1)`, `make_coordinator(4, 5)`, `enable_time_sync()`, `core.output` with `core.millis`, both offsets and accuracies at 10 ms; console 460800. Peers 4 and 5 = plexus ESP32-S3 on `Serial(39, 40, 460800, 1)`, `SerialBus(serial, N)`, `enable_time_sync()`, plus a `bus.send(1, "!!pN_parked=%s", …)` push every 100 ms as bus load. Peers 2 and 3 were not polled.
- Firmware: upstream lizard at `941bb34` on core and peers, plus a test-only `bus.sniff(true)` that prints every POLL/DONE frame a node receives on its console; peer 5 served as the wire sniffer (console at 460800, ~450 lines/s), so all pairing numbers are for peer 4 (a node cannot see its own DONE).
- All console lines of the three nodes were logged with host stamps; relock times come from the core line, pairing from the sniffer.
</details>

<details>
<summary>Runtime injections and peer reset (200 s)</summary>

Injections: `enable_time_sync()` at 30 s, `make_coordinator(4, 5)` at 60 s, `make_coordinator(5, 4)` at 90 s, RTS reset of peer 4 at 120 s, `enable_time_sync()` at 160 s.

| | result |
| --- | --- |
| POLLs to peer 4 on the wire | 29 777 (12 bare), DONEs 29 768 |
| second POLL to peer 4 before its DONE (< 250 ms) | 0 |
| DONE sequence mismatches | 0 |
| timeouts | 9, all during the 2.1 s reboot of peer 4 |
| relock after each injection | outstanding POLL finishes, 1 bare POLL (stamped DONE, seq 0), 1 sequenced POLL = sample; NaN for 5 to 20 ms, no timeout |
| relock after peer reset | NaN after the first timeout (+268 ms), bare POLLs every 250 ms while the peer boots, valid 10 ms after `Ready.` |
| relock sample vs. previous estimate | 21 to 127 µs (peer 4), ≤ 491 µs (peer 5), inside the published bounds |
| `offset_4` segment medians across the reconfigurations | -350548.479 / .482 / .482 / .479 ms; after the reset -473205.455 / .460 |
| accuracy median | 0.35 ms (peer 4), 0.34 ms (peer 5; 0.76 ms in the last 80 s while it printed the sniff at 20 KB/s) |
| POLL to DONE on the wire (peer 4) | median 1.82 ms, p90 2.16 ms, max 3.33 ms |
| core loop / heap | 10 ms median (max 19 ms), heap constant at 132 656 B over five reconfigurations |

Wire trace of the `make_coordinator(5, 4)` injection as the sniffer saw it (sequence numbers in the payload): `POLL 1->4 81811`, `DONE 4->1 …,81811`, `POLL 1->5 81812`, `POLL 1->5 81813`, `POLL 1->4` (bare), `DONE 4->1 …,0`, `POLL 1->5 81814`, `POLL 1->4 81815`, `DONE 4->1 …,81815` → core valid again.
</details>

<details>
<summary>Mixed firmware, both directions</summary>

- New coordinator, legacy peer (peer 5 on the stock firmware, 60 s): peer 5 keeps being polled with bare POLLs, zero timeouts after its boot, `offset_5` stays NaN as documented; peer 4 (new) locks 0.5 s after its `Ready.` with a 0.39 ms bound.
- New peer rebooting into a startup without `enable_time_sync()` while the coordinator held it sync-capable (30 s): only the reboot's own timeouts, then 29 s clean. The reboot timeout already clears `sync_capable`, so the fallback costs no extra round.
- Old coordinator, new peers with `enable_time_sync()` (the documented "update the coordinator first" case, 30 s): 120 poll timeouts and 120 `error: unexpected token … ","` on the core, i.e. every poll runs into the 250 ms limit and the stamped DONE is executed as a command. The docs cover it; the number is here so the update order is not treated as optional.
</details>

<details>
<summary>Code review (non-blocking)</summary>

Worth a look before or right after merging:

- Heap on every node: `Config` is 260 B (`uint8_t peer_ids[254]`) and exists three times (`config`, `received_config`, the length-1 queue) plus the offset queue, about 1.1 KB per `SerialBus` instance, paid by peers that never call `make_coordinator()`. A realistic cap (32 peers?) shrinks it to ~200 B; creating the two new queues lazily in `make_coordinator()`/`enable_time_sync()` would make peers pay nothing.
- The same cap covers a cliff: two properties per peer make the `!!` broadcast of a proxied coordinator exceed `Module::step()`'s 1024-byte buffer at about 16 peers, where `csprintf` throws `buffer too small` on every step and nothing of the module is broadcast any more. Worth a sentence in the Scope bullet of the docs.
- A bare `__DONE__` answering a sequenced POLL is dropped (`seq != poll_seq`), so the poll ends by the 250 ms timeout before `sync_capable` is cleared. Only reachable without a reboot (a board swapped in under the same id, or duplicate ids); the reboot path is clean as measured. Cheap to accept such a DONE from the polled peer as the poll's end while skipping the sample.

Design notes, no change requested:

- The window regime accepts the best sample of 2 s even when its bound is worse than the locked estimate's, so during an OTB update or another TX burst a 0.4 ms estimate can be replaced by a 5 ms one every 2 s instead of drifting ~40 µs. Comparing the candidate's bound against the locked one plus a drift allowance would keep the good estimate.
- T3 is stamped after `send_outgoing_queue()`, so a peer that sends data on every poll adds that airtime to the residual and its bound floors at half of it (60 B at 460800 ≈ 0.65 ms, above `SYNC_ACCEPT_ACCURACY_US`). Our peers pushed 20 B per 100 ms and reached 0.33 ms, so it does not show here.
- `__POLL__` with 11 to 18 digits is delivered as a user payload (the `seq <= UINT32_MAX` check), while the docs reserve every `__POLL__<digits>`.
- Style: the three main-task methods sit between the member groups in `serial_bus.h`; the comm task unpacks the received `Config` into three members that mirror it; `peer_clocks` is filled without time sync, which costs three `time_sync_enabled` guards; `parse_stamp()`'s empty-string sentinel (-1) is decoded at every caller.
</details>

<details>
<summary>Not covered</summary>

- Accuracy against an independent clock: the rig cannot resolve below the RS485 turnaround (F31 campaign), only self-consistency was checked again.
- A proxied bus behind an expander (needs #291 on the host).
- A late DONE arriving after a timeout was not provoked; the sequence check covers it by construction and the sniffer saw no mismatch.

Test-only additions (the `bus.sniff` patch and the F30 harness) are not proposed for the PR.
</details>

Reviewed and tested with Claude Fable 5.1 (`claude-fable-5-1`) in Claude Code; the hardware campaign, the measurements and the code review are the model's work, the verdict and the decision to approve are mine.

PR #280, opened 2026-08-21
Back off polling to unreachable serial bus peers

### LINKED ISSUE #279: Polling backoff for unreachable serial bus peers
## Problem

The coordinator polls its peers round-robin. A poll to an unreachable peer (unpowered, unplugged, crashed) only ends after `POLL_TIMEOUT_MS = 250`, and during that time `send_outgoing_queue()` does not run (`communication_loop` gates it on `!is_polling`). The outbound queue is shared by all receivers, so every `bus.send()` issued in that window — including those addressed to healthy peers — piles up. A `when` rule sending once per step (~100 msg/s) produces ~25 messages per stalled poll; with two unreachable peers or two sends per step the backlog doubles. Once the queue is full, `enqueue_outgoing_message` blocks 50 ms and throws on every step (`error in rule`), the message to the healthy peer is lost, and the main loop drops to ~20 Hz.

Increasing the queue length only moves the threshold (found while reviewing #273): the scenario has no upper bound, so it needs a fix in the polling, not in the buffering.

## Proposal

After a poll to a peer times out, skip that peer for a backoff interval (e.g. 1 s, possibly growing) before polling it again, and report once when a peer becomes unreachable / reachable again. Healthy peers then keep being served at the normal round-trip rate of a few ms, and the outbound queue depth stays at the ≤ 2 the rate analysis in #273 predicts.

Open questions: how quickly a returning peer should be picked up again (it cannot announce itself — broadcasts are not implemented yet), and whether the first poll after boot should be exempt so the "Ready." handshake is not delayed.

evnchn:

### Motivation

A poll to an unreachable peer stalls the round-robin for the full 250 ms timeout on every cycle, so a single unplugged peer throttles all healthy peers. Implements the polling-backoff proposal from the issue.

Closes #279.

### Implementation

After a timeout, that peer is skipped for a backoff window (1 s, doubling to a cap of 8 s) before being polled again, and reachability transitions are reported once instead of on every retry. Healthy peers keep being served at the normal round-trip rate regardless of how many peers are down.

<details>
<summary><b>Design notes / answers to the issue's open questions</b></summary>

- **First poll after boot is exempt from backoff automatically** — a peer only enters backoff after its first observed timeout (`backoff_duration_ms` starts at 0, which reads as "not backed off"), so the boot-time "Ready." handshake isn't delayed by this change.
- **How fast a returning peer is picked up**: purely time-based — once its backoff window (which grows 1s → 2s → 4s → capped at 8s while it keeps failing, and resets to 0 the moment it responds) elapses, the coordinator polls it again on the next round-robin pass. There's no out-of-band "peer came back" signal since broadcasts aren't implemented yet, so this seemed like the simplest fit; happy to tune the constants or add jitter if 1s/8s isn't the right range for real deployments.
- If **every** peer is currently backed off, the coordinator simply skips polling that tick rather than busy-spinning a poll that would only time out again.
- Shared coordinator state (`peer_ids`, `peer_poll_states`, `poll_index`, `is_polling`) is guarded by a `portMUX` spinlock against `make_coordinator` running on the other core; c411d34 closes a remaining cancellation race that was reproduced on hardware (details in [this comment](https://github.com/zauberzeug/lizard/pull/280#issuecomment-5369404818)) and keeps all reporting outside the critical sections.

</details>

### Progress

- [x] Documentation has been updated (or is not necessary — internal scheduling behavior; the module reference already describes polling at its current level of detail).

<details>
<summary><b>Verification</b></summary>

- `source ~/esp/esp-idf/export.sh && python3 build.py esp32` → `Build completed successfully for esp32`
- `source ~/esp/esp-idf/export.sh && python3 build.py esp32s3` → `Build completed successfully for esp32s3`
- `pre-commit run --files main/modules/serial_bus.cpp main/modules/serial_bus.h` → all hooks pass
- **Hardware-tested on an ESP32-S3**: two `SerialBus` nodes on one board (UART1/UART2 cross-wired through the GPIO matrix), coordinator polling a healthy peer plus a nonexistent one — backoff, boot exemption, one-shot unreachable/reachable-again reporting and the all-backed-off skip all observed on real UART traffic. The `make_coordinator` cancellation race was reproduced (437 spurious "unreachable" reports for a healthy peer in 731 bursts) and is gone after c411d34 (0 in 731, same protocol); full numbers in [this comment](https://github.com/zauberzeug/lizard/pull/280#issuecomment-5369404818).
- Got a second-lineage (Codex/GPT-5.2) adversarial review on the diff before opening this PR. It confirmed the core throughput claim holds, and caught one real gap I fixed before pushing: `make_coordinator()` replaced `peer_ids`/`peer_poll_states` without resetting `poll_index`/`is_polling`, so a second call while a poll was in flight could index past the freshly resized (possibly shorter) backoff-state vector. Fixed by clearing `is_polling` and resetting `poll_index` in the same place the peer list is replaced. A second Codex pass on c411d34 found no remaining blockers.

</details>

### DIFF
--- a/main/modules/serial_bus.cpp
+++ b/main/modules/serial_bus.cpp
@@ -136,6 +136,7 @@ void SerialBus::call(const std::string method_name, const std::vector<ConstExpre
         this->poll_index = 0;
         this->peer_ids.swap(new_peer_ids);
         this->peer_poll_states.swap(new_peer_poll_states);
+        this->coordinator = true;
         portEXIT_CRITICAL(&this->config_mux);
     } else {
         Module::call(method_name, arguments);
@@ -156,22 +157,34 @@ void SerialBus::call(const std::string method_name, const std::vector<ConstExpre
                 next_index = bus->next_pollable_peer();
                 have_peer = next_index < bus->peer_ids.size();
                 if (have_peer) {
+                    // commit the poll while still holding the mux so make_coordinator
+                    // cannot cancel it between selection and is_polling = true
                     bus->poll_index = next_index;
                     peer_id = bus->peer_ids[next_index];
+                    bus->poll_start_millis = millis();
+                    bus->is_polling = true;
                 }
                 portEXIT_CRITICAL(&bus->config_mux);
                 if (have_peer) {
                     bus->send_message(peer_id, POLL_CMD, sizeof(POLL_CMD) - 1);
-                    bus->poll_start_millis = millis();
-                    bus->is_polling = true;
                 }
             }
-            // handle poll timeout
+            // handle poll timeout (re-check is_polling under the mux: make_coordinator may
+            // have cancelled the poll after the unguarded pre-check)
             if (bus->is_polling && millis_since(bus->poll_start_millis) > POLL_TIMEOUT_MS) {
+                uint8_t peer_id = 0;
+                unsigned long backoff_ms = 0;
+                bool report = false;
                 portENTER_CRITICAL(&bus->config_mux);
-                bus->handle_poll_timeout();
-                bus->is_polling = false;
+                if (bus->is_polling && millis_since(bus->poll_start_millis) > POLL_TIMEOUT_MS) {
+                    report = bus->handle_poll_timeout(peer_id, backoff_ms);
+                    bus->is_polling = false;
+                }
                 portEXIT_CRITICAL(&bus->config_mux);
+                if (report) {
+                    bus->print_to_incoming_queue(
+                        "warning: serial bus %s: peer %u unreachable, backing off %lu ms", bus->name.c_str(), peer_id, backoff_ms);
+                }
             }
         } else {
             // respond to poll
@@ -229,12 +242,18 @@ void SerialBus::process_uart() {
 
         // handle done command
         if (std::strcmp(message.payload, DONE_CMD) == 0) {
+            uint8_t peer_id = 0;
+            bool report = false;
             portENTER_CRITICAL(&this->config_mux);
-            if (this->is_coordinator() && message.sender == this->peer_ids[this->poll_index]) {
+            // is_polling filters DONEs from polls that make_coordinator cancelled
+            if (this->is_polling && this->is_coordinator() && message.sender == this->peer_ids[this->poll_index]) {
                 this->is_polling = false;
-                this->handle_poll_success();
+                report = this->handle_poll_success(peer_id);
             }
             portEXIT_CRITICAL(&this->config_mux);
+            if (report) {
+                this->print_to_incoming_queue("serial bus %s: peer %u reachable again", this->name.c_str(), peer_id);
+            }
             continue;
         }
 
@@ -263,26 +282,34 @@ size_t SerialBus::next_pollable_peer() const {
     return peer_count;
 }
 
-void SerialBus::handle_poll_success() {
+// must be called under config_mux; returns true if the peer just became reachable again
+// (reporting happens outside the critical section, so the caller prints)
+bool SerialBus::handle_poll_success(uint8_t &peer_id) {
     PeerPollState &state = this->peer_poll_states[this->poll_index];
     state.backoff_duration_ms = 0;
     state.backoff_start_millis = 0;
     if (state.unreachable) {
         state.unreachable = false;
-        this->print_to_incoming_queue("serial bus %s: peer %u reachable again", this->name.c_str(), this->peer_ids[this->poll_index]);
+        peer_id = this->peer_ids[this->poll_index];
+        return true;
     }
+    return false;
 }
 
-void SerialBus::handle_poll_timeout() {
+// must be called under config_mux; returns true if the peer just became unreachable
+// (reporting happens outside the critical section, so the caller prints)
+bool SerialBus::handle_poll_timeout(uint8_t &peer_id, unsigned long &backoff_ms) {
     PeerPollState &state = this->peer_poll_states[this->poll_index];
     state.backoff_duration_ms = state.backoff_duration_ms == 0 ? POLL_BACKOFF_INITIAL_MS
                                                                : std::min(state.backoff_duration_ms * 2, POLL_BACKOFF_MAX_MS);
     state.backoff_start_millis = millis();
     if (!state.unreachable) {
         state.unreachable = true;
-        this->print_to_incoming_queue(
-            "warning: serial bus %s: peer %u unreachable, backing off %lu ms", this->name.c_str(), this->peer_ids[this->poll_index], state.backoff_duration_ms);
+        peer_id = this->peer_ids[this->poll_index];
+        backoff_ms = state.backoff_duration_ms;
+        return true;
     }
+    return false;
 }
 
 bool SerialBus::parse_message(const char *message_line, IncomingMessage &message) const {
--- a/main/modules/serial_bus.cpp
+++ b/main/modules/serial_bus.cpp
@@ -17,6 +17,8 @@ extern void process_line(const char *line, const int len);
 
 static constexpr size_t FRAME_BUFFER_SIZE = 512;
 static constexpr unsigned long POLL_TIMEOUT_MS = 250;
+static constexpr unsigned long POLL_BACKOFF_INITIAL_MS = 1000;
+static constexpr unsigned long POLL_BACKOFF_MAX_MS = 8000;
 static constexpr size_t OUTGOING_QUEUE_LENGTH = 32;
 static constexpr size_t INCOMING_QUEUE_LENGTH = 32;
 static_assert(OUTGOING_QUEUE_LENGTH > otb::BUS_OTB_WINDOW); // a stalled poll must not abort an OTB update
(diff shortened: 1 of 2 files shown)


evnchn:
Well this is not supposed to go here (should go to my fork first) but anyways... 

@falkoschindler As you always prefer me to do, I will review this with Fable and hardware verification.

evnchn:
**c411d34 fixes a hardware-reproduced race: `make_coordinator` could cancel an in-flight poll and then be overwritten by the communication task, falsely marking a healthy peer unreachable — 437 spurious events in 731 bursts before the fix, 0 after.**

<details>
<summary>Hardware verification + what the commit changes</summary>

**Rig:** single ESP32-S3, two `SerialBus` nodes (coordinator node 1, healthy peer node 2) on UART1/UART2 cross-wired internally through the GPIO matrix — a real cross-core race (main task on core 0, communication task on core 1) over real UART frames, no physical wiring. A 15 ms delay at the racy spot (same source site in both builds) widens the microsecond window so a 5-minute run is conclusive. The hammer sends `make_coordinator(2, 9)` pair-bursts while peer 2 answers every poll, so any "peer 2 unreachable" warning can only come from the race.

| build | spurious "peer 2 unreachable" | legit peer-9 timeouts | bursts |
|---|---|---|---|
| 7ba354c (before) | 437 | 294 | 731 |
| c411d34 (after) | 0 | 731 | 731 |

437 + 294 = 0 + 731: every mislabeled timeout became a correctly labeled one, same traffic level in both runs.

**What c411d34 changes:**
- commit `poll_index` / `poll_start_millis` / `is_polling = true` inside the selection critical section; only the UART write stays outside
- re-check the poll timeout under the mux before acting on it
- ignore DONE frames of polls that `make_coordinator` has already cancelled
- `handle_poll_success` / `handle_poll_timeout` now only mutate state under the mux and return report data; `print_to_incoming_queue` (vsnprintf + `xQueueSend`) moves outside the critical sections
- `is_polling` / `poll_start_millis` and a new `coordinator` flag become atomic, so the communication task's unguarded pre-checks are race-free against `make_coordinator` on the other core

Both esp32 and esp32s3 targets build locally; a second-model adversarial pass on the diff found no remaining blockers.
</details>

evnchn:
@falkoschindler end of my review and hardware verification then. You can take a look and double-check as well 🙏

falkoschindler:
@evnchn thank you for the thorough work here — the implementation is careful, the hardware verification is exemplary, and a deep review confirmed that the locking design holds up end-to-end. Still, after weighing the change against where the problem actually stands today, I'd like to propose closing this PR without merging and keeping #279 in *someday* for when it becomes acute.

To be clear, this is not a verdict on the idea — the proposal in #279 is my own, and the analysis there stands: on a bus with several peers, a stalled poll blocks `send_outgoing_queue` for all receivers, so one dead peer degrades service to healthy peers without bound, and backoff is the right kind of fix. But that scenario needs a multi-peer bus, and today all our deployments run a coordinator with a single peer, where a missing peer means a non-operational robot anyway; the failure mode is so far derived from the #273 rate analysis, not something we have hit in the field. Running several Plexus on one bus is on the roadmap, so this will matter eventually — just not yet.

Meanwhile the feature creates the most significant issue our review found: a slow-but-alive peer — one that needs longer than the 250 ms timeout to drain a full outgoing queue, ~740 ms at 115200 baud — times out on every poll, its late DONE is discarded, and it escalates to the permanent 8 s backoff, misreported as unreachable while answering 100% of polls (finding 1 below). And in the most common real event, an ESP rebooting, the backoff delays re-discovery by up to 8 s where main re-polls immediately. So merged today, the change would make the cases that do occur slightly worse while the case it improves doesn't occur yet.

The complexity also lands in an already delicate module: the cross-task shared state needed two follow-up race-fix commits within this very PR, and the (currently correct) synchronization rests on comment-enforced invariants — a maintenance burden I'd rather take on when the payoff is real.

So my suggestion is to park this until multi-peer buses are actually on the table, and then restart from this branch and the findings below — with finding 1 resolved (e.g. by treating any DONE as proof of reachability, or sizing the timeout to a full queue drain), and possibly weighing whether a fixed retry interval without escalation and latch buys most of the benefit at less complexity. The hardware-verification setup you built here will be just as valuable then.

<details>
<summary>Full review findings (for reference)</summary>

## 1. A slow-but-alive peer escalates to permanent 8 s backoff and is misreported as unreachable (medium)

main/modules/serial_bus.cpp (249): a DONE arriving after POLL_TIMEOUT_MS (250 ms) is dropped by the `is_polling` gate without touching `peer_poll_states`, so `handle_poll_success` never runs for it. A responder drains its whole outgoing queue before sending DONE, and a near-full queue (32 × ~270 framed bytes ≈ 8.5 KB) takes ~740 ms at the default 115200 baud — well over the timeout. Such a peer times out on every poll: backoff doubles 1 s → 2 s → 4 s → 8 s with no recovery path (for response latencies in the realistic (250 ms, 1250 ms] range the late DONE always lands while `is_polling` is false), the peer is reported "unreachable" although it answers 100% of polls, and its outbound data is throttled to one queue-drain per ~8.25 s. The effect is self-reinforcing: longer poll intervals mean fuller queues and longer drains. On main a slow peer only cost a warning per round and was re-polled immediately, so this is a regression for busy-but-alive peers. Possible directions: treat a DONE received while not polling as proof of reachability and clear that peer's backoff (the sender id identifies the peer regardless of `poll_index`), or make the timeout generous enough to cover a full queue drain at the configured baud rate.

## 2. Backoff escalation is silent, and the one-shot report can be lost (low)

main/modules/serial_bus.cpp (186): only the first timeout is reported, always with the initial 1000 ms; the doublings to 8000 ms and all repeat timeouts leave no trace, so the only log line understates the steady-state polling gap by 8×. Additionally, the transition reports travel through `print_to_incoming_queue`, which silently drops on a full inbound queue (aggregate counter only); since `unreachable` is latched, a dropped report is never re-emitted — main's per-round warning self-healed. Both are edge cases (the one-shot design is clearly intentional anti-spam, and a full queue at the transition instant is unlikely since a dying peer reduces inbound traffic), but one cheap change addresses both: also emit a report when the backoff reaches POLL_BACKOFF_MAX_MS.

## 3. Redundant heap copy in make_coordinator (simplification)

main/modules/serial_bus.cpp (129): `std::vector<uint8_t> new_peer_ids = peers;` copies the freshly built local only to swap it in; `peers` has no use after the swap. Swapping `peers` directly is behavior-identical and still deallocates the old vector outside the critical section (it dies at scope exit, after `portEXIT_CRITICAL`). Saves an allocation on a heap-constrained target and removes a reader's why-the-copy question.

## 4. PeerPollState::unreachable is derivable state (simplification)

main/modules/serial_bus.h (46): the invariant `unreachable == (backoff_duration_ms != 0)` holds everywhere — both fields are written only in `handle_poll_success` and `handle_poll_timeout`, always in lockstep, and initial/reset state is zero/false. The bool can be dropped and the transition reports keyed on the pre-mutation duration being zero/nonzero. Two fields encoding one fact invite a future edit that updates one but not the other, silently breaking the one-shot reporting.

## 5. Parallel vectors could be one vector of structs (design, optional)

main/modules/serial_bus.h (50): `peer_ids` and `peer_poll_states` are parallel vectors whose same-size/same-index invariant lives in a comment. Nothing needs `peer_ids` as a standalone contiguous vector (all uses are element indexing, size, and the swap), so a single `std::vector<Peer>` with `{id, backoff_start_millis, backoff_duration_ms}` would make the invariant structural — a future writer of one vector who forgets its twin causes silent OOB indexing in a critical section on ESP32.

## 6. Redundant is_coordinator() in the DONE guard (simplification)

main/modules/serial_bus.cpp (249): under `config_mux`, `is_polling == true` already implies `coordinator == true` — `is_polling` is only set true in the coordinator branch, `coordinator` never goes back to false, and `make_coordinator` clears `is_polling` in the same critical section that sets `coordinator`. The extra check implies an impossible state and can be dropped.

## 7. Hoist millis() out of the per-peer backoff loop (efficiency, minor)

main/modules/serial_bus.cpp (278): `next_pollable_peer` reads the hardware timer once per peer inside the critical section (interrupts disabled). One `const unsigned long now = millis();` before the loop is semantically equivalent and shortens the interrupt-off window. Negligible at 1–2 peers; cheap to do anyway.

## 8. Conventions

- Two new lines exceed the 120-character limit from CONTRIBUTING.md: main/modules/serial_bus.cpp (186) is 131 chars and (304) is 126 chars. Not enforced by clang-format (ColumnLimit 0), so it falls to review.
- main/modules/serial_bus.h (87): the new `std::atomic<bool> coordinator` member is declared between two methods; the file groups variables at lines 49–69, and its natural home is next to `is_polling`/`poll_index` where its config_mux comment also lands beside the mux it references.
- docs/module_reference.md (88) still says the coordinator "polls peers in turn" — the new 1–8 s backoff behavior and the two new operator-facing messages are undocumented.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Substance%20from%20the%20maintainer%27s%20assessment%20plus%20the%20model%27s%20multi-agent%20review%20findings%3B%20wording%20by%20the%20model.&a=claude-code "claude-fable-5: Substance from the maintainer's assessment plus the model's multi-agent review findings; wording by the model.")

evnchn:
Agreed. Let's just close this. 

And I didn't think about slow but reachable peer such as your "250 ms timeout to drain a full outgoing queue, ~740 ms at 115200 baud". 

I should've connected the lines with my past baud rate research as well.

PR #281, opened 2026-08-21
Bump owl so failed allocations while parsing return an error instead of aborting

### LINKED ISSUE #240: Parser can still abort/crash under low heap on inputs #239 doesn't cover

falkoschindler:
### Motivation

Under low heap, owl's generated parser could `abort()` or dereference a failed allocation while building the tree, before `process_lizard()` gets to check `owl_tree_get_error()` — so the ESP rebooted instead of dropping the line (#240). The length-based guard from #241 that tried to keep owl out of that state was reverted in #278 because no length-based estimate covers both real startup scripts and expression-dense lines (#272).

The fix is now upstream: ianh/owl#53 makes every allocation reached from `owl_tree_create_from_string()` report `ERROR_ALLOCATION_FAILURE` — the construction-phase allocators, the parse-tree growth, the string-contents buffer (which previously wrote past its end), both stacks, and the tree itself (`create` returns `NULL`, `owl_tree_get_error(NULL, …)` reports the failure). ianh merged it as 1b8b7f4.

### Implementation

- Point the `owl` submodule at ianh/owl 1b8b7f4 (from ae33e64; also picks up `OWL_PARSER_TYPEDEFS` and an out-of-bounds-read fix)
- Regenerate `main/parser.h` with `./gen_parser.sh`; the 9 remaining `abort()` calls are internal invariant checks, none of them allocation-related
- No firmware change: `process_lizard()` handles `ERROR_ALLOCATION_FAILURE` since #239, and `owl_tree_destroy` accepts `NULL`
- Correct the AGENTS.md note that called `parser.c` the generated file (it is `parser.h`)

Verified on the host with a wrapped-allocator harness against the regenerated `parser.h`: parse each of nine inputs (short and long statements, a 1239-allocation flat expression, escaped strings, nested parentheses, three invalid lines) once per allocation with that allocation failing, and once more with every allocation from that point on failing, under ASan/UBSan with leak accounting — every case returns `ERROR_ALLOCATION_FAILURE`, no crash, no leaked block. Before the bump, 26 of the 27 allocations of `x = 1 + 2` crashed.

Closes #240.

### Progress

- [x] Tested on the host (allocation-failure sweep) and built for esp32; no hardware test needed since the firmware code is unchanged.
- [x] Documentation has been updated (AGENTS.md).

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5&t=Substance%20from%20the%20conversation%20with%20the%20maintainers%2C%20wording%20from%20the%20model.&a=claude-code "claude-fable-5: Substance from the conversation with the maintainers, wording from the model.")

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,7 +102,7 @@ lizard/
 ├── espresso.py            # Flash & control script (ESP32 / ESP32-S3)
 ├── monitor.py             # Serial monitor
 ├── language.owl           # Lizard grammar definition (Owl parser generator)
-└── gen_parser.sh          # Regenerates parser.c from language.owl
+└── gen_parser.sh          # Regenerates parser.h from language.owl
 ```
 
 ### Key Entry Points
@@ -367,7 +367,7 @@ Always call parent implementations:
 
 ### Parser Regeneration
 
-After modifying `language.owl`, always run `./gen_parser.sh`. The `parser.c` file is **generated** – never edit it directly.
+After modifying `language.owl`, always run `./gen_parser.sh`. The `parser.h` file is **generated** – never edit it directly.
 
 ---
 
--- a/main/parser.h
+++ b/main/parser.h
@@ -281,6 +281,34 @@ struct parsed_number {
     double number;
 };
 
+#ifdef OWL_PARSER_TYPEDEFS
+typedef enum owl_error owl_error;
+typedef struct owl_ref owl_ref;
+typedef enum parsed_type parsed_type;
+typedef struct parsed_statements parsed_statements;
+typedef struct parsed_statement parsed_statement;
+typedef struct parsed_actions parsed_actions;
+typedef struct parsed_action parsed_action;
+typedef struct parsed_noop parsed_noop;
+typedef struct parsed_constructor parsed_constructor;
+typedef struct parsed_property_assignment parsed_property_assignment;
+typedef struct parsed_variable_assignment parsed_variable_assignment;
+typedef struct parsed_variable_declaration parsed_variable_declaration;
+typedef struct parsed_rule_definition parsed_rule_definition;
+typedef struct parsed_schedule_definition parsed_schedule_definition;
+typedef struct parsed_routine_definition parsed_routine_definition;
+typedef struct parsed_routine_call parsed_routine_call;
+typedef struct parsed_await_condition parsed_await_condition;
+typedef struct parsed_await_routine parsed_await_routine;
+typedef struct parsed_method_call parsed_method_call;
+typedef struct parsed_datatype parsed_datatype;
+typedef struct parsed_expression parsed_expression;
+typedef struct parsed_identifier parsed_identifier;
+typedef struct parsed_string parsed_string;
+typedef struct parsed_integer parsed_integer;
+typedef struct parsed_number parsed_number;
+#endif
+
 struct parsed_statements parsed_statements_get(struct owl_ref);
 struct parsed_statement parsed_statement_get(struct owl_ref);
 struct parsed_actions parsed_actions_get(struct owl_ref);
@@ -371,9 +399,13 @@ static bool grow_tree(struct owl_tree *tree, size_t size)
 }
 static void write_tree(struct owl_tree *tree, uint64_t value)
 {
+    if (tree->error != ERROR_NONE)
+        return;
     size_t reserved_size = tree->next_offset + RESERVATION_AMOUNT;
-    if (tree->parse_tree_size <= reserved_size && !grow_tree(tree, reserved_size))
-        abort();
+    if (tree->parse_tree_size <= reserved_size && !grow_tree(tree, reserved_size)) {
+        tree->error = ERROR_ALLOCATION_FAILURE;
+        return;
+    }
     while (value >> 7 != 0) {
         tree->parse_tree[tree->next_offset++] = 0x80 | (value & 0x7f);
         value >>= 7;
@@ -1280,10 +1312,10 @@ static size_t finish_token(uint32_t rule, size_t next_sibling, void *info) {
     return offset;
 }
 static void check_for_error(struct owl_tree *tree) {
-    if (tree->error == ERROR_NONE)
+    if (tree && tree->error == ERROR_NONE)
         return;
     fprintf(stderr, "parse error: ");
-    switch (tree->error) {
+    switch (owl_tree_get_error(tree, 0)) {
     case ERROR_INVALID_FILE:
         fprintf(stderr, "invalid file\n");
         break;
@@ -1807,6 +1839,8 @@ static size_t read_whitespace(const char *text, void *info);
 static size_t read_keyword_token(uint32_t *token, bool *end_token, const char *text, void *info);
 static void write_identifier_token(size_t offset, size_t length, void *info) {
     struct owl_tree *tree = info;
+    if (tree->error != ERROR_NONE)
+        return;
     size_t token_offset = tree->next_offset;
     write_tree(tree, token_offset - tree->next_identifier_token_offset);
     write_tree(tree, offset);
@@ -1816,6 +1850,8 @@ static void write_identifier_token(size_t offset, size_t length, void *info) {
 static void write_string_token(size_t offset, size_t length, const char *string, size_t string_length, bool has_escapes, void *info) {
     struct owl_tree *tree = info;
     size_t string_offset = has_escapes ? (uint8_t *)string - tree->parse_tree : 0;
+    if (tree->error != ERROR_NONE)
+        return;
     size_t token_offset = tree->next_offset;
     write_tree(tree, token_offset - tree->next_string_token_offset);
     write_tree(tree, offset);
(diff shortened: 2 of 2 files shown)


evnchn:
Flashed this branch to a real ESP32 and threw 50+ parses at it (100-term expressions, 30-deep nesting, 1000-char strings) — no crash, no abort, no reboot. [※](https://zauberzeug.github.io/colophon/#m=qwen-code&t=Hardware%20verification%20by%20operator%2C%20wording%20from%20the%20model. "qwen-code: Hardware verification by operator, wording from the model.")

JensOgorek (approved):
### Review passed, and the fix is now proven on hardware: without it the bench S3 aborts in `write_tree` and reboot-loops, with it the line is dropped and the board stays up

I reviewed the bump and then ran the two builds side by side on the two bench ESP32-S3 boards — `main` (`v0.13.0-18-g01cccee`) on one, this branch (`v0.13.0-18-ga37a5f6`) on the other. On a startup script big enough to exhaust the heap during the parse, `main` panics with `abort() was called` in `write_tree` (`parser.h:376`, exactly the site named in #240) and goes into a reboot loop, while this branch answers `error: allocation failure while parsing` and reaches `Ready.` normally. For everything that is not an allocation failure I could not find any behaviour change at all. Good to merge from my side.

One observation, and it is not about this PR's correctness: `owl_tree_create_with_error()` is the last allocation site in the generated file that upstream left unguarded. Not reachable from Lizard, but worth a note to ianh if you talk to him again — details below.

<details>
<summary>The regenerated <code>parser.h</code> is exactly what <code>gen_parser.sh</code> produces at owl <code>1b8b7f4</code></summary>

Regenerating on my machine gives a file that differs from the committed one in the state-machine section (`state_func_*` numbering), so a plain diff is not conclusive. But the same discrepancy exists for the *pre-PR* file against the *pre-PR* owl, so it is environmental, not staleness. The decisive check: regenerate `main` at `ae33e64` and this branch at `1b8b7f4` in the same environment, then compare the two regenerated deltas against the committed delta — **the change sets are identical line for line**. Both files have `state_funcs[493]`, and the generator is deterministic within one environment (three runs, identical md5).

</details>

<details>
<summary>Behaviour for normal input is unchanged — 9723 lines, byte-identical</summary>

Host harness compiled twice, once against `main`'s `parser.h` and once against this branch's, parsing the same corpus line by line and dumping the error code, the error range and the full `owl_tree_print()` tree for every successful parse:

| | |
|---|---|
| corpus | all `examples/*.liz`, every fenced block in `docs/`, plus generated statements covering every grammar production, operator and precedence level, escaped strings, 800-deep nesting, 800-term flat expressions, and 3000 mutated (invalid) variants |
| lines | 9723 — 6143 `ERROR_NONE`, 1016 `ERROR_INVALID_TOKEN`, 2493 `ERROR_UNEXPECTED_TOKEN`, 71 `ERROR_MORE_INPUT_NEEDED` |
| result | output **byte-identical** (`md5 bb1ec6f2…` for both) |
| sanitizers | ASan + UBSan + leak detection, clean on both sides |

The same holds on the device: with an identical startup script exercising every operator, both boards print the same values, the same escaped string and the same `checksum: b053`; on a syntax error both print `error: unexpected token at range 27 28 "…"`, on a bad character both print `error: invalid token at range 23 24 "~"`.

</details>

<details>
<summary>On-device allocation failure: <code>main</code> aborts, this branch drops the line</summary>

Both boards got the same ~14.8 KB startup script (347 × `int v<i> = 1 + 2 * 3 - 4 / 5 % 6 // 7 ** 8`), injected as an NVS image at `0x9000`, so that `process_lizard()` parses it in one go at boot and runs out of heap while building the tree.

`main`:

```
abort() was called at PC 0x42045b31 on core 0
Backtrace: 0x40375aee 0x403812f9 0x403863e9 0x42045b31 0x420462a6 0x42046bb4 …

write_tree                  parser.h:376
finish_node                 parser.h:1227
construct_expression_reduce parser.h:2312
construct_action_apply      parser.h:2411
apply_actions               parser.h:7870
build_parse_tree            parser.h:7905
parse_string                parser.h:5244
owl_tree_create_from_string parser.h:5252
process_lizard              main.cpp:328
```

Nine resets in ten seconds — the boot loop from #240.

This branch, same script, same board type:

```
error: allocation failure while parsing
Ready.
```

</details>

<details>
<summary>Independent allocation-failure sweep on the host</summary>

I did not reuse the PR's harness but wrote my own: `malloc`/`calloc`/`realloc`/`free` interposed by macro, ASan + UBSan with `-fno-sanitize-recover`, live-block accounting, each input run once per allocation with that allocation failing and once with every allocation from that point on failing. Inputs covered valid statements, constructors, `when/then/end`, `let/do/end`, escaped strings, deeply nested parentheses (which exercises the `bracket_entry_state` stack realloc), a flat expression with more than a thousand allocations, and invalid lines.

16 inputs, 4070 failure cases: every one ends in `ERROR_ALLOCATION_FAILURE` or the input's own baseline error — no crash, no leak, no aborted process. The same harness built against `main`'s `parser.h` crashes in 4038 of the 4070 cases and reproduces #240's numbers exactly: of the 27 allocations of `x = 1 + 2`, 26 crash before the bump and none after it.

I also checked the remaining `abort()` calls by hand: 16 before, 9 after, and the nine are genuine invariant checks (token-list exhaustion, capacity overflow, action-table miss, empty state stack, malformed length encoding). The token-list ones cannot be reached through a skipped `write_tree`, because `parse_string()` now bails on `tree->error != ERROR_NONE` before `build_parse_tree()` ever runs.

</details>

<details>
<summary>Builds — CI only runs pre-commit, so I built both targets locally</summary>

ESP-IDF v5.3.1, both targets build clean. `esp32s3` gets slightly smaller and does not cost any static RAM:

| section | main | this branch | Δ |
|---|---|---|---|
| `.flash.text` | 740630 | 739946 | −684 |
| `.flash.rodata` | 445236 | 444908 | −328 |
| `.dram0.data` | 18436 | 18436 | 0 |
| `.dram0.bss` | 29840 | 29832 | −8 |

</details>

<details>
<summary>The one leftover: <code>owl_tree_create_with_error()</code>, plus a cosmetic upstream nit</summary>

`owl_tree_create_with_error()` still writes through an unchecked `owl_tree_create_empty()`:

```c
static struct owl_tree *owl_tree_create_with_error(enum owl_error e) {
    struct owl_tree *tree = owl_tree_create_empty();
    tree->error = e;                 // NULL deref under OOM
    return tree;
}
```

It is the only allocation site in the file that `1b8b7f4` left unguarded, and it contradicts the invariant the rest of the change establishes. Not reachable from Lizard: `process_lizard()` always passes a non-NULL line, so neither the `INVALID_FILE` nor the `INVALID_OPTIONS` branch can fire. So this is not a blocker for the bump — just something for ianh.

Second, cosmetic: the new `break` when `allocate_string_contents()` fails runs *after* `encode_token_length()` has already appended the token's length pair, so the published run carries one length pair more than it has tokens. Every consumer is short-circuited by `tokenizer->allocation_failed`, so it cannot misbehave today, but the other `break`s in that loop all fire before the encoding — this one is the odd one out.

</details>

Reviewed with Claude Opus 5, hardware-verified on the bench ESP32-S3 pair (not on a robot).

PR #286, opened 2026-09-01
Allocate UART interrupts on core 1, away from the Bluetooth controller

JensOgorek:
### Motivation

With `Bluetooth(...)` active on a classic ESP32 core, a `SerialBus` at 460800 loses frames: on F31 (core as coordinator, plexus ESP32-S3 as peer over half-duplex RS485, peer pushing three values at 100 Hz) the coordinator logged 71 checksum mismatches and 30 poll timeouts in 64 s, plus `dropped 1 inbound messages` reports. The frames themselves carried valid checksums; they arrived merged (lost `\n`) or truncated, i.e. the UART RX FIFO overflowed on the core. With the same firmware and traffic but a minimal startup script (no `Bluetooth`, no `Can`) the bus ran clean for the whole session. Everything that competes for interrupt time in that configuration lives on core 0: the sdkconfig pins the Bluetooth controller there with high-level interrupts (`CONFIG_BTDM_CTRL_PINNED_TO_CORE_0`, `CONFIG_BTDM_CTRL_HLI`), the CAN driver is installed from the main task on core 0, and so is every UART driver, because `Serial::initialize_uart()` also runs on the main task. Whenever the UART interrupt is held off for longer than the 128-byte FIFO can buffer at 460800 (2.8 ms), bytes are lost. A bench check on a lone ESP32 with only `Bluetooth(...)` active and UART0 flooded at 85 % of 460800 showed no loss at all (details below), so Bluetooth advertising alone is not enough to trigger it; the robot's combined core-0 load is. #274 does not touch this; it only moves BLE line parsing between tasks.

Found while running the #277 A/B on F31 (2026-09-01); it applies independently of #252/#277 to every `Serial`-based bus next to Bluetooth.

### Implementation

`Serial::initialize_uart()` calls `uart_driver_install` through `esp_ipc_call_blocking(1, ...)`, so the UART interrupt is allocated on core 1 — the core the `SerialBus` communication task is pinned to anyway. `CONFIG_ESP_IPC_TASK_STACK_SIZE` is raised from the 1024-byte default to 2048 in both `sdkconfig.defaults`, because `uart_driver_install` allocates the ring buffers and the interrupt from the IPC task's stack. The console UART0 (`main.cpp`) and the BNO055 driver still install on core 0; UART0 showed no loss on the bench with Bluetooth active (see below), so it is left alone here.

<details>
<summary><b>Measurement on F31 (core with the production startup partially running, i.e. Bluetooth active; runtime bus setup <code>Serial(13, 4, 460800, 2)</code>, <code>SerialBus</code>, <code>make_coordinator(2)</code>; plexus pushing <code>wm</code>/<code>ws</code>/<code>wd</code> at 100 Hz)</b></summary>

| core firmware | checksum mismatches | poll timeouts | duration |
|---|---|---|---|
| main (incl. #274) | 71 | 30, spread over the whole run | 64 s |
| same + UART interrupt on core 1 (#277 branch with main merged) | 0 | 9, all within the first 2.5 s (peer reboot caused by opening its USB console) | 64 s |
| same, confirmation | 0 | 10, all within the first 2.5 s (same cause) | 312 s |
| this branch (`9aef0ab`), both nodes | 1, in the peer's boot frame (`__ECHO__Ready.` with a glitch byte from the S3 reset) | 8, all within the first 1.8 s (same cause) | 124 s |

After the peer was up, no warning on either console in any of the fix runs; 12 365 output lines at 100 Hz in the last run. Binary size unchanged within noise (esp32 `0x13c780`, esp32s3 `0x13a9d0`). The robot was restored afterwards (core on its production build with startup `4fb4`, plexus on its previous build and script).

</details>

<details>
<summary><b>Bench check of UART0 (lone ESP32 on a CP2102N, host flooding the console at 460800 with 40-byte checksummed no-op lines, 85 % line utilisation, 60 000 lines per run; test-only counter of the driver's <code>UART_FIFO_OVF</code>/<code>UART_BUFFER_FULL</code> events)</b></summary>

| UART0 interrupt on | Bluetooth | device checksum mismatches | FIFO overflows |
|---|---|---|---|
| core 0 (this branch) | off | 0 | 0 |
| core 0 (this branch) | on | 0 | 0 |
| core 1 (experimental) | on | 0 | 0 |
| core 1 (experimental) | off | 0 | 0 |

No difference in either direction, so UART0 is left where it is.

</details>

### Progress


JensOgorek:
Superseded by #287, which cherry-picks this commit and adds the same treatment for the console UART plus the buffer and FIFO changes; closing in favour of #287.

PR #287, opened 2026-09-07
Accept long console lines and keep UART0 receive intact under bursts at 921600 baud

### LINKED ISSUE #270: UART0 REPL: a line over 1 KB overruns process_uart()'s buffer and panics the device

**A single UART0 line longer than 1024 bytes writes past `process_uart()`'s buffer and panics the device.** `input` is `static char[1024]` (`main/main.cpp:414`) but the read is `uart_read_bytes(UART_NUM_0, (uint8_t *)input, pos + 1, 0)` (line 420) with no size guard, while the UART0 ring is installed at `BUFFER_SIZE * 2` = 2048 (line 462). So `pos` can be up to ~2047 and up to 1024 bytes land past the buffer.

Reproduced on an ESP32 devkit: **1279 'A's survive, 1280 'A's panic** with `LoadProhibited`. The boundary matches the link map exactly — `core_module` sits at `input + 1280`, and the crash begins on the first write that reaches it.

This is pre-existing, not a regression. It surfaced in the Fable review on [pull request 262](https://github.com/zauberzeug/lizard/pull/262#pullrequestreview-4928537151), which correctly flagged it as the same bug class that PR fixes for the expander, duplicated inline here without the guard.

<details>
<summary><b>MRE — paste and run against any lizard board</b></summary>

```python
#!/usr/bin/env python3
"""Lizard UART0 REPL: one oversized line overruns process_uart()'s 1 KB buffer."""
import sys
import time

import serial

s = serial.Serial(sys.argv[1] if len(sys.argv) > 1 else '/dev/ttyUSB0', 115200, timeout=0.5)


def send(n, label):
    s.reset_input_buffer()
    s.write(b'A' * n + b'\n')
    s.flush()
    time.sleep(2.0)
    out = s.read(30000).decode('utf-8', 'replace')
    crashed = 'Guru Meditation' in out
    print(f'[{label}] {n} chars -> {"CRASH" if crashed else "survived"}')
    return crashed


assert not send(1279, 'control'), 'control should not crash'
assert send(1280, 'oversized'), 'oversized should crash'
print('reproduced')
```

Nothing about the payload matters — plain `A`s, no valid command, no checksum. The overwrite happens in `uart_read_bytes` before anything is parsed.
</details>

<details>
<summary><b>Observed output</b></summary>

```
n=1278 (writes 1279 bytes) -> survived
n=1279 (writes 1280 bytes) -> survived
n=1280 (writes 1281 bytes) -> CRASH
n=1281 (writes 1282 bytes) -> CRASH
```

```
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.

Core  0 register dump:
PC      : 0x40150d53  PS      : 0x00060830  A0      : 0x800fb612  A1      : 0x3ffd32d0
A2      : 0x9e0a2fbe  A3      : 0x3ffd33ec  A4      : 0x00000005  A5      : 0x3ffd33f8
EXCCAUSE: 0x0000001c  EXCVADDR: 0x9e0a2fc2
```

A shorter overshoot is quieter and easy to mistake for a handled error. A 1114-byte line only reaches `process_tree`'s scratch buffer and surfaces as `error processing uart0: buffer too small` — thrown *downstream* by `csprintf` (`main/utils/string_utils.cpp:27`) and caught at `main/main.cpp:493`. The device stays up, so the out-of-bounds write is invisible unless you go looking for it.
</details>

<details>
<summary><b>Why 1280 — the link map</b></summary>

From `build/lizard.map`:

```
0x3ffc37f4  0x400  .bss._ZZ12process_uartvE5input          <- static char[1024]
0x3ffc3bf4  0x100  .bss._ZZ12process_treeP8owl_treebE6buffer
0x3ffc3cf4  0x008  .bss.core_module                        <- std::shared_ptr<Module>
```

`core_module` is at `input + 1280`. Writing `n + 1` bytes (the line plus its `\n`) first touches it at `n = 1280`, which is exactly where the observed boundary flips. Bytes 1024–1279 land in `process_tree`'s scratch buffer, which is rewritten before use — that is why a merely-somewhat-oversized line looks harmless.

The absolute addresses are build-specific, but the adjacency follows declaration order within `main.cpp`, so the shape should hold for any build of this file.
</details>

<details>
<summary><b>Test setup, and what it does not cover</b></summary>

ESP32 devkit (chip rev v3.1, 4 MB flash) on ESP-IDF v5.3.1, firmware built from `main` at 0fbae94 plus the PR-262 branch — neither touches `process_uart`.

Two deviations from a stock build, both needed only to fit a 4 MB devkit and neither affecting this code path: flashed with `--flash-size 4MB` (a stock build's image header says 8 MB and bootloops on a 4 MB board with `Detected size(4096k) smaller than the size in the binary image header(8192k)`), and a partition table shrunk to fit 4 MB (`coredump` in `partitions.csv` sits at `0x400000`, past the end of a 4 MB flash).

Not covered: ESP32-S3, and whether any realistic host ever sends a line this long. On a normal RoSys setup UART0 is the trusted host link, so I'd read this as robustness rather than an attack surface — but it needs no degraded driver state and no unusual configuration to fire.
</details>

### Suggested fix

The same guard `Serial::read_line` already has:

```cpp
if (pos >= static_cast<int>(sizeof(input))) {
    // discard the line, report it, continue
}
```

Happy to open a PR if you'd like it in the same shape as the expander fix, or to leave it to you if you'd rather handle the REPL path differently.

JensOgorek:
### Motivation

Configuring a robot brain with many modules over the console at 921600 baud silently loses lines: the four-arm F30 rig's startup script (142 lines, one `core.output` line of 993 characters) never arrived complete, and the truncated script left the core without any output. Three independent limits in the UART0 path were responsible: the 1024-byte line buffer (read without a length check, so a longer line overran it), the receive FIFO threshold (~70 µs of interrupt slack at 921600), and the UART0 interrupt sharing core 0 with the Bluetooth controller, which delays it long enough to drop whole chunks of a burst.

### Implementation

- The UART0 line buffer grows to `CONSOLE_LINE_SIZE` (2048 bytes) and `process_uart` discards (and reports) a line that still does not fit instead of reading past the buffer; an unterminated run longer than a line is discarded up to its line end. The driver's receive ring is 8 KB and the pattern queue holds 512 line ends, so a full startup script sent in one burst is buffered (a 16 KB ring was tried first and cost too much heap on a Bluetooth core).
- A receive FIFO threshold of 32 bytes fires the interrupt well before the 128-byte hardware FIFO fills, for UART0 and the `Serial` modules alike.
- All UART drivers are installed through one helper, `install_uart_driver_on_core1` (`esp_ipc_call_blocking(1, …)`), so their interrupts are allocated on core 1, away from the Bluetooth controller; the IPC task stack is 2048 bytes on the ESP32 and 4096 on the ESP32-S3 (the S3 default of 1280 boot-loops).
- `echo()` and `Core::get_output` share `CONSOLE_LINE_SIZE` (2048) with the receive path; an output line that does not fit becomes `warning: console line of N bytes exceeds 2048 bytes and was dropped` instead of a truncated line with a valid checksum (Evan's finding below), and an overlong core line reports one error per episode and is suppressed until it fits again.
- The serial-bus UART gets the same treatment: the commit from `uart-isr-core1` (#286) is cherry-picked here, so this PR supersedes #286. On the F30 core (Bluetooth on, four peers at 460800) that turned 4 poll timeouts per peer, bus checksum mismatches and `could not enqueue outgoing message` errors within 30 s into zero, and capped the main loop at 55 ms instead of 338 ms (every lost poll stalls it by the 250 ms timeout).
- `process_uart` runs every line of a read on its own: the IDF driver queues only the last line end of each receive chunk, so under load two close line ends come back as one read, and both lines used to fail the checksum and get dropped.

Measured on the F30 core (ESP32, Bluetooth on, four plexus peers on the bus) by replaying rosys' exact 142-line configure at 921600 with checksums and no pacing: before the core-1 change 3 of 4 runs lost lines (checksum mismatches, up to five consecutive lines), after it 5 of 5 runs stored every line. A second series on the same core at 460800 with its current 106-line script still lost a pair of adjacent lines in 3–10 of 60 bursts, which the per-line split removes (60 of 60 bursts, 10 of 10 full configures). Free heap with the current F30 configuration: 30.5 KB.

<details>
<summary>How the loss was pinned down</summary>

- rosys' configure (`!-`, 142× `!+line@xx`, `!.`, `core.restart()`) at 921600 stored 124–128 of 128 non-empty lines; the missing ones changed from run to run, the 993-character `core.output` line was always among them, and the core printed `warning: Checksum mismatch while processing UART0` — so bytes were lost on the way in, not rejected by length.
- At 460800 the same burst was clean only with 5 ms between lines; at 921600 not even then.
- With the 2048-byte buffer alone the long line was stored truncated at 993 characters (no error); before that the read overran the 1024-byte array.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Implementation%20and%20description%20drafted%20by%20the%20model%2C%20directed%20and%20hardware-tested%20by%20the%20author.&a=claude-code "claude-fable-5-1: Implementation and description drafted by the model, directed and hardware-tested by the author.")

### Progress

- [x] I chose meaningful labels (if GitHub allows me to so).
- [x] Tests with a real hardware have been successful (or are not necessary).
- [x] Documentation has been added (or is not necessary).

Fixes #270.


evnchn:
**The TX side still caps at 1024, and its truncation is checksum-valid.**

RX and `echo`'s output buffer were both 1024 before this PR, so the two directions matched. Now RX is 2048 and [`echo` still holds `static char buffer[1024]`](https://github.com/zauberzeug/lizard/blob/eb49587bcab85125ceced7908638f7c5145290fe/main/utils/uart.cpp#L15-L21), which the `sizeof(buffer) - 1` / `- 2` clamps cap at 1022 characters. The checksum is computed over the truncated text, so it validates: rosys gets a well-formed line with the tail fields missing, and `handle_core_output` pops positionally.

The core line takes that path ([`module.cpp:59`](https://github.com/zauberzeug/lizard/blob/eb49587bcab85125ceced7908638f7c5145290fe/main/modules/module.cpp#L59)). F30's is 453 B, so this is headroom rather than a live bug — but the four-arm rig's own follow-up list grows each plexus to 16 mirrored fields, and 1022 fails more quietly than the mismatch this PR fixes. Raise `echo`'s buffer alongside `BUFFER_SIZE`, or have it warn on truncation the way the new RX drop path does?

evnchn:
**Ran it: 983 characters lost, `checksum_ok` true in both cases.** `main/utils/uart.cpp` has no esp/freertos includes, so this builds and runs on a host in about a second — no ESP-IDF, no rig.

```
sent  900 chars -> body  905 chars, kept  900, lost    0, checksum_ok=TRUE
sent 2000 chars -> body 1022 chars, kept 1017, lost  983, checksum_ok=TRUE
```

`checksum_ok` is the repo's own `check()` from that same file, so the receiver-side verifier accepts the truncated line — the two rows are indistinguishable to whoever reads them. Clean under `-fsanitize=address,undefined`: this is silent loss, not the buffer overrun the new RX guard fixes.

It calls `echo()` directly rather than going through `core.output` on hardware, but with the same arguments `main/modules/module.cpp:59` passes.

<details>
<summary><code>repro.cpp</code> at <code>eb49587</code> — <code>c++ -std=c++17 -I main repro.cpp main/utils/uart.cpp -o repro && ./repro</code></summary>

```cpp
// echo() truncates a long line and still emits a checksum that validates.
// build: c++ -std=c++17 -I main repro.cpp main/utils/uart.cpp -o repro && ./repro
// stdout is the captured Lizard line; the report goes to stderr.
#include "utils/uart.h"
#include <cassert>
#include <cstdio>
#include <string>
#include <vector>

static std::string emit(const std::string &output) {
    FILE *file = freopen("echo_line.txt", "w", stdout);
    echo("%s %s", "core", output.c_str()); // exactly what modules/module.cpp:59 sends
    fflush(file);
    fclose(file);
    file = fopen("echo_line.txt", "r");
    char line[8192] = {0};
    (void)fgets(line, sizeof(line), file);
    fclose(file);
    std::string text(line);
    while (!text.empty() && (text.back() == '\n' || text.back() == '\r')) {
        text.pop_back();
    }
    return text;
}

int main() {
    for (const int payload : {900, 2000}) {
        const std::string sent = std::string(payload, 'x');
        const std::string line = emit(sent);
        std::vector<char> buffer(line.begin(), line.end());
        buffer.push_back('\0');
        bool checksum_ok = false;
        const int body = check(buffer.data(), static_cast<int>(line.size()), &checksum_ok);
        const int kept = body - 5; // minus "core "
        fprintf(stderr, "sent %4d chars -> body %4d chars, kept %4d, lost %4d, checksum_ok=%s\n",
                payload, body, kept, payload - kept, checksum_ok ? "TRUE" : "false");
        assert(checksum_ok); // the receiver cannot tell the two cases apart
    }
    return 0;
}
```

</details>

JensOgorek:
### Fixed: `echo()` shares the 2048-byte line size and drops an overlong line with a warning; verified on a bench ESP32-S3

Thanks, the repro was exactly right. `echo()` now formats into `CONSOLE_LINE_SIZE` (2048, shared with the receive path), and a line that still does not fit is replaced by `warning: console line of N bytes exceeds 2046 bytes and was dropped`, the same way `process_uart` reports an overlong input line. `Core::get_output` had its own 1024-byte buffer that threw `buffer too small` on every tick long before the echo limit, so it shares the constant now as well. Two more commits ride along: the ESP32-S3 IPC task needs a 4 KB stack for the `esp_ipc_call_blocking` driver install, otherwise every S3 boot-loops with this branch (found on the bench today), and the buffer constant lives in `uart.h`.

<details>
<summary>Host repro against the new <code>uart.cpp</code> (your program, more lengths)</summary>

```
sent  900 chars -> line  905 chars, checksum_ok=true, intact
sent 1500 chars -> line 1505 chars, checksum_ok=true, intact
sent 2041 chars -> line 2046 chars, checksum_ok=true, intact
sent 2042 chars -> line   70 chars, checksum_ok=true, warning: console line of 2047 bytes exceeds 2046 bytes and was dropped
sent 2500 chars -> line   70 chars, checksum_ok=true, warning: console line of 2505 bytes exceeds 2046 bytes and was dropped
```

</details>

<details>
<summary>Bench ESP32-S3 through the real <code>core.output</code> path</summary>

A `str` of 40 characters repeated 47 times in the output format plus one more string of chosen length gives core lines of exact sizes (115200 baud, 4efee55):

| console line | result |
|---|---|
| 2046 chars | intact, checksum valid |
| 2047 chars | `warning: console line of 2047 bytes exceeds 2046 bytes and was dropped` every tick |
| 2049 chars | same warning with 2049 |
| output 2048 chars and more | `error in module "core": buffer too small` every tick (the output buffer, before echo) |

</details>

Tested with Claude Fable 5.1.

evnchn:
**Verified on `4efee55` — holds.** Swept every `core.output` length 1..2100 rather than five samples: 2041 intact, 59 loud, **0 silent**, clean under `-fsanitize=address,undefined`. The same harness still finds 6 silent truncations on `main`, so it can see the failure it reports absent here.

falkoschindler (commented):
The hardware result is convincing and the three fixes (bounded read, lower FIFO threshold, ISR on core 1) are the right ones. The PR's stated invariant, one `CONSOLE_LINE_SIZE` "in either direction", does not hold yet on two paths, and one residual way to lose UART0 for good is still there. None of these are regressions against main; they are gaps between the description and the code. Details below, most important first.

## Correctness

**1. The expander receive path still stops at 1024 bytes.** `Expander::check_boot_progress` and `Expander::handle_messages` read into `static char buffer[1024]` (main/modules/expander.cpp (81) and (136)) via `Serial::read_line`, which rejects any line with `pos >= 1024`. With this PR an expander running the same firmware can emit a `core.output` line of up to 2050 wire bytes. The parent then prints `<expander>: error while handling messages: buffer too small. discarded line.` every 10 ms for anything between 1025 and 2048 bytes, and for 2049 to 2050 bytes `Serial::read_line` takes the `LINE_FLUSHED` branch, which calls `uart_flush_input` and drops every other buffered line with it. `RX_BUF_SIZE` in main/modules/serial.cpp (9) is 2048, so a maximal line cannot even be buffered by a Serial module. Before the PR the child threw "buffer too small" in `Core::get_output` above 1023 chars, so the failure has moved from the child to the parent rather than gone away. The 993-character line from the F30 rig would be fine over an expander link only because it is below 1020. Suggest sizing both expander buffers with `CONSOLE_LINE_SIZE` (expander.cpp already includes utils/uart.h via echo) and making `RX_BUF_SIZE` at least `CONSOLE_LINE_SIZE + 4`, or documenting that the expander path keeps the old limit. The send-side 1024 buffers for `!!` broadcasts (main/modules/module.cpp (63)) and proxy calls (expander.cpp (218)) throw above 1023 as before, so they are consistent but do not follow the constant either.

**2. An unterminated run longer than the ring still kills UART0 until reboot.** `process_uart` only reads when `uart_pattern_pop_pos` returns a position (main/main.cpp (416) to (419)). If 8 KB arrive without a `\n` (a pasted blob, a script with stripped newlines, a host bug), the ISR sets `rx_buffer_full_flg` and disables the RX interrupts; every `\n` that arrives afterwards is not enqueued (IDF uart.c, the `else` branch at 1182 to 1194 only sets a local flag), the hardware FIFO overflows and resets, and nothing ever calls `uart_read_bytes` or `uart_flush_input` to re-enable RX. The 20-entry event queue that would carry `UART_BUFFER_FULL` is never read. This is pre-existing at 2 KB, but the PR body says over-long lines are now discarded and reported, and the new drop branch only covers lines whose `\n` was detected while the ring had room. A stalled main loop with normal newline-terminated traffic is not affected: positions from before the stall are still queued, and the first `uart_read_bytes` replays the stash and re-enables RX, so that case costs a few checksum mismatches, not deafness. Three lines close the hole:

```cpp
if (pos < 0) {
    size_t buffered = 0;
    uart_get_buffered_data_len(UART_NUM_0, &buffered);
    if (buffered > BUFFER_SIZE) {
        uart_flush_input(UART_NUM_0);
        echo("warning: UART0 line longer than %d bytes discarded", BUFFER_SIZE);
    }
    break;
}
```

An unterminated run that is already longer than the line buffer can never be processed, so flushing early loses nothing usable, and `uart_flush_input` also clears the full flag, resets the FIFO and re-enables the interrupts if the check ever runs late.

**3. `Core::get_output` has a dead zone of six lengths where the line is dropped every tick.** `Module::step` (main/modules/module.cpp (59)) emits `echo("%s %s", name, output)`, so the wire payload is `"core "` plus the rendered output. `csprintf` in main/modules/core.cpp (224) accepts up to 2047 chars, `echo` drops payloads over 2046. An output of 2042 to 2047 chars therefore passes `get_output` and is replaced by `warning: console line of N bytes exceeds 2046 bytes and was dropped` 100 times a second; at 2048 and above `csprintf` throws and `run_step` prints `error in module "core": buffer too small` at the same rate. The effective limit is 2041, not 2048, and neither message names a field to drop. Cheapest fix: size `output_buffer` as `CONSOLE_LINE_SIZE - 6` with a comment pointing at the `Module::step` format. Nicer: detect the overflow once, echo one error with the element count, and set `output_on = false` so the storm stops.

**4. Output lines can be two bytes longer than the input side accepts.** `echo` caps the formatted payload at 2046 and then `printf("%s@%02x\n", ...)` appends the checksum and newline outside the buffer (main/utils/uart.cpp (37)), so the wire line reaches 2050 bytes. `process_uart` drops anything with `pos + 1 > 2048`. The comment on `CONSOLE_LINE_SIZE` in main/utils/uart.h (7) says "in either direction, including its terminator", which is off by the `@xx`. Exposure is small (echo output goes to the host or to a parent's Serial, which has its own smaller cap, see 1), but the guard should be `pos > CONSOLE_LINE_SIZE - 4` so payload plus `@xx\n` fits, and the warning should print that number.

**5. The IPC wrappers throw away every `esp_err_t`.** `install_console_uart` (main.cpp (409)) and `install_uart_driver` (serial.cpp (43)) are `void(void *)` callbacks, so the result of `uart_driver_install` is gone; the results of `esp_ipc_call_blocking` and `uart_set_rx_full_threshold` are dropped as well. Main never checked these either, but it could have wrapped the direct call in `ESP_ERROR_CHECK` (as can.cpp (71) does for TWAI); now the value has to be threaded out through the argument. If the install fails the ESP boots, prints "Ready." and is deaf on UART0, with `uart: uart driver error` from `uart_pattern_pop_pos` every tick as the only hint; a Serial module in that state reads nothing and reports "connection lost". Unlikely at boot (the install runs at near-maximal free heap) but reachable on a Serial reinstall after an expander flash at 38 KB free. A small `struct { uart_port_t port; esp_err_t err; }` as the IPC argument fixes it, removes the `const_cast` in serial.cpp (61) at the same time, and lets `Serial::initialize_uart` throw and `app_main` use `ESP_ERROR_CHECK`. While there: `uart_queue` in main.cpp (473) is never read after the install. Passing `NULL` and `0` drops the unused 20-entry queue and the `ESP_LOGI("queue free spaces")` that IDF runs on the IPC task stack for a non-NULL queue (uart.c (1732)), which is the one `vprintf` plus VFS call on that stack.

## Design questions

**6. Were the 8 KB ring, the 512-entry pattern queue and the threshold of 32 measured against the core move alone?** Reading the PR body, all three were already in place when 3 of 4 runs still lost lines, and the core-1 move is the only change with its own before/after measurement. The ring alone costs 6 KB of the 38 KB free heap, the pattern queue 1.6 KB more, and neither is shown to be needed once the ISR no longer competes with Bluetooth. The arithmetic for the ring is plausible (a 55 ms main-loop stall at 921600 baud is about 5 KB), so this is not an objection to the values, only a request for the measurement: one run on the F30 rig with the old 2 KB ring, 100-entry queue and default threshold but the core-1 move in place would tell whether the heap is well spent. That run fits into the A/B proposed in 7 as variant c.

**7. Proposal: replace the IPC installs with `CONFIG_ESP_MAIN_TASK_AFFINITY_CPU1=y`, verified by one A/B run on the F30 rig.** `esp_intr_alloc` always binds the ISR to the calling core (IDF intr_alloc.c (544)); there is no core flag, which is why the PR has to run the installs on the core-1 IPC task. Every `uart_driver_install` in Lizard runs on the main task (app_main, `process_line` from UART0, Bluetooth and SerialBus all step on the main task), so pinning the main task to core 1 puts all of them there with one sdkconfig line and no code: no trampolines, no `esp_ipc.h`, no `const_cast`, no lost `esp_err_t` (5), and no IPC stack bump. The bump is not free: `CONFIG_ESP_IPC_TASK_STACK_SIZE` applies to both IPC tasks, so the S3 pays 2 x (4096 - 1280) = 5.6 KB and the ESP32 2 x (2048 - 1024) = 2 KB of heap permanently, half of it on `ipc0` for nothing. It is also the usual IDF layout, protocol stacks on core 0 and the application on core 1, and it works forward: any driver installed from the main loop in the future lands away from the Bluetooth controller without the author having to know about IPC. Two things change beyond UART and are the reason for the bench run rather than a reason against: every other ISR installed from the main task (I2C in utils/i2c_bus.cpp (36), TWAI in can.cpp (71) and (271), PCNT in stepper_motor.cpp (74), the GPIO ISR service) moves to core 1 as well, arguably what we want but untested; and the main loop itself moves, so it is no longer preempted by the BT controller, NimBLE host, esp_timer and scheduler task on core 0 but only by the bus task on core 1, which mostly waits on UART. That may lower the 55 ms loop maximum further; a guess, not a claim. One edge stays: `expander.flash()` inside an `at`/`after` block runs on the scheduler task on core 0, so that reinstall would land on core 0 unless scheduler.cpp (83) pins the scheduler task to core 1 too, which seems reasonable anyway. Suggested A/B on the F30 rig, same protocol as in the description: 5 configure replays at 921600, 30 s bus error count with four peers, and the main-loop maximum, for (a) this branch as is, (b) affinity switch with the IPC code removed, and, to settle 6 in the same session, (c) variant b with the old 2 KB ring, 100-entry pattern queue and default threshold. If b matches or beats a, the IPC code goes; if c matches b, the ring and threshold go too and about 10 KB of heap come back. CAN is the path I would watch most closely in b and c. If the IPC approach stays after all, one shared helper in utils/uart (`uart_driver_install_on_core1(port, rx, tx, queue_len, queue) -> esp_err_t`) should replace both trampolines and carry the error back (see 5).

## Cleanup

- **`BUFFER_SIZE` is now a pure alias for `CONSOLE_LINE_SIZE`** (main.cpp (40)) with a second comment restating the first. Use `CONSOLE_LINE_SIZE` at the four sites in `process_uart` and delete the macro. The drop loop can use `std::min(remaining, CONSOLE_LINE_SIZE)`; the `read <= 0` break is unreachable for UART0 (the driver commits ring bytes before publishing the pattern position and replays the stash inside `uart_read_bytes`), so it is fine to keep as a guard but does not need the comment-free early exit.
- **Three phrasings and three numbers for one event.** `warning: UART0 line longer than 2048 bytes discarded` (counts the `\n`), `warning: console line of N bytes exceeds 2046 bytes and was dropped` (excludes `\n` and `@xx`), and `buffer too small. discarded line.` from `Serial::read_line_error` (no number, limit is the caller's 1024). A host log reader cannot tell whether the limit is 2046 or 2048. Once 4 is fixed the two UART0 numbers can agree; consider `warning: UART0: %s` with `Serial::read_line_error` if the drop logic is shared.
- **The comment above `esp_ipc_call_blocking` in main.cpp (474) repeats the sizing rationale** already attached to `RX_RING_SIZE` and `RX_PATTERN_QUEUE`; only its last line explains the call it sits on. The threshold comment below it is fine.
- **The two sdkconfig values are uncommented and differ without explanation.** A reader tidying `sdkconfig.defaults` cannot tell that `CONFIG_ESP_IPC_TASK_STACK_SIZE` is load-bearing (removing it boot-loops the S3, per 7cce04e) or why the ESP32 gets 2048 and the S3 4096. The ESP32 value did survive the F30 bench runs including restarts and four Serial installs, so it is verified for the boot path; a one-line `# uart_driver_install runs on the core-1 IPC task; IDF default (1024/1280) overflows` above each and, ideally, one `uxTaskGetStackHighWaterMark` reading per target in the PR would settle it. Passing `NULL` for the unused event queue (5) also removes the heaviest call from that stack.
- **Serial-module UARTs keep the default FIFO threshold of 120.** The description attributes the threshold to UART0 only and the bus measurement (zero errors at 460800 with four peers after the core move) shows the bus does not need it, but a user-defined `Serial` at 921600 has the same 87 us window UART0 had. Either apply `uart_set_rx_full_threshold(uart_num, 32)` in `Serial::initialize_uart` or say in the description that it was left deliberately.
- **`#include "esp_ipc.h"` sits above the sorted block in main.cpp (1)**; serial.cpp puts it in alphabetical position. Belongs between "compilation/variable_assignment.h" and "global.h".
- **The limit is user-visible but undocumented.** docs/machine_safety.md (8) still names checksum failure as the only drop condition. The old 1024 was equally undocumented, so ticking "not necessary" is defensible; one sentence next to the checksum paragraph would save the next person a grep.

## Out of scope, for the record

- **Flash writes still mask the UART ISR on both cores.** `CONFIG_UART_ISR_IN_IRAM` is unset, so spi_flash's cache stall disables non-IRAM interrupts on both cores for every NVS commit (`!.`, bus backup) and OTA chunk; at 921600 the 128-byte FIFO overflows after 1.4 ms and neither the threshold nor the core move can help. The loss pattern in #270 was during the `!+` phase, before any flash write, so this is not the root cause of that issue, and enabling the option costs a few KB of IRAM in a build that already pushed FreeRTOS out of IRAM. Worth a line in the docs: the host must not stream more than 128 bytes at UART0 while the ESP writes flash.
- **A 2 KB output line blocks the main loop for 178 ms at the default 115200 baud** (`echo` prints through the console VFS with no TX ring). Fine at 921600 (22 ms), but a `core.output` near the new limit on a 115200 link degrades the loop to about 5 Hz.

## Checked and dropped

- The `read <= 0` break in the drop loop is unreachable for UART0, so no tail is ever left in the ring.
- BNO055ESP32.cpp (808) still installs its UART directly, but the UART constructor has no caller (imu.cpp uses I2C), so nothing lands on core 0 there.
- The 2048 IPC stack on the ESP32 was exercised by the F30 runs; "unverified" does not hold, only "unmeasured" (see the sdkconfig point).
- A threshold of 64 instead of 32 for a lower ISR rate: the cost is a few percent of core 1 for the ~100 ms of a configure burst, and the PR's own metrics show no core-1 stress. Whether 32 is needed at all is question 6, not a sizing quibble.
- Splitting the PR: the serial-bus part (#286) shares the IPC install and the sdkconfig line with the UART0 part, so separating them again would mean two PRs editing the same helper; the output-direction fix is separable but small and already covered by Evan's test over all lengths. Both chips are tested as a whole, and every split piece would need that again. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20and%20wording%20from%20the%20model%2C%20drawn%20from%20a%20multi-agent%20review%20of%20the%20diff%20against%20the%20IDF%205.3.1%20source%3B%20the%20questions%20on%20splitting%2C%20the%20unmeasured%20buffer%20sizes%20and%20the%20affinity%20alternative%20were%20shaped%20in%20conversation%20with%20the%20reviewer.&a=claude-code "claude-fable-5-1: Findings and wording from the model, drawn from a multi-agent review of the diff against the IDF 5.3.1 source; the questions on splitting, the unmeasured buffer sizes and the affinity alternative were shaped in conversation with the reviewer.")

JensOgorek:
### Review rework pushed: findings 1 to 5 and the cleanup are in, and the F30 A/B answers 6 and 7: the IPC install and the 8 KB ring stay, the affinity variant loses lines

Thanks for the thorough pass. Everything under Correctness and Cleanup is addressed in the five new commits and verified on a bench pair of two ESP32s (core plus expander over UART1), details below. Along the way the payload budget turned out to be one byte short of what the code claimed: the console's stdout ends lines with CRLF, so a maximal core line arrived at its parent as 2049 bytes. That is fixed, the budget now counts `@xx\r\n`. The A/B you proposed for 6 and 7 ran on the F30 rig this afternoon (Bluetooth on, four peers): the affinity variant is slower and loses configure lines at 921600, and the old 2 KB ring loses the whole burst and leaves UART0 deaf until a reset, so the IPC install and the buffers stay; a `static_assert` now pins the ring above the line size.

<details>
<summary>Correctness 1 to 5</summary>

**One size, one unit.** `CONSOLE_LINE_SIZE` (2048) is the line on the wire in either direction, line ending included; `CONSOLE_PAYLOAD_SIZE` (2043) is what a line may carry before `@xx\r\n`. Every warning names the wire size, so a host log reader sees 2048 everywhere.

1. **Expander receive path.** Both receive paths in `Expander` share one `CONSOLE_LINE_SIZE` line buffer (one static buffer instead of two 1024-byte ones, so no extra RAM), and a `Serial` module's receive ring is `2 * CONSOLE_LINE_SIZE`, a maximal line plus what arrives while the main loop handles it. That costs 2 KB of heap per `Serial`; if that is too much for the expander robots, `CONSOLE_LINE_SIZE + 128` (line plus FIFO stash) would do. Bench pair: a 1493-character child output (over the old 1024) is forwarded intact, a 2034-character one arrives at the parent as exactly 2048 wire bytes and is forwarded, and the parent reads a 2048-byte child line without `buffer too small`. The 1024-byte send buffers for `!!` broadcasts and proxy calls are unchanged: they throw loudly at 1023 and are not part of the console-line invariant.
2. **Unterminated run.** `process_uart` checks the buffered length when no line end is queued and flushes anything over `CONSOLE_LINE_SIZE`, as proposed, then keeps discarding up to the next line end, so the tail of the run does not prefix the following command (with the plain flush, a 9000-byte run left 296 bytes that made the next command an `unknown module "yyyy…"`). Bench: 9000 bytes without a line end give one warning, the command after the run's line end answers; a 4000-byte line arriving at 115200 gives one warning and the next command answers; a 2048-byte line is still accepted.
3. **Core output dead zone.** `Core::get_output` formats into `CONSOLE_PAYLOAD_SIZE - 5` bytes (the `core ` prefix from `Module::step`), so a core line can reach exactly 2048 wire bytes and nothing between fits and drops is left. An overflow is reported once, `error: the core line with N output fields exceeds 2048 bytes and is suppressed until core.output is set again`, and the line stays off until the next `core.output(...)`, so no storm and no silent `core.mute()` that would need an `unmute` to undo. `csprintf` throws a `BufferTooSmallError` (a `std::runtime_error`, existing catches keep working) to make the overflow distinguishable. Bench: 114 fields rendering to 2038 characters give 2048-byte lines, 2039 characters give one error and silence, a shorter `core.output` resumes, the next overflow is reported again.
4. **Output longer than the input accepts.** `echo` caps the payload at `CONSOLE_PAYLOAD_SIZE`, so payload plus `@xx\r\n` is at most 2048 bytes; the warning reports the wire size the line would have had. Host sweep of `core.output` payloads 1 to 2200 with `-fsanitize=address,undefined`: 2038 intact, 162 loud, 0 silent, 0 over 2048 bytes, 0 checksum failures.
5. **Lost `esp_err_t`.** One helper, `install_uart_driver_on_core1(port, rx, tx)` in `utils/uart_driver`, replaces both trampolines: it carries the install result out of the IPC call, applies the FIFO threshold and returns the first error. `app_main` wraps it in `ESP_ERROR_CHECK`, `Serial::initialize_uart` throws. The unused event queue is gone (`NULL`, 0), which also drops the `ESP_LOGI` from the IPC task stack. `utils/uart.cpp` stays free of esp includes so Evan's host harness keeps working. Bench: `p0.disconnect()` and a new `Serial` on the same UART (driver deleted from core 0, reinstalled on core 1; IDF frees a foreign-core interrupt over IPC itself) brings the child back.

</details>

<details>
<summary>Design questions 6 and 7: the F30 A/B</summary>

Same protocol as in the description, on the F30 core (ESP32, Bluetooth on, four plexus peers at 460800, rosys stopped): `frames_monitor` 30 s at 460800 for the loop period and the bus (sequence gaps per peer, warning lines), then the console at 921600 and five replays of rosys' 142-line configure (993-character `core.output` line, checksums, no pacing) into the RAM copy, verified with `core.startup_checksum()`. Firmware is lizard-zz with this branch merged into the frames branch the rig runs (#290), because the peers send frames.

| Variant | free heap | loop median / p90 / max | bus 30 s: seq gaps / warnings | configure bursts at 921600 |
|---|---|---|---|---|
| rig firmware before (this PR as of 11.09., IPC) | 33.0 KB | 10 / – / 22 ms | 0 / 0 | 5 of 5 complete |
| (a) this rework, IPC, 8 KB ring, 512 queue, threshold 32 | 30.5 KB | 10 / 14 / 21 ms | 0 / 0 | 5 of 5 |
| (b) `CONFIG_ESP_MAIN_TASK_AFFINITY_CPU1`, no IPC, same buffers, scheduler on core 1 | 32.6 KB | 13 / 19 / 29 ms | 0 / 0 | 3 of 5 (two bursts lost a line each) |
| (c) affinity, 2 KB ring, 100-entry queue, default threshold | 40.3 KB | 11 / 17 / 24 ms | 0 / 0 | 0 of 5, UART0 deaf until a hardware reset |
| rig firmware restored (control) | 33.1 KB | 10 / 14 / 21 ms | 0 / 0 | 5 of 5 |

7. Affinity is worse than the IPC install on this rig: 13 instead of 10 ms median, 29 instead of 21 ms max, and lines lost at 921600. My reading: the main task now shares core 1 with the bus task (pinned there) and every UART interrupt. The IPC stack lines are the price, 2 KB on the ESP32; the variant is on a branch if you want to look at it, but I would drop it.
6. The buffers are needed. With the 2 KB ring and the 100-entry queue the 6.5 KB burst is lost entirely, and the core goes deaf: 142 line ends overflow the queue, the full ring disables the receive interrupts, and nothing ever pops a position to read. The flush from your point 2 cannot catch that when the ring is not larger than a line, so `static_assert(RX_RING_SIZE > CONSOLE_LINE_SIZE)` now says so. Cost of ring plus queue on the rig: 7.7 KB (c against b).

Heap on the rig: (a) is 2.5 KB below the firmware it replaces, 2 KB of it the doubled receive ring of the bus `Serial`; if that matters more than a maximal expander line, `CONSOLE_LINE_SIZE + 128` gets most of it back.

</details>

<details>
<summary>Cleanup</summary>

- `BUFFER_SIZE` is gone; `process_uart` uses `CONSOLE_LINE_SIZE` and `std::min`, the `read <= 0` guard stays.
- The drop messages agree on one number: `UART0 line of N bytes exceeds 2048 bytes and was discarded`, `UART0 input exceeds 2048 bytes without a line end and is discarded up to the next line end`, `console line of N bytes exceeds 2048 bytes and was dropped`. `Serial::read_line_error` is unchanged, its limit is the caller's buffer.
- The sizing rationale lives once above `RX_RING_SIZE` / `RX_PATTERN_QUEUE`; the IPC comment moved into the helper.
- Both `CONFIG_ESP_IPC_TASK_STACK_SIZE` lines carry a comment saying why they are there. The S3 default boot-loops (7cce04e); the ESP32 value is verified on hardware but not measured with a high-water mark, the comment says so.
- `Serial` UARTs get the same FIFO threshold of 32 through the shared helper.
- `esp_ipc.h` is only included in the helper now.
- `docs/machine_safety.md` names the 2048-byte limit and the two warnings next to the checksum paragraph.

</details>

<details>
<summary>Seen on the bench, not changed</summary>

- A child whose core line is between 2039 and 2043 payload characters fits its own console but not behind the parent's `p0: ` prefix; the parent replaces every such line with the echo warning, at the child's rate. Inherent to forwarding with a prefix.
- After `p0.flash()` the expander never comes back on `main` because `step()` only reads while `is_ready`; that is #269, not this PR. The child on the bench was flashed once and the core restarted to continue.
- Flash writes masking the UART ISR (`CONFIG_UART_ISR_IN_IRAM`) and the 178 ms a 2 KB line blocks at 115200 (visible on the bench: four 2048-byte core lines per 0.6 s): out of scope, as you say.

</details>

Verified and tested with Claude Fable 5.1; bench pair of two ESP32-D0WD (core with a `Serial(26, 27, 115200, 1)` expander link), F30 rig for the A/B, builds for esp32 and esp32s3.

falkoschindler (approved):
@JensOgorek thanks for the rework and the A/B, that settles 6 and 7 for me: the IPC install and the buffers stay. I went over the reworked branch once more (an AI-assisted pass, I did not verify every point on hardware) and pushed three commits on top of d6cdc92 instead of another round trip:

- ae18ceb: the discard branches in `process_uart` read exactly the measured bytes instead of calling `uart_flush_input`. The flush also dropped whatever arrived after the pattern pop, including the line end that terminates the run, so `discarding` stayed set and the next valid line went silently. The branch now also re-checks the pattern queue after measuring, so complete lines that landed in between are popped rather than counted as an unterminated run. Built for esp32, not bench-tested; your 9000-byte run and the 4000-byte line at 115200 are the two tests I would rerun.
- d37c40b: `Core::get_output` clears `output_overflow_reported` after every line that fits, so a second overflow after the output resumed is reported again instead of suppressed silently. The message now says "suppressed until it fits again".
- 216b9dc: the `s_` prefix row for static variables is gone from CONTRIBUTING.md. Nothing in the tree ever followed it (no `s_` identifier, 53 unprefixed statics); it came in with the guideline draft in #188.

Two things stay open, neither blocks the merge:

- The 2048-byte IPC stack on the ESP32 is verified for the boot path but not measured, as your sdkconfig comment says. Could you print `uxTaskGetStackHighWaterMark` of `ipc1` once after the install, ideally also with a deliberately oversized ring so the error path (`ESP_LOGE` on the IPC task) runs, and put the number into the comment in a follow-up PR? If the margin is thin, 4096 on both targets is the fallback.
- The Serial modules have the same deaf mode when their ring fills without a line end, and `Serial::read_line` still discards an overlong line one byte at a time. Both predate this PR and are now #300, with a shared helper in `uart_driver.cpp` as the proposal.

<details>
<summary>Review findings (lines as of d6cdc92)</summary>

**1. IPC stack on the ESP32 estimated, not measured** sdkconfig.defaults.esp32 (52) — only the successful install path was exercised; a failing install logs via `ESP_LOGE` on the same stack and could panic instead of returning the error. Follow-up PR with a high-water-mark reading, see above.

**2. Serial modules go deaf when the receive ring fills without a line end** main/modules/serial.cpp (121) — the IDF ISR disables RX on a full ring and enqueues no pattern position, so `has_buffered_lines` stays false until a reset; the UART0 guard has no counterpart here. Pre-existing, #300.

**3. The next valid line can be lost after an overlong run** main/main.cpp (428) — `uart_flush_input` in the `discarding` branch also removed a line end that arrived after the pop, so `discarding` stayed set and the following line was dropped without a warning. Fixed in ae18ceb.

**4. The overlong core line is reported only once** main/modules/core.cpp (257) — the flag was cleared only by `core.output(...)`, so after the line fit again a later overflow silenced the output without a message. Fixed in d37c40b.

**5. Complete lines can be counted as an unterminated run** main/main.cpp (429) — `buffered` was measured after the queue was found empty; more than 2048 bytes of complete lines arriving in between would have been discarded as one run. Needs a stall of about 2 ms between two pops, so a corner case, fixed alongside 3 in ae18ceb.

**6. Serial::read_line discards an overlong line byte by byte** main/modules/serial.cpp (145) — one driver call per byte, now up to 4096 of them per line, several milliseconds in one main-loop tick. Pre-existing, folded into #300.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20and%20wording%20from%20the%20model%3B%20which%20of%20them%20to%20fix%20on%20the%20branch%2C%20hand%20to%20the%20author%20or%20file%20as%20an%20issue%20was%20decided%20by%20the%20reviewer.&a=claude-code "claude-fable-5-1: Findings and wording from the model; which of them to fix on the branch, hand to the author or file as an issue was decided by the reviewer.")

JensOgorek:
### Falko's three commits pass on the bench and on the F30 rig; F30 still drops line pairs at 460800 until one more commit splits merged reads (60/60 and 10/10 clean with it)

ae18ceb and d37c40b behave as intended on hardware: an unterminated 9000-byte run and a 4000-byte line each give exactly one warning and every valid line after them runs, at 115200 and 921600, and a second overflow of the core line is reported again (bench ESP32 with this branch merged into current `main`, plus the F30 core). Replaying rosys' configure burst on the F30 core at 460800, however, this branch still loses two adjacent lines in 7 and 10 of 60 bursts (3 of 60 before your commits, the same pattern, so they are not the cause): the IDF driver queues only the last line end of each receive chunk, so two close line ends come back as one read that fails the checksum and both lines are dropped. fc81ef1 splits every read at its line ends; with it, 60 of 60 bursts and 10 of 10 full configures on F30 stored the whole script. I would add it to this PR (22 lines in `process_uart`) and merge after your look.

<details>
<summary>Falko's commits on hardware</summary>

| Check | Bench ESP32, this branch + `main` | F30 core |
|---|---|---|
| 9000 B without a line end, pause, `\n` + line | 1 warning, line runs (115200, 921600) | 1 warning, line runs (460800) |
| 9000 B + `\n` + line in one write, then another line | 1 warning, both lines run (115200, 921600) | 1 warning, both lines run (460800) |
| 3998 B line + line in one write, then another line | 1 warning, both lines run (115200, 921600) | 1 warning, both lines run (460800) |
| core line overflows, fits, overflows again | one error per episode, output resumes in between (115200, 921600) | – |
| 30 s on the bus with four peers at 460800 | – | loop 10 / 14 / 21 ms, 0 sequence gaps, 0 bad frames, 0 warnings |

The F30 column ran with fc81ef1 on top, which leaves the discard branches untouched. Scenario 2 is the one ae18ceb is about: the line end and the next line arrive while the run is being discarded, and the flush used to swallow them.

</details>

<details>
<summary>Remaining loss on F30 and its cause</summary>

Replay of rosys' configure burst (`!-`, 106 × `!+line@xx`, no pacing, byte-identical to `RobotBrain.configure()`) into the RAM copy of the F30 core at 460800, rosys stopped, Bluetooth on, four peers on the bus; compared with `core.startup_checksum()`:

| Firmware, all on the same frames base | Complete bursts | Full configure (`!.` + restart) |
|---|---|---|
| pre-#287 emulation (2 KB ring, 100-entry queue, default threshold, ISR on core 0) | – | 8/10 |
| this PR at c9a0db8, before your commits | 57/60 | 10/10 |
| this PR at 216b9dc | 53/60 and 50/60 | 9/10 |
| 216b9dc + fc81ef1 | 60/60 | 10/10 |

Every loss came with exactly one `Checksum mismatch while processing UART0` and removed two adjacent lines, mostly two short ones such as `bool p4_parked_zero = false` + `bool delta_4_interlock = true`. Stored like that, the next boot stopped at `error while loading startup script: unknown variable "delta_3_interlock"` (the failed one of the ten full configures on 216b9dc), the same message a field robot reported for `estop_active` after a configure.

A diagnostic build that splits the read and reports it found a read holding two or three lines in 19 of 60 bursts, each of which would have been a lost pair; that build also stored 60 of 60 bursts and 10 of 10 full configures. In IDF 5.3.1, `uart_rx_intr_handler_default` enqueues one pattern position per FIFO read, found by `uart_find_pattern_from_last` from the end of the chunk (esp_driver_uart/src/uart.c (1122)), so when two `\n` land in the same chunk the earlier one is never queued and `uart_read_bytes(pos + 1)` returns both lines. On an idle bench ESP32 the same burst stayed clean 60/60 at 921600 without the fix; it takes the interrupt load of a real core to show it. With a field robot's 99-line core script the pre-#287 emulation stored 8 of 10 bursts complete under the F30 load, and under an F27-like load (BNO055 read in the loop, 314-byte core line, loop 46 ms) only 4 of 10, hanging the core completely in 4 of them until an EN pulse; this PR alone gets that load to 10 of 10.

</details>

<details>
<summary>What fc81ef1 changes</summary>

- `process_uart` still reads up to the queued line end, then hands every `\n`-terminated piece of the read to a new `process_uart_line`, which checks and runs it on its own.
- A failing line reports `error processing uart0: …` in place, the text `app_main` prints, so the lines after it in the same read still run. For single-line reads the only difference is that lines queued after a failing one now run in the same loop pass instead of the next.
- Limit: if the missed line end sits between two lines that together exceed 2048 bytes, the read is still discarded as one overlong line. Configure lines stay well below 1 KB, so this has not come up.

</details>

<details>
<summary>Serial modules</summary>

`Serial::read_line` (expander links, `SerialBus`) pops pattern positions the same way, so the same merge can drop bus or expander lines under load. Not measured; it would fit into #300.

</details>

Verified and tested with Claude Opus 5.5; bench ESP32-D0WD (this branch merged with `main`, with and without fc81ef1), F30 rig (lizard-zz frames base with this branch), builds for esp32.

falkoschindler (approved):

Issue #289, opened 2026-09-08
core.frame drops all frames on a peer until the first poll reveals the coordinator


JensOgorek:
> To be checked first whether this is really relevant (Jens); noticed while bringing up `core.frame` on the F30 plexus (branch `frames-experiment`).

### Observation

A serial-bus peer learns its coordinator from the first `__POLL__` it receives. Until then `core.frame` frames have no route and are dropped (`core.frame_drops` counts them): on F30 every plexus shows ~220 dropped frames after boot and none afterwards, i.e. roughly 11 s of telemetry at 2 frames per 100 ms.

### Question

Is that acceptable (the frames are telemetry and the values are stale after a few seconds anyway), or should the peer buffer the latest frame per id and send it once the coordinator is known, so the first poll delivers a complete state? Related: rosys notices a plexus as "stopped pushing" only via the frame heartbeat, so a long silent start looks like a dead peer for those seconds.

falkoschindler:
@JensOgorek could you add a pointer to #290 in the description? `core.frame` and `core.frame_drops` only exist on that branch, so without the link the issue is hard to place when it shows up on its own in the issue list. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Substance%20from%20the%20author%2C%20wording%20from%20the%20model.&a=claude-code "claude-fable-5-1: Substance from the author, wording from the model.")

Issue #288, opened 2026-09-08
Startup script is parsed as one tree; an allocation failure while loading it leaves the core in a boot loop


JensOgorek:
> To be checked first whether this is really relevant in practice (Jens); observed once on the F30 core while adding gripper mirrors to a 142-line startup.

### What happened

On the F30 core (ESP32, Bluetooth on, four serial-bus peers) a 154-line startup script (the rosys-generated F30 configuration with 12 extra `float` mirror variables and a 93-field `core.output`) made the core abort while loading it — `abort()` via `std::terminate` about 100 ms after the `battery_control` lines, deterministic on every boot, so the core sat in a boot loop. The same 12 variables and the same `core.output` line applied volatile over the console on the previous 142-line startup worked and left 19.6 KB of heap. The previous startup (0b16) boots fine.

### Why (reading the code)

`app_main` runs `process_lizard(Storage::startup.c_str())`, i.e. the whole startup is parsed as one owl tree and then executed statement by statement, so during execution the heap is lower than at runtime by the size of that tree. The `try`/`catch` around it only catches `std::runtime_error`; a `std::bad_alloc` (or any other exception) escapes and terminates the task. Because the startup runs before the main loop, the console is never serviced, so `!-`/`!.` cannot fix it — the only way out was erasing the NVS partition with esptool through espresso and writing the startup again.

### Ideas

- Process the startup line by line (each line its own parse tree, as the console path does), so the peak heap does not scale with the script length.
- Catch every exception while loading the startup (or at least `std::bad_alloc`) and continue booting with an empty startup and an error line, so the console stays reachable.
- Optionally: a startup that crashed the last boot (core dump present) is skipped once.

F30 later moved its telemetry to frames (fewer variables on the core), so the trigger is gone there; whether other robots come close to the limit is the open question.

falkoschindler:
> Process the startup line by line

I think it's non-trivial to divide the startup script into independent parts that can be parsed one by one.

> Catch every exception while loading the startup (or at least `std::bad_alloc`)

That sounds much more feasible.

JensOgorek:
### Going with the catch: PR follows

[...]

PR #290, opened 2026-09-08
Send peer telemetry as binary frames over the serial bus and pass it through to the console

JensOgorek:
### Motivation

A robot brain with several serial-bus peers mirrors every peer field as a core variable: each peer pushes `!!`-assignments, the core parses them with the owl parser, stores the mirrors and prints them in its `core.output` line. That is fine for one arm (F27: 3 push lines per 100 ms, 12 mirrors, a 300-byte line). With four arms on F30 every stage breaks: the parse trees and mirrors ate the heap (22 KB free with Bluetooth; the 154-line startup with the gripper mirrors no longer loaded, #288), twelve parses per 100 ms plus a blocking 453-byte line pushed the loop to 18–24 ms instead of 10, the line forced 921600 baud on the console where the Jetson overruns, and every peer needed three bus messages per round.

The fix has two parts, which this PR carries together: a **protocol** (binary telemetry frames the console can carry next to text lines) and a **topology** (peers send frames over the bus, the coordinator passes them through unparsed). rosys reads the frames (zauberzeug/feldfreund#636).

### Protocol

`core.frame(id, "field[:type[digits]] …", interval_ms)` streams the listed properties as a binary frame: numbers as little-endian integers scaled by 10^digits (or float32), bools packed as bits, in a body `0x00 | src | id | seq | millis[4] | len | payload | crc16` that travels COBS-encoded between the text lines of the console.
`core.frame_add` continues a field list that does not fit one startup line.
The full description is the new docs page [Telemetry Frames](https://github.com/zauberzeug/lizard/blob/frames-experiment/docs/telemetry_frames.md) (`docs/telemetry_frames.md`, linked from the module reference).

### Topology (serial bus)

- A peer remembers the node that polls it and hands each frame to `SerialBus::send_frame`, which byte-stuffs the body for the line-based bus (marker `0x01`, escape `0x7d`, `byte ^ 0x50` for `00 09 0a 0d 20 7d`). Stuffing is needed because a bus message is a C string terminated by `\n`, checked with `@xx` and passed through `strip()`; those six byte values would end or trim the message. Stuffing costs about 2 % on typical payloads, Base64 would cost 33 %.
- The coordinator recognises the marker before the OTB, echo and `!` branches, checks length, CRC and that `src` is the sender, and writes the body to the console as a COBS frame — no variables, no parsing. Malformed frames are counted (`frame_errors`, warning at most once per second); frames a peer cannot hand over before its first poll are counted in `core.frame_drops` (#289).
- The ESP32-S3 IPC task stack is 4 KB (installing the UART driver from core 1 overflows 2 KB there); that fix lives in #287, which this branch merges.

### Measured on F30 (core ESP32 + 4 plexus S3, bus 460800, Bluetooth on)

| | text mirrors | frames (one per peer) |
|---|---|---|
| core loop | median 18–24 ms, p90 36–40 | median 10 ms, p90 14, max 23 |
| core heap free | 22–26 KB, gripper startup does not load | 35–37 KB, flat over 300 s |
| console | 453-byte line every 10 ms at 921600 | 144-byte line plus 47 bytes per peer per 100 ms, runs at 460800 |
| bus per peer per 100 ms | 3 text messages | 1 frame, ~57 bytes |

<details>
<summary>Frames versus an ASCII pass-through (same peers, same 15 fields, 300 s each, 2026-09-10)</summary>

The pass-through alternative keeps text: the peer sends `__ECHO__T1 <seq> <millis> <values>` lines that the coordinator prints unparsed as `bus[N]: …` (Lizard's existing echo prefix; without a prefix the coordinator parses the line as a command). Four lines per peer, because 15 field names do not fit one startup line either. Both runs with the core's own telemetry as a frame and rosys stopped.

| | frames | ASCII pass-through |
|---|---|---|
| console total | 12.9 KB/s (28 % of 460800) | 17.3 KB/s (38 %), 18.2 KB/s with the timestamp field |
| peer telemetry on the console | 1.9 KB/s, 47 B per peer per 100 ms | 6.2 KB/s, 168 B in four lines |
| bus per peer per 100 ms | 1 message, ~57 B, 1.2 ms | 4 messages, ~210 B, 4.6 ms |
| bus load, 4 peers | 8 % at 460800, 30 % at 115200 | 21 % at 460800, 83 % at 115200 |
| command round trip, `bus.send(N, "core.version()")` to the echoed reply, 100 pings per peer | median 40.3 ms, p90 60.4, max 60.5, 400/400 answered | median 40.3 ms, p90 60.4, max 60.6, 400/400 answered |
| core loop | 10 / 14 / 23 ms | 10 / 12 / 21 ms |
| peer interval from the peer's millis | median 100 ms, max 111 | median 110 ms, max 112 (rule `> 100` on 10 ms ticks) |
| losses in 300 s | 0 | 1 line (caught by the XOR checksum this time) |
| loss detection | CRC16 per frame | 1-byte XOR, 1 in 256 slips through |
| timestamp, layout check | in the header, fixed length | only if the generator adds a field; word count |

With four peers on a 460800 bus both variants are far from saturation and the command round trip is the same, quantised by the coordinator's poll cycle (20, 40, 60 ms); the difference is bytes (3.3× on the console, 3.6× on the bus for the peer share), the wire time per peer answer, and the detection.

</details>

<details>
<summary>Console reliability (Jira FF-676)</summary>

At 921600 the Jetson UART (PIO receive, 36-byte FIFO, interrupt shared with the camera ethernet on CPU0) overruns about every 4 s; the lost bytes pass the XOR checksum once in 256 times. Two stop bits reduce that by a third, 460800 by two orders of magnitude (84 overruns per hour), the frame telemetry makes the short line at 460800 possible, and with the core line as a frame (feldfreund `CoreFrames`) every remaining overrun is detected by CRC. Preventing them entirely needs RX-DMA on the Jetson.

</details>

### Follow-up

- Peer behaviour before the first poll: frames are counted in `core.frame_drops` and dropped (#289).
- The two parts could also be reviewed as two PRs: protocol (frame definition, body, COBS, monitor, docs) first, bus transport (stuffing, pass-through, `core.frame_add`) second. Say if you prefer that split.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Implementation%20and%20description%20drafted%20by%20the%20model%2C%20directed%20and%20hardware-tested%20by%20the%20author.&a=claude-code "claude-fable-5-1: Implementation and description drafted by the model, directed and hardware-tested by the author.")

### Progress

- [x] I chose meaningful labels (if GitHub allows me to so).
- [x] Tests with a real hardware have been successful (or are not necessary).
- [x] Documentation has been added (or is not necessary).


main/utils/frame.h
```diff
@@ -0,0 +1,36 @@
...
+// length field and CRC of a complete body are consistent
+bool verify_body(const uint8_t *body, size_t length);
+// write a complete body to the console as 0x00 | COBS(body) | 0x00
+void write_console(const uint8_t *body, size_t length);
+// build and write in one go (console sink)
+void write(uint8_t src, uint8_t id, uint8_t seq, uint32_t millis, const uint8_t *payload, size_t length);
+
+// bus transport: 0x01 marker plus the body (without its leading 0x00) with 0x00/0x09/0x0a/0x0d/0x20/0x7d escaped
+// as 0x7d, byte ^ 0x50; the result never contains those bytes, so it survives the bus' C-string handling and strip()
+size_t bus_stuff(const uint8_t *body, size_t length, char *output, size_t capacity);
```
thomas-zauberzeug, line 32: Wozu wird ein stuffing überhaupt gebraucht? Rückwärtskompatibilität zu ASCII-Botschaften?

main/modules/core.cpp
```diff
@@ -29,6 +35,147 @@ void Core::step() {
...
+            break;
+        default:
+            throw std::runtime_error("unsupported frame field type");
+        }
+        if (field.type == '?') {
+            if (value != 0) {
+                bits[bit / 8] |= 1 << (bit % 8);
+            }
+            ++bit;
+            continue;
```
evnchn, line 101: **`core.frame` with more than 128 boolean fields writes past `bits[16]` on the stack.**

- [`uint8_t bits[16]`](https://github.com/zauberzeug/lizard/blob/405b6a1bbc229368daf8171f76c695cee5f2c9a1/main/modules/core.cpp#L78) is indexed as `bits[bit / 8]`, with `bit` incremented once per `?` field and no bound on `bit`.
- The only size check, [`pos + bit_bytes > sizeof(payload)`](https://github.com/zauberzeug/lizard/blob/405b6a1bbc229368daf8171f76c695cee5f2c9a1/main/modules/core.cpp#L153-L156), guards `payload` — 200 bytes, i.e. 1600 bools — so it cannot fire before `bits` (16 bytes, 128 bools) is overrun.
- [The trailing `memcpy`](https://github.com/zauberzeug/lizard/blob/405b6a1bbc229368daf8171f76c695cee5f2c9a1/main/modules/core.cpp#L157) then also reads `bit_bytes` (17 and up) out of the 16-byte array, so all-false fields overflow too.
- Nothing caps the field count: one 2048-byte console line holds several hundred elements, the same variable may be listed repeatedly, and `core.frame_add` exists to append across lines.

Reproduced against `405b6a1` by compiling the bool-packing path on the host with `-fsanitize=address,undefined` — 128 fields pack fine, 129 gives `stack-buffer-overflow … 'bits' <== Memory access at offset 48 overflows this variable`.

REQUIRED:

```suggestion
        if (field.type == '?') {
            if (bit / 8 >= sizeof(bits)) {
                throw std::runtime_error("too many boolean frame fields");
            }
            if (value != 0) {
                bits[bit / 8] |= 1 << (bit % 8);
            }
            ++bit;
            continue;
```

falkoschindler (commented):
@JensOgorek Thanks for this PR, the F30 numbers make a convincing case for binary telemetry that the coordinator passes through without parsing. I went through the diff with AI support and collected the findings below; I did not verify every point myself. I'm posting them now so they don't get lost, not as a request to fix them right away: we are still thinking about the overall concept (for example holding the latest frame per id in a slot instead of queueing frames, Base64 lines instead of stuffing and COBS, and a layout line so rosys doesn't have to know the frame format), and we'll discuss that in the team first. Some of the findings may resolve themselves depending on where that ends up.

Two of them are worth knowing now, because F30 runs this branch: finding 1 can stall a peer's main loop for 50 ms per frame once the coordinator stops polling, and I suspect it is where the ~220 drops after boot in #289 come from, since a full outbound queue is the only path that increments `core.frame_drops`. Finding 6 is @evnchn's thread. @thomas-zauberzeug's question about the stuffing is part of the concept discussion.

<details>

**1. A full outbound queue blocks the peer's main loop.** main/modules/core.cpp (175), main/modules/serial_bus.cpp (328) — `send_frame` goes through `enqueue_outgoing_message`, which waits up to 50 ms for a free slot. After the first poll, `is_peer_with_coordinator()` stays true. If the coordinator then stops polling for more than about 1.6 s (reboot, OTA, unplugged cable, a coordinator busy with its startup), the 32-slot queue is full and every frame costs 50 ms of loop time. At two frames per 100 ms the peer's loop practically stops. The stale frames also take the slots the `static_assert` keeps free for OTB. Suggestion: enqueue frames with a zero timeout and count a drop when the queue is full, and only enqueue a frame while more than `BUS_OTB_WINDOW` slots are free.

**2. Frames before the first poll go to the peer's console, not into `core.frame_drops`.** main/modules/core.cpp (163), docs/telemetry_frames.md (52) — while `coordinator_id` is 0, `is_peer_with_coordinator()` is false, so `emit_frame` writes the frame to the peer's own UART0 with `src=0`. The docs and the PR description say these frames are counted as drops, and the "no coordinator has polled yet" branch in `send_frame` is unreachable. The `coordinator()` accessor in serial_bus.h (19) is unused.

**3. Frames on a port expander break its line stream to the core.** main/utils/frame.cpp (69) — `write_console` writes the COBS frame to UART0 without a newline. On an expander, that is the line stream the core's `Expander` module reads and checksums, so every frame corrupts the next text line. Nothing rejects or documents this.

**4. Typos in a frame definition only fail at runtime.** main/modules/core.cpp (61) — with an explicit type, `parse_frame_fields` checks neither the type letter nor whether the property exists. `core.frame(1, "motor.positon:i", 100)` is accepted, and `Core::step` then throws every 100 ms. The exception also aborts the loop over the frames, so later frames are skipped too. Suggestion: resolve the property at definition time, store the `Variable_ptr` in `frame_field_t` and check the type letter against `?bBhHiIf`.

**5. Out-of-range values wrap silently.** main/modules/core.cpp (117) — the scaled value is converted with a plain `static_cast` from `double`. Out of range, negative for an unsigned type, or NaN, that is undefined behaviour, and in practice the value wraps while the CRC stays valid. Examples: `motor.position:h2` above 327.67, a negative speed as `H`, and `core.millis` as the inferred `i` after 24.8 days of uptime. Suggestion: clamp to the target type's range before the cast, with a fallback for NaN.

**6. More than 128 bool fields overflow a stack buffer.** main/modules/core.cpp (78) — same as @evnchn's thread. Only a frame assembled with `core.frame_add` could get there, but sizing `bits` like `frame::MAX_PAYLOAD` lets the existing size check catch it.

**7. Frame id and interval are narrowed without a range check.** main/modules/core.cpp (229) — `core.frame(256, ...)` silently defines frame 0, and a negative interval becomes about 49 days. Minor.

**8. `write_console` spins forever without a UART0 driver.** main/utils/frame.cpp (78) — `uart_tx_chars` returns -1 without a driver, and the loop only advances on positive values. The console driver is always installed, so this only matters if the setup in main.cpp already failed. Minor.

**9. A second parser for the `module.property:suffix` syntax.** main/modules/core.cpp (47) — `parse_frame_fields` repeats the parsing of `core.output`. A shared helper would keep the two from drifting apart. Minor.

**10. Modules and properties are looked up on every emission.** main/modules/core.cpp (80, 161) — resolved once as in finding 4, this goes away. Minor.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5-5&t=Findings%20and%20wording%20from%20the%20model%2C%20checked%20against%20the%20code%3B%20the%20author%20chose%20the%20direction%20and%20what%20to%20send.&a=claude-code "claude-opus-5-5: Findings and wording from the model, checked against the code; the author chose the direction and what to send.")

PR #291, opened 2026-09-09
Add `nan` and `inf` literals to the language

falkoschindler:
An expander's `!!` broadcast prints number properties with `%f`, so a NaN property arrives at the host as `nan`. The grammar only knows that as an identifier, so the host reports `unknown variable "nan"` on every step and drops the remaining properties of that broadcast line. `temperature_sensor.temperature_c` already hits this with an unplugged NTC, and #277 adds `offset_<id>` properties that are NaN by design while no valid estimate exists. The review of #277 first worked around it by leaving NaN properties out of the broadcast, which replaced a loud error with a silently stale proxied value; this PR fixes the cause instead, and #277 drops the workaround.

**Change**

- `language.owl`: `nan` and `inf` as number literals next to `true` and `false`; `main/parser.h` regenerated with the pinned owl (`gen_parser.sh`). Regenerating the unchanged grammar reproduces the committed `parser.h` byte for byte, so the large diff is purely the two literals.
- `main.cpp`: two cases in `compile_expression` returning a `NumberExpression` with `NAN` and `INFINITY`; `-nan` and `-inf` go through the existing negate.
- `docs/language.md`: the literals are named in the data types section.

**Breaking change**

`nan` and `inf` are reserved words now; a variable or module named exactly `nan` or `inf` no longer parses. Owl picks the longest token, so identifiers such as `info`, `nano` or `infinity` are unaffected. A host running firmware older than this PR still reports `unknown variable "nan"` for NaN properties of a newer expander, exactly as before, so update the core before its expanders.

**Verification**

- esp32s3 build clean.
- Host-side parse test with the generated parser: `x = nan`, `y = -inf`, `z = inf`, `bus.offset_2=nan;bus.offset_2_accuracy=-nan;` and `w = nan > 0` parse as expected; `core.info()`, `nano = 1` and `infinity = 2` stay identifiers; `nan = 1` and `inf = 1` are rejected.

#277 depends on this PR. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Diagnosis%20from%20the%20review%20discussion%3B%20grammar%20change%2C%20parser%20test%20and%20wording%20by%20the%20model.&d=2026-09-09&a=claude-code "claude-fable-5-1: Diagnosis from the review discussion; grammar change, parser test and wording by the model.")


docs/language.md
```diff
@@ -207,6 +207,8 @@ Lizard currently supports five data types:
 
 Note that identifiers cannot be created via variable declarations, but only via constructors.
 
+Floating point numbers can also be `nan` (not a number) and `inf` (infinity), both as literals and in outputs; comparisons with `nan` are always false.
```
evnchn: **`docs/language.md` line 210 says comparisons with `nan` are always false — `!=` is always true, and `!=` is the only `nan` test the language has.**

- [`UnequalExpression::evaluate_boolean()`](https://github.com/zauberzeug/lizard/blob/b27790fd750c2399c99e9ab96cc22648acc4e8f4/main/compilation/expressions.cpp#L354-L356) returns the native `left != right`, so `nan != nan` and `nan != 0` are both true.
- `>`, `<`, `>=`, `<=` and `==` are false as documented, `nan == nan` included.
- There is no `isnan` in the language, so `x != x` is the test the sentence rules out.

Consequence: `when x != 0 then ... end` reads as inert on `nan` under the documented rule and actually fires. Built and ran the evaluator on a host (`main/compilation/` plus `string_utils`, one stub for `Module::get_property`): `nan == nan` false, `nan != nan` true, `nan != 0` true.

REQUIRED — the sentence is wrong for `!=`:

```suggestion
Floating point numbers can also be `nan` (not a number) and `inf` (infinity), both as literals and in outputs; every comparison with `nan` is false except `!=`, which is always true, so `x != x` tests for `nan`.
```

JensOgorek (approved):
### Bench test of `7312188` on an ESP32-S3 passed: literals, negation, reserved words and the `%f` round-trip behave as described, cost is +35 KB flash — approving

Flashed the branch on a bench ESP32-S3 and ran the cases from the description plus the round-trip that motivates the PR: `0.0 / 0.0` prints `nan`, `1.0 / 0.0` prints `inf`, both read back as literals, and a `when x != x` rule fires on a NaN variable, so the comparison sentence in the docs (with Evan's `!=` correction, already applied) matches the hardware. The reserved words are rejected exactly as the description says, `nano` and `infinity` stay identifiers. The only number worth knowing before merging: the regenerated parser costs 35 KB of flash (rodata), RAM is unchanged. Good to merge; #277 waits for it.

<details>
<summary>Bench cases (ESP32-S3, `7312188`, console at 115200)</summary>

| command | result |
| --- | --- |
| `float x = nan`, `float y = -inf`, `float z = inf`, `core.print(x, y, z)` | `nan -inf inf` |
| `core.print(nan != nan)` / `nan == nan` / `nan > 0` / `nan < 0` / `nan != 0` | `true` / `false` / `false` / `false` / `true` |
| `core.print(inf > 1e300)` / `core.print(-inf < 0)` | `true` / `true` |
| `float nano = 1`, `int infinity = 2`, `core.print(nano, infinity)` | `1.000000 2` |
| `nan = 1`, `inf = Input(1)` | `error: unexpected token at range 4 5 "="` |
| `float nan = 3`, `int inf = 4` | `error: unexpected token at range 6 9 "nan"` / `… 4 7 "inf"` |
| `float q = 0.0 / 0.0`, `float r = 1.0 / 0.0`, `float s = -1.0 / 0.0`, `core.print(q, r, s)` | `nan inf -inf` (newlib prints the positive spelling for `0.0 / 0.0` here) |
| `float q2 = -nan`, `core.print(q2)` | `-nan` |
| `when x != x and not hit then hit = true; core.print("rule fired on nan") end` | fires once |
| `core.output("x y z q r s")` | `core nan -inf inf nan inf -inf` |
</details>

<details>
<summary>Flash and parser</summary>

- `idf.py size` A/B for esp32s3, main `eb49587` vs this branch: image 1 288 448 → 1 323 628 B (+35 KB), `.rodata` 445 196 → 478 488 B (+33 KB, the grown `action_table`/`actions` of the generated parser), `.text` +1.9 KB, IRAM/DRAM/bss identical. The app partition is 2 031 616 B, so the image sits at 65 %.
- Regenerating `parser.h` with the pinned owl (`1b8b7f4`) on my machine does not reproduce the committed file byte for byte, but the same is true for main's committed `parser.h` (3278/3005 lines differ there), so the tables depend on the owl build, not on this PR. The two-line grammar change is the whole semantic difference; parsing the repo's example scripts with the base and the PR parser gives identical trees.
</details>

<details>
<summary>Code review (non-blocking)</summary>

- The description names `nan` and `inf` as reserved words, the docs sentence does not; one more sentence in `docs/language.md` ("`nan` and `inf` are reserved words like `true` and `false`") would spare the next person the `unexpected token` hunt.
- Everything typed `numbery` can now receive `nan`/`inf` by one token: `at inf do … end` casts to `int64_t`, `RmdMotor::speed(inf)` / `StepperMotor::position(inf)` cast into the CAN or step payload, `Output::pulse(inf)` latches the output high, `MotorAxis::can_move()` passes NaN through its end-switch guard because `nan < 0` and `nan > 0` are both false. All of these were reachable before via `0.0 / 0.0` and `1.0 / 0.0` (the float division has no zero guard), so nothing new is broken, but an `isfinite` check at the two or three casting sites would be a cheap follow-up now that the values are spelt out in the language.
- `<cmath>` is the only C++ math header in `main/`; the rest of the tree uses `<math.h>`. Cosmetic.
</details>

<details>
<summary>Not covered</summary>

- The proxied path (expander broadcasting `nan` into a core with this firmware) was not run; the bench boards are not wired to each other and F30 has no expander-mounted bus. The parse side of it is the bench case above, the emitting side is unchanged `%f`.
</details>

Reviewed and tested with Claude Fable 5.1 (`claude-fable-5-1`) in Claude Code; bench run, size measurement and code notes are the model's work, the approval is mine.

PR #292, opened 2026-09-11
Catch every exception while loading the startup script

### LINKED ISSUE #288: Startup script is parsed as one tree; an allocation failure while loading it leaves the core in a boot loop

JensOgorek:
### Motivation

Fixes #288 and decides the open question of #246 as explicit fail-safe.

`app_main` only caught `std::runtime_error` around `Storage::init()` and `process_lizard(startup)`. A `std::bad_alloc` thrown while the startup statements were executed escaped, terminated the main task and left the core in a boot loop with the console never serviced, so `!-`/`!.` could not recover it (observed on the F30 core with a 154-line startup). Merely widening the catch would have kept the state #246 describes: a core that boots with part of its startup and looks alive.

### Implementation

A boot counter in RTC memory (`main/utils/boot_guard.{h,cpp}`) makes a failed startup fail safe:

- An exception while loading the startup script (now `std::exception` and a catch-all) stores its reason, prints `error while loading startup script: …` and `restarting (failed boot N of 2)`, and restarts. A panic, watchdog or brownout reset during the startup or the first 5 s of the main loop counts as a failed boot as well.
- After two failed boots the startup script is skipped, one line `error: startup skipped after 2 failed boots (<reason>)` names the cause, and the core reaches `Ready.` with a working console so the script can be fixed with `!-`, `!+` and `!.`. No `core.output` runs, so RoSys stays "not ready".
- The counter is cleared after 5 s of main loop, by a power-on or EN reset, and a deliberate `esp_restart()` (`core.restart()`) never counts, so RoSys configuring and restarting twice in a row does not trigger the skip.

Hardenings from the review: a 512-byte emergency pool for C++ exceptions in both sdkconfig defaults, so `std::bad_alloc` can still be thrown when the heap is exhausted; the console loop catches `std::exception`, so `!.` under memory pressure no longer kills the main task; `Storage::read()` checks its `malloc`; `SerialBus` gets a destructor that stops its communication task, frees its queues and unregisters its echo callback, so a startup that fails after creating the bus does not leave a task and a callback on a freed object. A storage error while reading the startup (other than a missing script) is now reported with `error while loading startup script: …` and the core continues with an empty script instead of restarting, so a freshly flashed or erased core reaches `Ready.`.

Tested on a bench ESP32 and on the F30 core (rb68), see the comment below: a failing module or a panic in the middle of a startup gives two restarts and then a skipped startup with a working console, quick deliberate restarts do not count, and a repaired startup loads again. The original heap exhaustion of #288 could not be reproduced with this build, because the F30 startup needs modules from lizard-zz.

### Progress

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -467,8 +467,10 @@ void app_main() {
     try {
         Storage::init();
         process_lizard(Storage::startup.c_str());
-    } catch (const std::runtime_error &e) {
+    } catch (const std::exception &e) {
         echo("error while loading startup script: %s", e.what());
+    } catch (...) {
+        echo("error while loading startup script: unknown exception");
     }
 
     bus_backup::save_if_present();


falkoschindler (changes requested):
@JensOgorek thanks for the quick fix. I ran a review over it and checked the findings against the code and against how RoSys detects a ready core. My conclusion: the three lines are correct, but they widen the door to the state #246 describes, a core that boots with part of its startup and looks alive. I think we have to settle that question before merging, so I am handing the PR back with a proposal. The full review is folded in at the end so the reasoning stays traceable.

### What I found

- **The catch may not fire in the #288 scenario.** `CONFIG_COMPILER_CXX_EXCEPTIONS_EMG_POOL_SIZE` is unset in both sdkconfig defaults and the ESP-IDF default is 0 (Kconfig 474-480, `cxx_init.cpp` passes it straight to libstdc++ as the arena size). Without a pool, `__cxa_allocate_exception` has to malloc the exception object itself. If the heap is exhausted by small allocations, that malloc fails too and libstdc++ calls `std::terminate` before any catch runs. Only a failed large allocation while small ones still succeed is caught. The hardware test threw a synthetic `bad_alloc` with heap available, so it did not exercise this.
- **RoSys does not parse `Ready.` at all.** `is_ready` means a `core` message has arrived, and those only start once `core.output(...)` has run. In the generated startup that is the second to last line (rosys `robot.py:67`), so a failure anywhere before it leaves RoSys on "not ready" forever and motor commands raise `EspNotReadyException`. RoSys robots are protected by the ordering, by accident. Hand-written startups or `when` rules placed after `core.output` are not.
- **The checksum check cannot see a partial load.** `core.startup_checksum()` sums the stored text, not what was executed. A half-loaded startup with a matching checksum looks correct to RoSys.
- **The partial state is not safe to keep.** A `bad_alloc` late in the `SerialBus` constructor now leaves a running task and an echo callback pointing at a freed object (no destructor). `bus_backup::restore_if_needed()` sees "no bus" after a partial startup, erases the user's `Serial` modules and fails to create the backup because the UART is still in use. Both were terminate-and-reboot before, now they are live states.

### Proposal: fail safe after a reboot instead of continuing with half a startup

Falko's idea, which I think is stronger than any catch because it does not depend on an exception being throwable and also covers panics and NULL derefs:

- **Boot counter in RTC memory** (`RTC_NOINIT_ATTR`, survives panic and soft reset, cleared by a power cycle). Increment before `process_lizard(startup)`, reset to 0 after a few seconds of main loop. If the counter is already at 2 on entry, skip the startup entirely, print `error: startup skipped after 2 failed boots`, and continue to the console. No `core.output`, so RoSys stays "not ready", and `!-` / `!.` work again. That alone fixes the boot loop of #288 regardless of exceptions.
- **A startup exception no longer continues with the modules created so far.** Store the reason in RTC memory, `esp_restart()`, and let the next boot skip the startup and print the reason. No teardown needed, no ordering dependence, one clear message. The widened catch from this PR stays useful for exactly that message.
- **This decides #246 as "explicit fail-safe".** A wrong IMU driver then means no modules at all instead of the ones before the failing line. Optional modules can come later as a refinement. That is the trade-off we should agree on before building it.

What that means for the findings below: the `bus_backup` one disappears, and four small hardenings remain worth doing in this PR: the emergency pool in both sdkconfig defaults, widening the `process_uart` catch so the console survives a `bad_alloc` in `!.`, a NULL check in `Storage::read()`, and a `SerialBus` destructor plus an unregister for the echo callback so the unwinding after a late `bad_alloc` does not leave a task and a callback on a freed object. The last one is about twenty lines in two files, `SerialBus` is the only module that spawns a task or registers an echo callback.

If you agree, would you extend this PR with the boot counter and the restart-on-failure? I would keep the module step, rule and routine catches narrow: an unexpected exception mid-step is a bug or memory pressure, and a reboot that drops the outputs and takes RoSys to "not ready" is the safer reaction than a printed line and a control loop in an undefined state.

<details>
<summary>Full review (6 findings)</summary>

**1. std::bad_alloc may not be throwable in the real #288 scenario.**

Neither sdkconfig.defaults.esp32 (2) nor sdkconfig.defaults.esp32s3 (3) sets CONFIG_COMPILER_CXX_EXCEPTIONS_EMG_POOL_SIZE, and the ESP-IDF v5.3.1 default is 0 (Kconfig 474-480). components/cxx/cxx_init.cpp (21-25) returns that value from __cxx_eh_arena_size_get(), so libstdc++ has no emergency arena. When the heap is exhausted by small allocations (string nodes, map nodes), __cxa_allocate_exception cannot malloc the exception object and libstdc++ calls std::terminate before the new catch ever runs. Only a failed large allocation while small ones still succeed is caught. Suggestion: set the pool to a few hundred bytes in both sdkconfig defaults. It is reserved once at boot, so the cost is fixed and small.

**2. The main loop still catches std::runtime_error only.**

main/main.cpp (428, 488, 506, 515) keep the narrow catch around run_step, process_uart, rules and routines. The console session this PR wants to keep alive dies on the first non-runtime_error exception, for example bad_alloc in Storage::put on `!.` or std::invalid_argument from stoi. Scenario: core boots to Ready. after a caught startup bad_alloc with the heap still tight, the user sends `!-battery_control` and `!.`, Storage::put throws bad_alloc, the task terminates, the reboot loop continues. Widening the process_uart catch (488) belongs in this PR. For run_step, rules and routines I would deliberately stay narrow, see above.

**3. A bad_alloc late in the SerialBus constructor now becomes a use-after-free instead of a terminate.**

main/modules/serial_bus.cpp (57) spawns the communication task with `this`, then (63) registers an echo lambda capturing `this`. If push_back at 63 or Global::add_module at main/main.cpp (216) throws bad_alloc, the half-built SerialBus is destroyed during unwinding, but there is no destructor that deletes communication_task or unregisters the callback (main/utils/uart.cpp has no unregister). Before this PR that was a clean terminate and reboot, now the task and the echo callback keep dereferencing freed memory. With restart-on-failure the window shrinks to the time between catch and esp_restart(), but a destructor that deletes the task and the queues and unregisters the echo callback closes it properly. main/utils/uart.cpp only has register_echo_callback, so it needs a matching unregister, for example by returning a handle. Fix in this PR.

**4. Storage::read() mallocs unchecked.**

main/storage.cpp (49) passes a possible NULL to std::string(value), and for size 0 it reads uninitialized memory. Under the same low-heap condition this is a LoadProhibited panic inside Storage::init(), before any startup statement runs, so no catch helps there. The boot counter covers the loop, a NULL check throwing runtime_error gives a readable message instead of a backtrace.

**5. bus_backup::restore_if_needed() after a partial startup.**

main/utils/bus_backup.cpp (38-82) erases every Serial module and creates _backup_serial/_backup_bus when no SerialBus exists. If the startup fails between the Serial and the SerialBus line, the user's serial name disappears from Global::modules while an Expander still holds it, and make_shared<Serial> fails with "serial interface is already in use". This was already reachable with a runtime_error from any module constructor before this PR. With restart-on-failure there is no partial startup any more, so the restore only runs in the case it was built for: an empty or skipped startup.

**6. Unconditional Ready. after a failed startup.**

The only signal is one echo line before Ready., which RoSys does not parse. A robot can come up with motors and buses but without the safety rules defined after the failing line. Same as the existing #246 behaviour for runtime_error, which this PR extends to every exception. RoSys is currently protected only because core.output is the second to last generated line. The proposal above replaces the partial state with a skipped startup and a stored reason, which is the "explicit fail-safe" option from #246.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20an%20automated%20code%20review%20checked%20against%20the%20code%20by%20the%20model%3B%20the%20safe-mode%20proposal%20and%20the%20decision%20to%20hand%20the%20PR%20back%20are%20Falko%27s%2C%20wording%20and%20structure%20from%20the%20model.&d=2026-09-22&a=claude-code "claude-fable-5-1: Findings from an automated code review checked against the code by the model; the safe-mode proposal and the decision to hand the PR back are Falko's, wording and structure from the model.")

JensOgorek:
### Boot counter and restart-on-failure implemented as proposed, verified on a bench ESP32 and on the F30 core

@falkoschindler thanks, I agree with the direction and built it: the widened catch no longer continues with half a startup, a boot counter in RTC memory restarts after a failed startup and skips the script after two failed boots, and the four hardenings are in as well. Verified on a bench ESP32 and on the F30 core (rb68): a failing module in the middle of a 133-line startup gives two restarts, then `error: startup skipped after 2 failed boots (imu setup failed: …)`, `Ready.` and a working console; a panic during the startup ends the same way with `(panic)`. This decides #246 as explicit fail-safe: a failing required module now means no modules at all instead of the ones before the failing line. For RoSys robots that is what effectively happened already, since `core.output` is one of the last lines and RoSys never became ready.

Two deviations from your sketch, both deliberate: the counter is cleared by reset reason (`ESP_RST_POWERON`, `ESP_RST_EXT`) rather than by "RTC memory is cleared on power cycle", because `RTC_NOINIT` memory is undefined after power-on, not zero. And a deliberate `esp_restart()` never counts as a failure, only a startup exception, a panic, a watchdog or a brownout while the guard is armed (first 5 s of the main loop). Otherwise two quick `core.restart()` in a row, which RoSys does when it configures and restarts, would skip the startup by mistake. The price is that a `core.restart()` inside a startup script still loops; I think that is acceptable.

Not covered: the original #288 heap exhaustion could not be reproduced with this build, because the F30 startup needs the Innotronic modules from lizard-zz and the lizard-zz pin does not build against this branch. Ready for another look; the PR description is rewritten to match.

<details>
<summary>What changed per finding</summary>

- **Boot counter (new `main/utils/boot_guard.{h,cpp}`)**: `RTC_NOINIT_ATTR` magic, counter, armed flag and a 128-byte reason. `should_run_startup()` runs after `Storage::init()`: clears everything on power-on/EN/unknown reset, counts a panic/watchdog/brownout reset if the last boot was still armed, and skips the startup at two failed boots with one error line naming the reason. The main loop disarms and clears the counter after 5 s. An exception while loading the startup goes to `startup_failed()`: reason stored, `error while loading startup script: …` and `restarting (failed boot N of 2)` echoed, `uart_wait_tx_done`, `esp_restart()`.
- **Finding 1, emergency pool**: `CONFIG_COMPILER_CXX_EXCEPTIONS_EMG_POOL_SIZE=512` in both sdkconfig defaults.
- **Finding 2**: the `process_uart` catch takes `std::exception`; `run_step`, rules and routines stay at `std::runtime_error` as you suggested.
- **Finding 3**: `~SerialBus` unregisters the echo callback (`register_echo_callback` now returns a handle, `unregister_echo_callback(handle)` added), asks the communication task to stop via an atomic flag, waits up to 100 ms for it to exit on its own, deletes it otherwise, then frees both queues. The loop is no longer `[[noreturn]]`.
- **Finding 4**: `Storage::read()` throws a `runtime_error` when `malloc` returns NULL.
- **Finding 5**: obsolete with restart-on-failure, `restore_if_needed()` now only ever sees an empty or skipped startup.
- **Finding 6**: `Ready.` is still printed after a skipped startup, but no `core.output` runs, so RoSys stays "not ready", and the skip line is the only text line before `Ready.`.
- **Docs**: a paragraph in "Getting started → Startup Script" describes the restart, the skip after two failed boots, the recovery over the console and what clears the counter.

</details>

<details>
<summary>Bench ESP32 (classic, console 115200)</summary>

- Good startup (`before = Output(4)`, `after = Output(5)`) boots to `Ready.` with both modules; two `core.restart()` 3 s apart do not count.
- `imu = Imu()` inserted between the two lines with no chip attached: `error while loading startup script: imu setup failed: I2CError: Check your wiring.`, `restarting (failed boot 1 of 2)`, same again with `2 of 2`, then `error: startup skipped after 2 failed boots (imu setup failed: …)` and `Ready.`. `core.version()` answers, `before.on()` and `after.on()` report unknown modules, `!?` prints the stored script. `!-imu`, `!.`, `core.restart()` boots normally with both outputs.
- Panic path, with a temporary `core.__crash__()` (null store) in a local build only: two `Guru Meditation Error … StoreProhibited` boots, then `error: startup skipped after 2 failed boots (panic)` and a working console. Repaired startup loads. A panic 7 s after boot, twice, does not count and the startup loads on every boot.

</details>

<details>
<summary>F30 core (rb68, ESP32 with Bluetooth, four plexus peers on the bus, console 460800, RoSys stopped)</summary>

- The stored RoSys startup (checksum 503c) needs `InnotronicDriveMotor` from lizard-zz, so on this build it fails at line 6: two restarts with `unknown module type "InnotronicDriveMotor"`, then the skip line and `Ready.`, console reachable, `!?` prints the full 106-line script. That is the boot-loop scenario of #288 with a real production startup, only with a different exception.
- An Innotronic-free variant of the 154-line gripper startup (132 lines: Bluetooth, CAN, `p0` expander, bus with `make_coordinator(2, 3, 4, 5)`, the same number of `core.output` fields) boots to `Ready.` with about 10 core lines per second and 41 KB free heap.
- The same variant with `imu = Imu(0, 21, 22, 0x29)` at line 7 (wrong address, the board's BNO055 is at 0x28): two restarts, skip with `imu setup failed`, `core.version()` answers, `rdyp.on()` gives `unknown module "rdyp"`, `!?` lists all 133 lines, 129 KB free heap. Writing the loading variant back and restarting gives `Ready.` with core lines again.
- Two `core.restart()` 3 s apart: startup loads, 101 core lines in 10 s.
- Restored afterwards: RoSys startup 503c, previous lizard-zz firmware, RoSys reconnected with matching checksum.

</details>

Implemented and tested with Claude Fable 5.1.

falkoschindler (approved):
@JensOgorek two of us read the branch independently and merged the findings. Everything from the first round is addressed, the design and the hardware runs look right. Three things came up in the new code, all small and unambiguous, so @falkoschindler and I fixed them on the branch instead of another round trip: commit 4eb7ac5. The remaining catches in the main loop stay narrow on purpose, that PR is about startup failures and a wider catch would be a separate decision.

**1. A freshly flashed or NVS-erased core restarted forever instead of reaching Ready.** Storage::init() ran inside the guarded try. It throws on every boot while NVS has no startup yet: Storage::get() reads num_chunks and read() turns ESP_ERR_NVS_NOT_FOUND into "could not peek storage storage.num_chunks", Storage::put() even had a catch for exactly that case. Before this PR the error was echoed and the core reached Ready., which is why `espresso.py erase` was the documented way out of #288. On the branch every boot went to startup_failed() and restarted, should_run_startup() was never reached, and on the first power-on the RTC counter was still uninitialised at that point, so the line read something like "restarting (failed boot 3054 of 2)". Fix: Storage::init() has its own try that echoes and continues, only process_lizard() is guarded, and Storage::get()/put() treat a missing num_chunks as "no startup yet".

**2. The SerialBus destructor did not run if the constructor itself threw.** The echo callback was registered after the communication task existed. emplace_back and the std::function copy can throw bad_alloc under exactly the memory pressure of #288, and C++ does not run the destructor of an object whose constructor did not finish, so the task kept dereferencing the destroyed bus until the restart. Fix: everything that can throw now comes before the task is created, and the task-creation failure path unregisters the callback. The comment above the destructor also described a case that does not exist, a bus in Global::modules is never torn down by a later failing line; the real window is Global::add_module() throwing.

**3. The destructor's forced vTaskDelete() raced the task's self-delete.** After the 100 ms wait, "if (!stopped) vTaskDelete(task)" could run right after the task set stopped and deleted itself, deleting the same TCB twice. Fix: the task parks itself in vTaskSuspend() once it has left its loop, and the destructor is the only place that deletes it.

Approving with one ask before merging: boot one core after `espresso.py erase` once. That path is the reason for the fix, and nobody has run it on hardware yet, neither of us had a board at hand. Expected: no error line, `Ready.`, `!+`/`!.` and `core.restart()` load the new script. The PR description could also mention that a storage error is now reported and the core continues with an empty script.

Checked and fine from our side: the boot counter semantics (no double count on the ESP_RST_SW boot after a startup exception, magic check for undefined RTC memory, counter reset on skip), clearing by reset reason, not counting a deliberate restart, the reset-reason list (checked against the IDF 5.3.1 esp32 and esp32s3 ports, nothing else can report a crash), the 5 s settle window, the emergency pool, the console catch, the Storage::read() NULL check and the docs. Built for esp32 and esp32s3.

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Findings%20from%20two%20model%20reviews%20of%20the%20PR%20head%2C%20fixes%20implemented%20by%20the%20model%20on%20the%20branch%3B%20the%20decision%20to%20keep%20the%20main-loop%20catches%20narrow%20is%20Falko%27s%3B%20wording%20from%20the%20model.&d=2026-09-22&a=claude-code "claude-fable-5-1: Findings from two model reviews of the PR head, fixes implemented by the model on the branch; the decision to keep the main-loop catches narrow is Falko's; wording from the model.")

JensOgorek:
### Boot after a full erase passed on a bench ESP32: `Ready.` with a blank NVS, no error line, `!+`/`!.` + `core.restart()` load the new script

@falkoschindler thanks for the fixes. I ran the one open check on a bench ESP32 (classic, the core of the bench pair) with your head `4eb7ac5`: full `erase_flash`, then a complete flash of bootloader, partition table, otadata and app. The first boot reaches `Ready.` after about 2 s with no `error while loading startup script`, no `restarting (failed boot …)` and no storage line, so the restart loop from a fresh or erased NVS is gone. Writing a startup with `!+`/`!.` and `core.restart()` loads it, and a second cold reset via EN boots the same script cleanly. I also added the sentence about the reported storage error to the PR body. Merging now, #295 follows after this one as agreed.

<details><summary>Console log of the first boot with blank NVS and the restart with the new script</summary>

```
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
I (657) app_init: App version:      v0.14.1-10-g4eb7ac5
I (635) boot: Set actual ota_seq=1 in otadata[0]      <- fresh otadata, nothing left from before
I (2279) uart: queue free spaces: 20

Ready.
> core.info()
version: v0.14.1-10-g4eb7ac5
> !+before = Output(4)
> !+after = Output(5)
> !.
> !?
before = Output(4)
after = Output(5)
> core.restart()
rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
I (2242) gpio: GPIO[4]| InputEn: 0| OutputEn: 0| ...
I (2245) gpio: GPIO[5]| InputEn: 0| OutputEn: 0| ...

Ready.
> before.on()
> core.print(before.level, after.level)
1 0
```

Second cold reset (RTS/EN pulse): `Ready.`, `core.print(before.level, after.level)` → `0 0`, no error line.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Erase%20and%20boot%20test%20run%20by%20the%20model%20on%20the%20bench%20ESP32%3B%20wording%20from%20the%20model.&d=2026-09-23&a=claude-code "claude-fable-5-1: Erase and boot test run by the model on the bench ESP32; wording from the model.")

PR #293, opened 2026-09-11
Refuse Expander.disconnect() and flash() while another module uses the Serial

### LINKED ISSUE #259: SerialBus communication task spins forever on a deleted serial driver

JensOgorek:
### Motivation

Deleting the UART driver of a `Serial` that a `SerialBus` still uses (`Expander.disconnect()` or `Expander.flash()` on the same `Serial`) left the bus communication task spinning on the dead driver and flooding the log with `uart driver error` lines until reboot (#259). The first version of this PR silenced the flood with `uart_is_driver_installed()` guards in every `Serial` read and write. The review (below) argued that this is an error state that should be refused once, not a runtime state to be tolerated silently, and this PR now does that.

### Implementation

`Serial` keeps the names of the modules that use it: `Expander`, `SerialBus` and `RoboClaw` call `serial->claim(name)` in their constructors. `Expander.disconnect()` and `Expander.flash()` call `serial->require_sole_user(name)` before doing anything, which throws `serial "s" is in use by "bus"` if another module shares the `Serial`. The invalid state can no longer be reached, so the `uart_is_driver_installed()` short-circuits in `Serial` are gone, including the older ones in `available()` and `has_buffered_lines()` from #261. A `Serial` that was disconnected on purpose reports the IDF error line once per command again, as before #261.

#259 also asks for an explicit `SerialBus` teardown that stops the communication task. That remains open and is a separate change; this PR only makes sure the driver cannot be deleted underneath a running bus.

Verified on a bench pair of two ESP32 (core with `s = Serial(26, 27, 115200, 1)`, target flashed as expander replica). With `bus = SerialBus(s, 1)`, `bus.make_coordinator(2)` and `p = Expander(s, 25, 14)` on the same `Serial`, both `p.disconnect()` and `p.flash()` are refused with the error above, the bus keeps polling and no `uart driver error` line appears. With the expander alone, `p.disconnect()` works as before; a subsequent `p.run(...)` or `s.read()` prints the IDF error line without an exception and without a flood. Binary size on esp32 grows by 704 bytes.

### Progress

### DIFF
--- a/main/modules/serial.cpp
+++ b/main/modules/serial.cpp
@@ -80,6 +80,9 @@ void Serial::reinitialize_after_flash() const {
 }
 
 size_t Serial::write(const uint8_t byte) const {
+    if (!uart_is_driver_installed(this->uart_num)) {
+        return 0;
+    }
     const char send = byte;
     uart_write_bytes(this->uart_num, &send, 1);
     return 1;
@@ -90,6 +93,9 @@ void Serial::write_checked_line(const char *message) const {
 }
 
 void Serial::write_checked_line(const char *message, const int length) const {
+    if (!uart_is_driver_installed(this->uart_num)) {
+        return;
+    }
     static char checksum_buffer[16];
     uint8_t checksum = 0;
     int start = 0;
@@ -120,16 +126,25 @@ bool Serial::has_buffered_lines() const {
 }
 
 void Serial::flush() const {
+    if (!uart_is_driver_installed(this->uart_num)) {
+        return;
+    }
     uart_flush(this->uart_num);
 }
 
 int Serial::read(uint32_t timeout) const {
+    if (!uart_is_driver_installed(this->uart_num)) {
+        return -1;
+    }
     uint8_t data = 0;
     const int length = uart_read_bytes(this->uart_num, &data, 1, timeout);
     return length > 0 ? data : -1;
 }
 
 int Serial::read_line(char *buffer, size_t buffer_len) const {
+    if (!uart_is_driver_installed(this->uart_num)) {
+        return 0;
+    }
     int pos = uart_pattern_pop_pos(this->uart_num);
     if (pos >= static_cast<int>(buffer_len)) {
         if (this->available() <= pos) {


falkoschindler (changes requested):
@JensOgorek the diff does what it says: the five guards mirror the existing `available()` / `has_buffered_lines()` short-circuits, change nothing while the driver is installed, and your bench capture shows the `uart driver error` flood is gone. My problem is with the direction, not the code. I think "driver deleted while someone still uses the Serial" is an error state, not a runtime state we should quietly tolerate, and the silent returns hide it in eight places instead of reporting it once.

## The design question

Only two places delete a live driver: `Expander.disconnect()` (documented as handing the wire over to USB access of the other microcontroller) and `Expander.flash()` (transient, reinstalls afterwards). Serial itself has no DSL method to deinstall. So the state this PR guards against only arises when an Expander and a SerialBus share the same Serial and one of those two methods is called. Two protocols on one wire is not a configuration we run; it is the bench reproduction. That makes it an exception, and I'd rather have it reported as one.

Concretely, I see two consistent options:

- **Refuse the teardown while the Serial is in use.** Serial knows its users (a registration or a counter), `disconnect()` and `flash()` throw "serial s is in use by bus b", and the invalid state cannot be reached at all. The guards from #261 and the older one in `available()` become unnecessary too.
- **Tear the bus down first and throw afterwards.** SerialBus gets the explicit `disconnect()` from #259 that stops the comm task, and the Serial methods throw `std::runtime_error("serial interface is not installed")` when the driver is gone. On the main task those exceptions are caught per module step and echoed, which is exactly the signal a user needs when `s.send()`, `p.run()` or `bus.send()` go nowhere.

What does not work is throwing (or asserting) without either of the above: the coordinator branch of `communication_loop` in main/modules/serial_bus.cpp (146) has no try/catch, so an exception from `write_checked_line` on core 1 aborts the ESP, and `p.disconnect()` with a bus present would become a reliable reboot.

## Consequences for this PR as it stands

1. **"Closes #259" is not accurate.** The coordinator still spins on the dead driver and pushes `warning: serial bus X poll to N timed out` every 250 ms until reboot, which the PR description confirms. The issue's headline is still true; only the log lines changed shape.
2. **The check-then-act race stays.** The comm task on core 1 (main/modules/serial_bus.cpp (56)) can pass a guard just before `uart_write_bytes` while `uart_driver_delete` on the main task frees the ring buffers and `tx_mux`. The window is small and pre-existing, but it is exactly why #259 asks for stopping the task first.
3. **User commands now fail silently.** `Serial::call("send")` at main/modules/serial.cpp (195) ignores the new `write() == 0`, and `expander.run(...)`, proxy calls and `bus.send(...)` all return without echo or exception. The IDF error line was the one signal that a command went nowhere.
4. **`uart_is_driver_installed()` keys on global port state, not on this Serial's driver.** During `Expander.flash()` the loader installs its own driver on the same port after `deinstall()` (main/modules/expander.cpp (202–203)), so a sharing SerialBus passes all guards again and writes poll frames into the bootloader stream. Pre-existing, but another sign that the predicate is the wrong tool.
5. **`Serial::write()` still returns a hard-coded 1** after `uart_write_bytes`, so 0 now looks meaningful but 1 is not derived from the write result. Minor; no caller uses the value.

## Verdict

I'd not merge the guards. My preference is the first option (refuse `disconnect()` / `flash()` while a bus uses the Serial), because it removes the state instead of handling it, and the teardown from #259 then stays a separate, smaller change. If you'd rather keep this PR minimal for now, drop "Closes #259" and add a sentence that concurrent deinstall is still not safe; but I think we'd be extending a pattern (#261) that already went the wrong way.

JensOgorek:
### Reworked to the first option: `disconnect()` and `flash()` now refuse while the `Serial` is shared, all driver guards dropped, bench test 9/9 passed

@falkoschindler thanks, agreed on the direction. The PR now removes the state instead of handling it: `Serial` knows its users, `Expander.disconnect()` and `flash()` throw `serial "s" is in use by "bus"` before touching anything, and the `uart_is_driver_installed()` short-circuits are gone, including the two older ones from #261 and `available()`. The description is rewritten accordingly and no longer claims to close #259; the explicit bus teardown stays a separate change. Ready for another look.

<details>
<summary>What changed since your review</summary>

- `Serial::claim(name)` is called in the constructors of `Expander`, `SerialBus` and `RoboClaw`; `Serial::require_sole_user(name)` throws if any other name is registered.
- The check sits at the top of `disconnect()` and of `flash()` (before the boot pin is pulled low), so a refused call has no side effect.
- All seven `uart_is_driver_installed()` guards in `Serial` are removed. On a deliberately disconnected `Serial`, `p.run(...)` and `s.read()` print the IDF error line again, once per command, and the exception path on the main task is untouched.
- Your points 2, 4 and 5 (check-then-act race, port-global predicate during `flash()`, hard-coded return of `write()`) disappear with the guards; 1 and 3 are addressed by refusing instead of silencing.
- Branch merged with current main (#292, #295); docs got one sentence at `disconnect()`/`flash()`.
</details>

<details>
<summary>Bench test (two ESP32, core + expander replica, this branch on the core)</summary>

Shared `Serial` (`s`, `bus = SerialBus(s, 1)`, `bus.make_coordinator(2)`, `p = Expander(s, 25, 14)`):

| Step | Result |
| --- | --- |
| `p.disconnect()` | `error processing uart0: serial "s" is in use by "bus"` |
| `p.flash()` | same error, startup checksum unchanged |
| bus after both refusals | keeps polling (40 poll timeouts in 10 s) |
| `uart driver error` lines over the whole run | 0 |

Expander alone (`s`, `p = Expander(s, 25, 14)`, `p.is_ready` true):

| Step | Result |
| --- | --- |
| `p.disconnect()` | no error, `p.is_ready` false |
| `p.run("core.info()")` | 2 IDF `uart_write_bytes ... uart driver error` lines, no exception |
| `s.read()` | 2 IDF error lines (`uart_get_buffered_data_len`, `uart_read_bytes`) |
| idle 5 s afterwards | 0 lines |

Not exercised: a full `p.flash()` without a bus; that path only gained the check before the first side effect. Side note from the shared setup: the bus logs `could not parse message` for every expander reply, which confirms that two protocols on one wire is not a configuration to support.
</details>

<details>
<summary>Size</summary>

esp32 `lizard.bin`: main `28ad386` 1 335 376 B, this branch `dfbbae0` 1 336 080 B (+704 B).
</details>

Reworked and bench-tested with Claude Fable 5.1 (`claude-fable-5-1`).

falkoschindler (approved):
@JensOgorek thanks, this is the shape I was hoping for: the invalid state is refused once, before the first side effect, and the seven driver guards are gone instead of multiplying. Bench evidence with both the shared and the standalone setup is exactly what this needed.

I pushed two small commits on top; nothing left for you to change.

- 231a86b: `Serial::available()` now initializes its byte count to zero. `uart_get_buffered_data_len()` returns `ESP_FAIL` without writing `*size` when the driver is gone, so after `p.disconnect()` the method returned an uninitialized stack value and `s.read()` echoed the buffer of the previous read. Your bench table already shows the symptom: `s.read()` produced two IDF error lines, and the second one (`uart_read_bytes`) only appears when `available()` was nonzero. The guard you removed had hidden this since df36b31. The unused `Serial::clear()` went with it.
- 411b91c: `SerialBus` claims the `Serial` at the end of its constructor, after the queue and task creation that can throw. A claim is never released, so a claim before a failing constructor would have left a phantom user that blocks `disconnect()` and `flash()` until reboot.

Not part of this PR, for the record: `is_ready` staying `false` after `flash()` or `restart()` is #266 with #269 open, and the explicit bus teardown from #259 stays a separate change as you wrote.

<details>

**1. `available()` reads an uninitialized value once the driver is deleted** main/modules/serial.cpp (122) — `uart_get_buffered_data_len()` leaves `*size` untouched on `ESP_FAIL`, so `get_output()` saw a random nonzero count, called `read()`, got `-1`, and returned the static buffer with stale content. Fixed in 231a86b.

**2. `SerialBus` claims the `Serial` before the constructor's throw points** main/modules/serial_bus.cpp (59) — a failed `xQueueCreate` or `xTaskCreatePinnedToCore` destroyed the bus but left its name in `users`. Fixed in 411b91c by claiming last. The duplicate-name path does not reach the constructor, main/main.cpp (214) rejects the name first.

**3. `Serial::clear()` has no callers** main/modules/serial.cpp (163) — dead code that would have spun on the garbage count from finding 1. Removed in 231a86b.

</details>

This is an AI-assisted review pass; I did not verify every point myself. [※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Review%20pass%20and%20both%20fix%20commits%20by%20the%20model%2C%20decisions%20and%20wording%20direction%20from%20the%20reviewer.&a=claude-code "claude-fable-5-1: Review pass and both fix commits by the model, decisions and wording direction from the reviewer.")

PR #295, opened 2026-09-11
Blank the expander's NVS while replicating instead of copying the core's

### LINKED ISSUE #254: p0.flash() silently replaces the expander's NVS with a copy of the core's

JensOgorek:
### Motivation

Closes #254.

`flashReplica` copies the core's physical flash from address 0, so every `p0.flash()` also copied the core's NVS onto the expander, including `uart/baudrate` and any `bus_backup`. `Storage::clear_nvs()` only covered the startup string. On a core with a non-default console baud rate the expander came back on that rate and the core could not talk to it anymore.

### Implementation

The replica keeps the full image but writes 0xFF over the blocks of the NVS partition (offset and size from the core's own partition table), so the expander boots with a blank NVS: empty startup, default baud rate, no bus backup. The core configures the expander dynamically at boot anyway, so a persistent expander startup is not needed. The `clear_nvs()`/`save_startup()` dance in `Expander::flash()` and the now unused `Storage::clear_nvs()` are removed. Factory-fresh expanders still work because bootloader and partition table are copied as before.

Verified on a bench pair of two ESP32 dev boards wired like the Robot Brain (`Serial(26, 27, 115200, 1)`, `Expander(serial, 25, 14)`), reading the target's own console in parallel. With the core's NVS set to 460800: on `main` the flashed target boots at 460800 and the core never gets `Ready.` at 115200; with this branch it boots at 115200 with `ESP_ERR_NVS_NOT_FOUND` for the startup, the expander becomes ready and `p0.info()` reports the replica version. A second flash from a core with erased NVS behaves the same. Replica takes 48 s, MD5 verified.

### Progress

### DIFF
--- a/main/modules/expander.cpp
+++ b/main/modules/expander.cpp
@@ -183,7 +183,6 @@ void Expander::call(const std::string method_name, const std::vector<ConstExpres
         if (this->boot_pin == GPIO_NUM_NC || this->enable_pin == GPIO_NUM_NC) {
             throw std::runtime_error("expander \"" + this->name + "\" does not support flashing, pins not set");
         }
-        Storage::clear_nvs();
         gpio_set_level(this->boot_pin, 0);
         if (!force) {
             char command[32];
@@ -206,7 +205,6 @@ void Expander::call(const std::string method_name, const std::vector<ConstExpres
                                                     this->serial->rx_pin,
                                                     this->serial->tx_pin,
                                                     this->serial->baud_rate);
-        Storage::save_startup();
         delay(100);
         this->serial->reinitialize_after_flash();
         if (!success) {
--- a/main/storage.cpp
+++ b/main/storage.cpp
@@ -147,10 +147,6 @@ void Storage::save_startup() {
     Storage::put(Storage::startup);
 }
 
-void Storage::clear_nvs() {
-    Storage::put("");
-}
-
 void Storage::set_user_pin(const std::uint32_t pin) {
     write_u32("ble_pins", "user_pin", pin);
 }
--- a/main/storage.h
+++ b/main/storage.h
@@ -17,7 +17,6 @@ public:
     static void remove_from_startup(const std::string substring = "");
     static void print_startup(const std::string substring = "");
     static void save_startup();
-    static void clear_nvs();
 
     static void set_user_pin(const std::uint32_t pin);
     static bool get_user_pin(std::uint32_t &pin);
--- a/main/utils/serial-replicator.cpp
+++ b/main/utils/serial-replicator.cpp
@@ -141,13 +141,22 @@ static auto flash(uint32_t usedSize, uint32_t transferBlockSize) -> bool {
     auto bytePtr{reinterpret_cast<const std::byte *>(ptr)};
     uint32_t toSend = usedSize;
 
+    // the target gets a blank NVS instead of a copy of ours, so it boots with an empty startup and default settings
+    const esp_partition_t *nvs = esp_partition_find_first(ESP_PARTITION_TYPE_DATA, ESP_PARTITION_SUBTYPE_DATA_NVS, nullptr);
+    static std::vector<std::byte> blank;
+    if (nvs != nullptr) {
+        blank.assign(transferBlockSize, std::byte{0xFF});
+    }
+
     /* Send all non-partial block */
     int count = 0;
     while (toSend >= transferBlockSize) {
         if ((count++) % 10 == 0) {
             ESP_LOGI(TAG, "%lu/%lu kb", (usedSize - toSend) / 1000, usedSize / 1000);
         }
-        status = esp_loader_flash_write(const_cast<std::byte *>(bytePtr), transferBlockSize);
+        const uint32_t offset = usedSize - toSend;
+        const bool in_nvs = nvs != nullptr && offset >= nvs->address && offset < nvs->address + nvs->size;
+        status = esp_loader_flash_write(in_nvs ? blank.data() : const_cast<std::byte *>(bytePtr), transferBlockSize);
         ESP_LOGD(TAG, "esp_loader_flash_write(0x%08lX)", usedSize - toSend);
 
         HANDLE_ERROR(status, "writing target flash");


JensOgorek:
### Bench re-test of `1d8fb72` passed: expander boots with a blank NVS at 115200 while the core's NVS holds 460800

Same bench pair as in the description (two classic ESP32 wired like the Robot Brain, `Serial(26, 27, 115200, 1)` + `Expander(serial, 25, 14)`), now with the final commit that frees the blank block after the flash. With the core's console persisted at 460800, `p0.flash(true)` took 46 s with MD5 verified, and the target's own console showed it booting at 115200 with `ESP_ERR_NVS_NOT_FOUND` for the startup. After a core restart the expander reports `Ready.` and `p0.info()` returns this branch's version. Ready for review.

<details><summary>Test steps and observations</summary>

- Core flashed with this branch (`v0.14.1-1-gaee82c4-dirty`, esp32), `core.set_baudrate(460800)` persisted, console reopened at 460800.
- Target console (`/dev/ttyUSB1`, 115200, read-only) opened before the flash, since opening it resets the board.
- Before the flash the old replica on the target was ready at 115200 as expected (its NVS copy was empty).
- `p0.flash(true)`: `waiting for download` on the target, 46 s later `Replica complete`, target reboots with `error while loading startup script: could not peek storage storage.num_chunks (ESP_ERR_NVS_NOT_FOUND)` followed by `Ready.`.
- Directly after the flash `p0.is_ready` stays false; that is the existing behaviour tracked in #269 and unrelated to this change. After a core restart: `p0: Ready.`, `is_ready` true, `p0.info()` shows `v0.14.1-1-gaee82c4-dirty`.

</details>

Reviewed and tested with Claude Fable 5.1.

falkoschindler (approved):
@JensOgorek the approach is right: blanking the NVS blocks inside the replica is simpler and more complete than the old `clear_nvs()`/`save_startup()` dance, and it fixes #254 at the root, since baud rate and bus backup are no longer inherited either. I checked the diff against the partition table and the callers.

I pushed one commit with the small things so we do not need a review round-trip; none of it changes the flashed bytes, so your bench result stands:

- module_reference.md (1127): one sentence that `flash()` erases the expander's NVS (startup, persisted console baud rate, bus backup). Users who persisted a baud rate on the expander would otherwise lose it without a hint in the reference.
- serial-replicator.h (8): the header comment still promised a clone "up until the end of the last partition". It now describes the running-partition-size cutoff and the blank NVS, so nobody builds on the expander inheriting the core's `uart/baudrate` again.
- serial-replicator.cpp (138): the blank block is allocated unconditionally (4 KB for 48 s, no empty-vector reasoning), a warning is logged when no NVS partition is found, and the flash is refused before the target is erased when `nvs->address` or `nvs->size` is not a multiple of the block size. With the default 0x1000 this never triggers; with a caller passing e.g. 0x4000, `in_nvs` would copy the first 12 KB of the core's NVS verbatim and blank otadata instead.

Two points that need a decision rather than code:

**Merge order with #292.** With a blank NVS, `Storage::get()` in storage.cpp (112) throws on the missing `num_chunks` key, so every boot of a freshly flashed expander prints `p0: error while loading startup script: could not peek storage storage.num_chunks (ESP_ERR_NVS_NOT_FOUND)` through the expander echo. Your bench log in the PR body shows exactly this line. The old `clear_nvs()` wrote `num_chunks=0`, so the expander booted quietly. #292 treats a missing `num_chunks` as "no startup yet" and makes the line disappear. I would merge #292 first; if #295 goes first, please note it in the PR body so nobody hunts the message.

**OTA slot, pre-existing.** serial-replicator.cpp (228) flashes bytes 0 to `running_partition->size`, i.e. 0x1f0000: bootloader, partition table, NVS, otadata, phy_init and the ota_0 slot, regardless of which slot the core runs from. After an OTB update the core runs from ota_1. `p0.flash()` then copies the stale ota_0 image plus an otadata that says "boot ota_1"; the expander's ota_1 is never written and `esp_loader_flash_start` only erased 0 to 0x1f0000, so the expander boots whatever was in its ota_1 before, or falls back to the stale ota_0, and `p0.info()` reports a version that differs from the core. Not introduced here and it needs a hardware test with an OTB-updated core, so I would keep it out of this PR: blank otadata alongside the NVS and read the ota_0 blocks from the running partition. I will open an issue and a follow-up PR once this is merged.

Approving with the pushed commit.

PR #297, opened 2026-09-16
Read the BNO055 in its own task instead of the main loop

### LINKED ISSUE #296: BNO055 reads block the main loop for 26 ms — nine blocks read per step, two consumed
## Summary

`Imu::step()` reads the BNO055 synchronously from the main loop. With the default `data_select = 0xffff` that is **nine blocks = 18 blocking I²C transactions per loop pass**, measured at **26 ms** — on a core whose whole loop takes 60 ms. Consumers of the module typically use two of those nine blocks.

Three quarters of the 26 ms are not bus time. At 100 kHz the 18 transactions need ~6.6 ms on the wire; the rest is fixed cost per `i2c_master_cmd_begin`, so raising the I²C clock is the weakest of the available levers.

This is the sibling of #213: same disease (blocking I/O in the main loop), different device. #216 fixed the loop's sleep and #218 the baud rate, but both blocking calls are still in the loop — UART0 is installed with `tx_buffer_size = 0`, and the IMU read is synchronous.

## Measurement

ESP32 Robot Brain core, BNO055 on I²C at 100 kHz (module default), host link at 115200. `imu.data_select` switched at runtime, two minutes per window (one for the last). Cycle time taken from Lizard's own timestamp in the `core <millis> …` line, so no host-side log latency enters the numbers.

| `data_select` | blocks read | n | median | p5 | p95 | min | max |
|---|---|---|---|---|---|---|---|
| 65535 (default) | 9 | 2003 | 60 ms | 55 | 65 | 50 | 74 |
| 33 (`cal` + quaternion) | 2 | 2929 | **41 ms** | 36 | 46 | 30 | 57 |
| 0 (no I²C at all) | 0 | 3486 | **34 ms** | 31 | 39 | 29 | 83 |
| 65535 (control) | 9 | 997 | 60 ms | 55 | 66 | 51 | 75 |

The status line stayed **262 characters in every window** (min = max = 262 over 9 419 lines), so the console write is a constant across all four rows and the deltas are purely the I²C reads. `data_select` only gates the reads in `Imu::step()`; the field list in `core.output` is fixed at startup, and a gated block simply keeps printing its last value.

**The IMU costs 26 ms of the 60.** Nineteen of those are blocks nobody reads.

## Where the time goes

Bit time at 100 kHz, 9 bits per byte including ACK. Each `BNO055::i2c_readLen` is two transactions: a 2-byte register write terminated with STOP, then a fresh START for the read — no repeated START.

| block | payload | bus bytes | bit time |
|---|---|---|---|
| `cal`, `temp` | 1 B each | 2 × (2 + 2) | 0.72 ms |
| 6 vectors | 6 B each | 6 × (2 + 7) | 4.86 ms |
| quaternion | 8 B | 2 + 9 | 0.99 ms |
| **total** | | **18 transactions** | **6.6 ms** |

Measured 26 ms against 6.6 ms of bus time leaves **~19 ms**, i.e. ~1.1 ms per `i2c_master_cmd_begin`. The `data_select = 33` window gives the same answer independently: 1.35 ms of bus time against 7 ms measured, ~1.4 ms per transaction over four transactions.

So the cost scales with the **number of transactions**, not with the number of bytes.

## Options

1. **Move the read off the main loop into its own task.** The only option that removes the cost entirely and keeps all nine blocks: the loop drops to the measured 34 ms floor regardless of what the bus does. `SerialBus` already establishes the pattern with `xTaskCreatePinnedToCore`. Two constraints: the task must not hold the `InterpreterLock` across the I²C read — otherwise the main loop blocks on the lock instead of the bus and nothing is won — so read lock-free into a local buffer and take the lock only to publish the properties. And the age of the values in the status line becomes variable, which matters for consumers that timestamp the quaternion by the line's arrival.
2. **One contiguous read.** All nine blocks live in `0x08`–`0x35`, 46 contiguous bytes. A single `readLen` over the range is 2 transactions instead of 18 and ~4.4 ms of bus time, with no change to what the module exposes.
3. **A better default for `data_select`**, or letting consumers set it — the workaround that needs no firmware change, worth 19 ms today.
4. **400 kHz I²C** (already reachable via the 5th constructor argument) attacks only the 6.6 ms and can therefore save ~5 ms of the 26 at best. Listing it so nobody reaches for it first.

1 and 2 are independent and compose.

## What was not measured

How the ~19 ms of non-bus time splits between the BNO055's clock stretching and ESP-IDF driver/task latency. That split decides how much option 2 alone would buy; option 1 does not depend on it.

[※](https://zauberzeug.github.io/colophon/#m=claude-opus-5&t=Measurements%20taken%20on%20hardware%20and%20written%20up%20by%20the%20model%3B%20the%20decision%20to%20cross-check%20a%20second%20robot%20and%20to%20file%20this%20came%20from%20the%20author.&a=claude-code "claude-opus-5: Measurements taken on hardware and written up by the model; the decision to cross-check a second robot and to file this came from the author.")

Johannes-Thiel:
### Motivation

Closes #296 (option 1). `Imu::step()` called the synchronous ESP-IDF I2C API straight from the single main loop, so every BNO055 read was loop time. With the default `data_select = 0xffff` that is nine blocks, 18 transactions, **26 ms per pass** — measured on an ESP32 Robot Brain core at 100 kHz, cycle time taken from Lizard's own timestamp in the `core` line, two minutes per window:

| `data_select` | blocks read | n | median | p5 | p95 |
|---|---|---|---|---|---|
| 65535 (default) | 9 | 2003 | 60 ms | 55 | 65 |
| 33 | 2 | 2929 | 41 ms | 36 | 46 |
| 0 (no I²C) | 0 | 3486 | 34 ms | 31 | 39 |

Only ~6.6 ms of those 26 are bus time, so the cost is fixed overhead per `i2c_master_cmd_begin`, not bandwidth. That rules out the two obvious levers: a faster I²C clock attacks only the 6.6 ms, and reading fewer blocks still pays the per-transaction cost for the ones that remain (`data_select = 33` still costs 7 ms). Taking the read off the loop removes all 26 ms and keeps every block.

This is the same disease as #213, which #216 and #218 addressed for the console write. The IMU was the other blocking call in the loop.

### Implementation

- Add an `imu_read` task (core 1, priority 5, 4096 stack — the values `SerialBus` already uses) that fills a one-element queue with `xQueueOverwrite`, so the producer never blocks and the loop always gets the newest sample.
- Drain the queue non-blockingly in `step()` and write the properties there, under the `InterpreterLock` the main loop already holds. The task never touches properties: holding that lock across the read would move the stall from the bus to the lock and win nothing.
- Carry `data_select` in the sample so `publish()` gates exactly as before — a deselected block keeps its last value, unchanged semantics.
- Add a `bno_mutex` serializing device access against `set_mode`, which shares the BNO055's page state. No deadlock: `call()` holds the `InterpreterLock` and waits for `bno_mutex`, the task never takes the `InterpreterLock`.
- Pace the task with `xTaskDelayUntil` at 10 ms, the same drift-free cadence as the main loop since #216 and the BNO055's fusion output rate.
- Raise read errors from `step()` rather than echoing from the task, since `echo()` formats through one `static` buffer. While reads fail, the error repeats at most once per second, and a message is printed when reads work again. This is a deliberate change: on `main` a hard read failure is not caught by `run_step` (`BNO055I2CError` derives from `std::exception`, not `std::runtime_error`) and terminates the firmware; with this PR the system keeps running and the properties keep their last values.
- Back off 1 s after a read that fails all 64 retry rounds, and silence the driver's `BNO055` log tag: it prints one unchecksummed line per failed retry round straight to UART0, even for reads that eventually succeed.
- Stop the task cooperatively in `~Imu()` (stop flag, notification to cut the back-off short, confirmation semaphore) instead of `vTaskDelete` from outside, which could catch the task holding the heap lock.
- `imu.set_mode(...)` waits on `bno_mutex` for up to one read pass (~26 ms) before its own 30 ms `setOprMode` delay. Only on mode changes.

### Measured on hardware

Flashed to a Robot Brain core and measured, two minutes per window in one session: **64 ms unpatched, 40 ms with this branch**, all nine blocks still read and the IMU values still updating. Flashing the unpatched build back reproduced 62–67 ms. Numbers and caveats in the comment below.

### Follow-up

`ImuBno085` stays on the main loop for now, as agreed in review; whether it gets its own task is a follow-up question.

Also fixed on the branch (13b562b): `temp` is declared `IntegerVariable` but was assigned through `number_value`, and `print_to_buffer` switches on the type — so `imu.temp` always printed 0.

---
⬛ claude-opus-5 (extended thinking)


Johannes-Thiel:
**Measured on hardware.** Both windows two minutes on the same robot in one session, cycle time from Lizard's own timestamp in the `core` line:

| firmware | n | median | p5 | p95 | min | max |
|---|---|---|---|---|---|---|
| unpatched | 1880 | **64 ms** | 58 | 69 | 51 | 136 |
| this branch | 2844 | **40 ms** | 35 | 47 | 29 | 125 |

**24 ms off the loop**, with all nine blocks still read — roll, pitch, yaw and the gyro calibration keep updating in the UI throughout. That matches the 26 ms the same robot showed when the reads were switched off entirely with `data_select = 0`, so the task removes essentially all of the blocking time rather than trading data for it.

Flashing the unpatched build back afterwards reproduced 62–67 ms, so the difference is the patch and not drift.

One caveat on transferability: the robot runs a downstream fork of this firmware, so the absolute numbers include modules that are not in this repo. The delta is what carries over.

evnchn:
I'd land option 2 from #296 first and rebase this on top:

- **Payoff.** One `readLen` over `0x08–0x35` — the whole page-0 data area, all nine blocks, 46 bytes per the enum in `BNO055ESP32.h` — is 2 transactions instead of 18: ~6.6 ms instead of 26 ms, i.e. ~19 of the 26 ms the issue measured.
- **Cost.** A small addition to `BNO055ESP32.cpp`, versus a task, a mutex and changed error semantics here.
- **They compose.** Option 1 still wins the remaining 6.6 ms, so it is the better end state — but it is the larger surface, so landing it second gets ~3/4 of the win without the concurrency.
- **It also fixes this PR's premise.** `READ_PERIOD = 10 ms` is unreachable at 26 ms per read: the task overruns every iteration and free-runs at roughly the read time. At ~6.6 ms the deadline becomes real.

The concurrency looks right — queue of one + `xQueueOverwrite`, no `InterpreterLock` from the task, `bno_mutex` only against `set_mode`. And the parse belongs in `BNO055` next to `getVector`: the scales are `protected` and `setUnits` rewrites them.

Happy to flash it on our robot and put numbers next to yours. [※](https://zauberzeug.github.io/colophon/#m=deepseek-v4.1-flash&t=Register%20map%2C%20library%20API%20and%20the%20transaction%20and%20timing%20arithmetic%20from%20the%20model%3B%20the%20decision%20to%20raise%20option%202%20on%20this%20PR%2C%20and%20to%20post%20it%2C%20came%20from%20the%20author.&a=opencode "deepseek-v4.1-flash: Register map, library API and the transaction and timing arithmetic from the model; the decision to raise option 2 on this PR, and to post it, came from the author.")

Johannes-Thiel:
The `READ_PERIOD` point is a straight hit and is now fixed in d6df2a1 — the comment claimed a drift-free 100 Hz cadence that never happens, since a 26 ms read overruns the 10 ms period every iteration and the task free-runs. Thanks. Agreed too that the parse belongs in `BNO055` next to `getVector`, for the reason you give: `accelScale` and friends are `protected` and `setUnits` rewrites them.

On the ordering I'd rather not, and I think the case for option 2 is weaker than it looks once option 1 is in.

Loop time is the whole motivation in #296, and option 1 takes all of it: measured 64 → 40 ms on hardware, with all nine blocks still read. After that, a cheaper read buys the loop nothing — the reads are no longer on it. That also makes the "it fixes this PR's premise" argument circular: the 10 ms deadline only needs to be reachable if we want that cadence, which is the sample-age argument below, not a loop-time one.

What option 2 would still buy, honestly enumerated:

- **Bus occupancy.** The task reads continuously, so the I²C port is busy ~100 % of the time. That only hurts if something shares the port — `mcp23017` and `imu_bno085` can. On our robots nothing does: one BNO055, no MCP23017. Dead argument for us; may not be for others.
- **Sample age.** The quaternion is ~27 ms old with the task against ~64 ms today, and would be ~10 ms with option 2. This is the one real remainder — but the task already improves it 2.4×, and nobody has shown that the remaining 27 ms costs anything downstream, including in the Kalman-based locator that weights the IMU at 90 % of the angular rate.

Against that, option 2 forks `BNO055ESP32.cpp`, which is third-party MIT code. Option 1 touches only Lizard's own module. So the surface comparison goes the other way from how it reads: a task and a mutex in our own file, versus a divergence in a vendored driver.

And the ~6.6 ms rests on the one thing #296 flags as unmeasured — that the ~19 ms of non-bus time is per transaction rather than per byte. If the BNO055 stretches the clock per byte, a 46-byte burst saves much less. Option 1's number does not depend on that assumption at all.

So: option 2 as a follow-up question rather than a prerequisite, motivated by sample age, and only after that assumption is checked.

JensOgorek (changes requested):
### Loop win confirmed on our F30 (IMU cost 18 ms → 0), but the failure path needs work before merge: on a bad bus the task retries at full rate and floods the console

The task/queue/mutex structure looks right and I have nothing to add to what @evnchn already said about it. On our F30 the PR removes the whole IMU cost from the loop (main: 20 ms without IMU, 38 ms with all nine blocks; this branch: 20 ms in both cases, values live), and on a flaky bus the loop stays at 10–20 ms where main blocks for up to 1.5 s. What I would change before merging is what the new code does once the read fails: the retry loop has no backoff and the library floods UART0 with unchecksummed log lines from the task (measured 106 lines/s), the failure is reported once and then the properties silently keep stale values, and the module owns a background task it cannot stop. Details below.

<details>
<summary>1. Failed reads retry at full rate and flood the console (should fix)</summary>

When `read_sample` throws, the `catch` sets `read_failed` and the loop immediately starts the next pass. Every pass goes through `i2c_readLen`, which retries `UART_ROUND_NUM = 64` times and prints `ESP_LOGE(... "(i2c RL1) Error")` on each failed round. Those `ESP_LOGE` calls sit outside the `BNO055_DEBUG_OFF` guard, `CONFIG_LOG_DEFAULT_LEVEL_INFO=y`, and there is no `esp_vfs_dev_uart_use_driver`, so each line busy-waits on the UART0 TX FIFO from a priority-5 task on core 1, the same priority and core as `serial_bus_comm`.

Measured on the F30 core with a deliberately disturbed bus (see the hardware block): 106 library lines per second between Lizard's protocol lines, continuously, while the properties keep updating from the passes that succeed. A NACKing round takes ~1.5 ms (128 lines in 199 ms for the two transactions of a failed `begin()`), so an unplugged sensor would produce roughly 600 lines/s with no pause. A `vTaskDelay(pdMS_TO_TICKS(1000))` in the `catch` block would be enough and keeps the vendored library untouched.

</details>

<details>
<summary>2. Error semantics: reported once, no recovery message, and the "one error per pass" premise is probably not what main does</summary>

`run_step` in `main.cpp` catches only `std::runtime_error`. The BNO055 exceptions derive from `BNO055BaseException : public std::exception`, and the old `Imu::step()` did not catch them, so by my reading a hard read failure on `main` terminates the firmware rather than printing one error per pass. I could not reproduce that on hardware: my fault injection only produced a flaky bus where no read lost all 64 rounds, so neither build threw. If the reading is right, this PR's `catch (const std::exception &)` also turns a reboot into a running system, which the description should say.

The new behaviour then reports the transition once and afterwards keeps broadcasting the last sample indefinitely, with no message when the bus recovers. A host reading `imu.yaw` cannot tell a live sensor from a dead one. I would at least report the recovery as well, or re-report periodically; invalidating the values on failure would be clearer still.

</details>

<details>
<summary>3. Resource hygiene: task handle discarded, no destructor</summary>

`xTaskCreatePinnedToCore` gets `nullptr` for the handle, the task captures raw `this`, and `bno_mutex` and `sample_queue` are never deleted. Today only `Serial` modules are ever erased (`bus_backup.cpp`), so nothing dereferences freed memory yet, but the module now owns a task it cannot stop. Keep the `TaskHandle_t` and add `~Imu()` that deletes task, queue and mutex; `SerialBus` at least keeps its handle.

</details>

<details>
<summary>Measured on F30 (rb68, classic ESP32 core, console 460800)</summary>

Both builds are plain upstream Lizard for esp32 (`50f45e6` = main, `d6df2a1` = this PR), flashed A/B in one session with rosys stopped, `core.output` reduced to `core.millis core.heap`, and the BNO055 added live with `imu = Imu()`. Loop period from `core.millis` deltas over 30 s per phase (median / p90 / max in ms):

| build | no IMU | IMU 65535 (9 blocks) | IMU 33 | IMU 0 | heap without / with IMU |
|---|---|---|---|---|---|
| main | 20 / 20 / 24 | 38 / 40 / 43 | 20 / 21 / 26 | 20 / 20 / 24 | 91.7 / 86.7 KB |
| this PR | 20 / 21 / 24 | 20 / 20 / 23 | 20 / 20 / 24 | 20 / 20 / 24 | 91.7 / 81.9 KB |

Values keep changing between samples (`acc_z`, `gyr_x`), `quat_w` stable, so all nine blocks are really read. The task costs about 4.8 KB heap. The 20 ms floor without IMU is specific to this robot's setup and not investigated; only the delta matters here.

Failure path: no one could unplug the sensor, so I disturbed the bus by adding `imu2 = Imu(1, 21, 22)`, which steals the pins of port 0 through the GPIO matrix. That gives a flaky bus, not a dead sensor, on both builds (mostly `-1`, some `263` timeouts):

| build | loop during the fault | BNO055 log lines on the console | properties |
|---|---|---|---|
| main | 99 / 101 / 1526 ms | 374 in 20 s (19/s, ~50 ms per timed-out round) | – |
| this PR | 10 / 20 / 25 ms | 2117 in 20 s (106/s, ~1 ms per NACKed round) | keep updating |

No `error in module imu` and no abort on either build, because no single read lost all 64 rounds. Driving SDA or SCL low through an `Output` module does not disturb the reads at all, in case anyone tries that route.

</details>

<details>
<summary>Minor, take or leave</summary>

- `imu.set_mode(...)` runs in the main loop under the `InterpreterLock` and waits on `bno_mutex` for up to a full pass (~26 ms) before its own 30 ms `setOprMode` delay. Only on mode changes, but worth a note.
- `MutexGuard` will be needed again for `ImuBno085`; it could live next to `interpreter_lock.h` in `main/utils` instead of an anonymous namespace in `imu.cpp`.
- 4096 bytes of stack with a C++ throw plus 64 `vprintf` calls on the failure path: a `uxTaskGetStackHighWaterMark` check under a forced failure would settle whether that is enough.
- On the open question in the description: keep `ImuBno085` out of this PR, a follow-up is fine.

</details>

Reviewed and tested with Claude Fable 5.1 (`claude-fable-5-1`).

Johannes-Thiel:
Thanks, all three main points confirmed against the code. Addressed in 0c1342c:

**1. Console flood.** A `vTaskDelay` in the `catch` would not have fixed the case you measured. In your disturbed-bus run no read failed completely, so the `catch` never ran. The 106 lines/s come from the `ESP_LOGE` in the retry rounds of reads that eventually succeed. So now it is both:
- `esp_log_level_set("BNO055", ESP_LOG_NONE)` in the constructor, which leaves the vendored driver untouched;
- a 1 s back-off after a read that fails all retries. The back-off sleep can be woken early, which the destructor below uses.

The downside: a bus that is flaky but still recovering is now invisible on the console. I think that is right, because those lines broke the protocol stream and told the host nothing it could act on.

**2. Error semantics.** Your reading is right: `run_step` catches only `std::runtime_error` and `BNO055I2CError` derives from `std::exception`, so on `main` a hard read failure calls `std::terminate`. I've corrected the PR description. This PR turns that reboot into a running system, so the failure has to stay visible:
- every failed read is reported, which the back-off limits to about once per second while the sensor is down;
- `module "imu": reading the imu works again` is printed when reads succeed again;
- `Module::step()` now runs before the error is thrown, so the broadcast continues on failing cycles.

A validity property (`imu.ok` or similar) would be cleaner for hosts, but it changes the API. I'd leave it for a follow-up.

**3. Resource hygiene.** The handle is kept now, and `~Imu()` stops the task and deletes queue, mutex and a stop semaphore. I did not use `vTaskDelete` from outside, even while holding `bno_mutex`. The guard is released during unwinding, so the task can still be inside the exception runtime (`__cxa_free_exception` → `free`) holding the heap lock. Instead the destructor sets a stop flag, wakes the back-off with a notification and waits for the task to confirm before freeing anything.

**Minor:** `MutexGuard` moved to `main/utils/mutex_guard.h`. The `set_mode` latency is noted in the description, with no code change. The stack high-water mark under forced failure is not measured yet; with the driver silenced, the 64 `vprintf` calls are gone from that path. `ImuBno085` stays out, as agreed.

Built for esp32 with IDF 5.3.1, not yet re-tested on hardware. A run of your fault setup on the F30 would be appreciated. [※](https://zauberzeug.github.io/colophon/#m=claude-opus-5-5&t=Code%20review%20findings%20checked%20and%20fixes%20implemented%20by%20the%20model%20at%20the%20author%27s%20direction%3B%20wording%20from%20the%20model.&a=claude-code "claude-opus-5-5: Code review findings checked and fixes implemented by the model at the author's direction; wording from the model.")

JensOgorek (approved):
### Rework 0c1342c re-tested on our F30: console silent on a bad bus (0 library lines instead of 2117), loop unchanged at 20 ms, all three points addressed

Same setup as yesterday, upstream esp32 build of `0c1342c` on the F30 core with the BNO055 added live. The loop stays at 20 ms with all nine blocks read and values updating, the wrong-address constructor produces 0 library lines (128 before), and the 25 s flaky-bus run produces 0 lines (2117 before) with the loop at 20 ms and no reset. The three review points are addressed the way I would have done it, so this is good to merge from my side. Two things remain untested on hardware, see below; neither blocks the merge.

<details>
<summary>Measured on F30 (rb68, classic ESP32 core, console 460800)</summary>

| phase | `d6df2a1` (yesterday) | `0c1342c` (today) |
|---|---|---|
| loop without IMU / with IMU 65535 / 33 / 0 | 20 / 20 / 20 / 20 ms | 20 / 20 / 20 / 20 ms, max 25 |
| heap without / with IMU | 91.7 / 81.9 KB | 91.8 / 81.8 KB |
| wrong-address `Imu(0, 21, 22, 41)` | 128 library lines in 199 ms | 0 library lines |
| flaky bus via `imu2 = Imu(1, 21, 22)`, 25 s | 2117 library lines (106/s), loop 10–20 ms | 0 library lines, loop 20 / 20 / 24 ms, no millis reset |

As yesterday, no read failed all 64 rounds in the flaky-bus run, so the back-off, the once-per-second error and the "works again" line did not trigger here. Verifying those needs the sensor unplugged, which nobody could do at the robot today. I also did not catch the constructor's "imu setup failed" line in the wrong-address run; that may be my capture window, worth a glance when someone has a console open.

</details>

<details>
<summary>Code notes on 0c1342c</summary>

- `run_step` wraps each module's `step()` separately, so the per-second throw only costs the IMU its own step, and `Module::step()` running before the throw keeps the broadcast going. Good.
- The cooperative stop in `~Imu()` is right (deleting the task from outside could catch it inside the exception runtime). One consequence to know: if the destructor runs while the task is inside a failing read, it waits for that pass to finish, up to ~3 s for timeouts (64 rounds × 50 ms), under the `InterpreterLock`. Nothing calls the destructor today, so just a note.
- `esp_log_level_set("BNO055", ESP_LOG_NONE)` also silences the driver during `begin()`; the constructor's own `runtime_error` still carries the driver's message, so nothing is lost.

</details>

Reviewed and tested with Claude Fable 5.1 (`claude-fable-5-1`).

Johannes-Thiel:
Okay, should we merge or do we wait for final approval from @falkoschindler ?

falkoschindler (approved):
Thanks @Johannes-Thiel, this is good to merge from my side. The task, the one-slot queue and `bno_mutex` are the right shape, the cooperative stop in `~Imu()` is correct, and a failed read now keeps the firmware running instead of terminating it as on `main`. Jens's F30 measurements cover the loop side.

I pushed two small commits to the branch so nothing needs another round-trip. 13b562b fixes the `imu.temp` bug you mention in the description: `temp` is an `IntegerVariable`, so the value has to go into `integer_value`; it was already broken on `main`. be9c85c changes "about once per second" to "at most once per second" in the docs and the comment in `step()`, because on a stuck bus a failing read spends all 64 retry rounds in the timeout before the back-off, so the error comes every few seconds rather than every second.

One point stays open on purpose: the read task's 4 KB stack is not measured under a C++ throw on the failure path. The BNO055 is soldered onto the Robot Brain, so forcing a dead sensor needs a bench setup, and the worst case is a reboot, which is what `main` does on a dead sensor today anyway. If someone is at the F30 with a console open, a throwaway build that switches the driver to a wrong I²C address after a few reads would exercise that path without touching hardware.

This is an AI-assisted review pass; I did not verify every point myself.

<details>

**1. `imu.temp` always prints 0** main/modules/imu.cpp (240) — `temp` is created as an `IntegerVariable`, but `publish()` wrote to `number_value`; `print_to_buffer` reads `integer_value` for integers. Pre-existing on `main`. Fixed in 13b562b.

**2. "About once per second" only holds for a NACKing sensor** docs/module_reference.md (337), main/modules/imu.cpp (183) — On a stuck bus the failing transaction alone takes 64 timeout rounds (2–3 s) before the 1 s back-off. Fixed in be9c85c by saying "at most once per second".

**3. Read task stack under a C++ throw is unmeasured** main/modules/imu.cpp (9) — 4096 bytes with `BNO055I2CError` thrown and unwound inside the task. The `vprintf` calls are gone with the silenced log tag, and ESP-IDF's own exception example throws in a 3.5 KB main task, so it is probably fine, but not shown. Left open, see above.

</details>

[※](https://zauberzeug.github.io/colophon/#m=claude-fable-5-1&t=Review%20pass%2C%20code%20checks%20and%20wording%20by%20the%20model%3B%20which%20findings%20to%20act%20on%20and%20what%20to%20leave%20open%20was%20decided%20by%20the%20author.&a=claude-code "claude-fable-5-1: Review pass, code checks and wording by the model; which findings to act on and what to leave open was decided by the author.")

